PR #12, opened 2021-10-14
fix: Update traefik docs link

gracedo:
https://jira.d2iq.com/browse/D2IQ-80123


msdolbey (approved):
👍 
Thanks!
It is true!

PR #21, opened 2021-10-29
fix: make clean install of Istio work again

asekretenko:
Now that Istio is installed into the `istio-system` namespace, two more changes are needed  to make the clean install work:
 - the namespace itself needs to be created before additional objects (`istso-ca` Certificate, etc.) can be created in it
 - we have to switch to using the ClusterIssuer instead of an Issuer, as there are no Issuers in the `istio-system` NS

### DIFF
--- a/services/istio/1.9.1/helmrelease.yaml
+++ b/services/istio/1.9.1/helmrelease.yaml
@@ -6,7 +6,7 @@ metadata:
 spec:
   force: false
   prune: true
-  interval: 1m0s
+  interval: 10s
   path: ./services/istio/1.9.1/helmrelease
   dependsOn:
     - name: istio-ca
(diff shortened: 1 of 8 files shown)


branden (dismissed):
I put together a kommander branch that pulls in this change and ran kuttl tests against it, but the istio_stack test still failed: https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_Kommander2_KuttlTests/3270278?buildTab=log&focusLine=2281&expandAll=true

aweris (changes requested):
This will not work. applications have defined resources created outside of the helm release.  

We need to create a namespace outside of the helm release or create a dependency between helm release and additional resources

https://github.com/mesosphere/kommander-applications/blob/172ca024c793014fa0ee5a119f8e70c9925bf841/services/istio/1.9.1/istio.yaml#L32-L52

vespian (commented):
Considering the fact that we are already using a custom istio chart:

https://github.com/mesosphere/charts/tree/master/staging/istio

Would it make sense to move these changes to the chart instead and keep the kustomizations schema simple?

vespian (approved):

asekretenko:
Force-pushed with signed commits.

PR #22, opened 2021-11-01
feat: create Ingress for Gitea

makkes:
This is needed so that attached clusters can access the Git
repository.

All repositories are accessed through the URL

https://INGRESS_HOST/dkp/kommander/gitserver/USER|ORG/REPO

refs https://jira.d2iq.com/browse/D2IQ-80105

### DIFF
--- a/services/gitea/4.1.1/defaults/cm.yaml
+++ b/services/gitea/4.1.1/defaults/cm.yaml
@@ -17,14 +17,14 @@ data:
         traefik.ingress.kubernetes.io/router.tls: "true"
       hosts:
         - paths:
-          - path: /dkp/kommander/gitserver
+          - path: /dkp/kommander/git
             pathType: ImplementationSpecific
     gitea:
       admin:
         existingSecret: ${adminCredentialsSecret}
       config:
         server:
-          ROOT_URL: https://localhost/dkp/kommander/gitserver
+          ROOT_URL: https://localhost/dkp/kommander/git
           PROTOCOL: https
           CERT_FILE: /git-tls/tls.crt
           KEY_FILE: /git-tls/tls.key
--- a/services/gitea/4.1.1/defaults/cm.yaml
+++ b/services/gitea/4.1.1/defaults/cm.yaml
@@ -24,7 +24,7 @@ data:
         existingSecret: ${adminCredentialsSecret}
       config:
         server:
-          ROOT_URL: https://localhost
+          ROOT_URL: https://localhost/dkp/kommander/gitserver
           PROTOCOL: https
           CERT_FILE: /git-tls/tls.crt
           KEY_FILE: /git-tls/tls.key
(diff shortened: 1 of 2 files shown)


services/gitea/4.1.1/defaults/cm.yaml
```diff
@@ -9,11 +9,22 @@ data:
...
       rootless: true
+    ingress:
+      enabled: true
+      annotations:
+        kubernetes.io/ingress.class: kommander-traefik
+        traefik.ingress.kubernetes.io/router.middlewares: kommander-stripprefixes@kubernetescrd
+        traefik.ingress.kubernetes.io/router.tls: "true"
+      hosts:
+        - paths:
+          - path: /dkp/kommander/gitserver
```
jimmidyson, line 20: what about just `/dkp/kommander/git`?

makkes: I'm fine with that. `gitserver` is what's used in Kommander 2.0 but since the Git repo URL will change, anyway we can just as well use `git` here.

makkes, line 20:
```suggestion
          - path: /dkp/kommander/git
```

makkes: Duh, the mobile app didn't let me edit the commit msg

services/gitea/4.1.1/defaults/cm.yaml
```diff
@@ -9,11 +9,22 @@ data:
...
+      hosts:
+        - paths:
+          - path: /dkp/kommander/git
+            pathType: ImplementationSpecific
     gitea:
       admin:
         existingSecret: ${adminCredentialsSecret}
       config:
         server:
+          ROOT_URL: https://localhost/dkp/kommander/git
```
jimmidyson, line 27: Won't this need to be changed to use the ingress hostname?

services/gitea/4.1.1/defaults/cm.yaml
mikolajb: Should it be added to traefik prefix overrides? (it resides in this repo and in kommander-cli as well - in the migration manifests).

mikolajb (approved):
Changeing root url will require changes in a flux controller and in a few other places but if not now then when 🙂

services/gitea/4.1.1/defaults/cm.yaml
makkes: The only reason I'm setting this here is that the Helm chart won't install with it not being set if the ingress has no hostname set: https://gitea.com/gitea/helm-chart/src/branch/master/templates/gitea/config.yaml#L58-L68

Since we obviously can only know the hostname at runtime, I'm just setting `ROOT_URL` to a hard-coded value here. For our purposes of simply serving a Git repository over HTTPS this `ROOT_URL` doesn't matter.

Eventually we may want to adapt installation to set the actual hostname but that doesn't need to happen right now.

services/gitea/4.1.1/defaults/cm.yaml
makkes: @mikolajb done

PR #32, opened 2021-11-05
refactor: Deploy kubefed to kube-federation-system

gracedo:
Deploy kubefed control plane to dedicated `kube-federation-system` namespace instead of deploying it to `kommander`. This is mainly done so that we are able to federate the `kommander` namespace to use it as the workspace ns. I was unable to federate that namespace until I moved the kubefed control plane out of it (looks like this is by design in kubefed https://github.com/kubernetes-sigs/kubefed/blob/master/pkg/controller/sync/accessor.go#L183-L186).

Regardless of when we get the change in to use `kommander` ns for the kommander workspace, I still think it's a good idea to deploy kubefed to a dedicated namespace, that seems to be the recommended way of doing things. And in case we ever need to federate anything to itself in the `kommander` ns, we wouldn't get blocked by this kubefed restriction.

### DIFF
--- a/services/kubefed/0.8.1/defaults/cm.yaml
+++ b/services/kubefed/0.8.1/defaults/cm.yaml
@@ -1,0 +1,32 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubefed-0.8.1-d2iq-defaults
+  namespace: ${releaseNamespace}
+data:
+  values.yaml: |
+    controllermanager:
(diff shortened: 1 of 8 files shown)


takirala (approved):

branden (approved):

gracedo:
I'm going to hold on merging this until we can coordinate with the UI team so we get these changes in together to avoid breaking the UI during bug bash

PR #37, opened 2021-11-09
chore: Use Sqlite3 for gitea

mikolajb:
Using sqlite3 storage reduces a number of statefulsets.


mikolajb:
Closing it since there is a fix #38

PR #41, opened 2021-11-10
fix: allow plugin for kube-prome-stack

cwyl02:
applied the same thing to the centralized grafana already but missed this one


branden (changes requested):
`grafana.plugins` is a list of plugins to install, see https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L442. This is plugin configuration that needs to go under `grafana."grafana.ini".plugins`, like https://github.com/mesosphere/kommander/pull/844.

gracedo (approved):

branden (approved):

PR #42, opened 2021-11-11
fix: Add missing relations to traefik

mikolajb:
Adding missing dependencies to traefik.


mikolajb:
Closed in favor of #43

PR #45, opened 2021-11-12
fix: don't retry installing Kommander

makkes:
The Kommander chart cannot be uninstalled so it isn't worth even
trying to remediate any installation failure.

### DIFF
--- a/services/kommander/0.1.0/kommander.yaml
+++ b/services/kommander/0.1.0/kommander.yaml
@@ -19,12 +19,21 @@ spec:
         namespace: kommander-flux
       version: "${chartVersion}"
   interval: 15s
+  # Kommander is quite a big chart and it may need some more time than
+  # other charts to get ready so setting this to 10 minutes increases
+  # the chance of the installation not timing out.
   timeout: 10m
   install:
     remediation:
+      # The Kommander chart cannot be uninstalled in its current form. That's
+      # a known issue and trying to remediate an installation failure by
+      # deleting and retrying won't work so we can just as well not retry to
+      # begin with.
       retries: 0
   upgrade:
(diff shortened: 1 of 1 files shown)


jimmidyson (approved):
LGTM - can we also bump the timeout to 10m or something?

services/kommander/0.1.0/kommander.yaml
```diff
@@ -21,10 +21,10 @@ spec:
   interval: 15s
   install:
     remediation:
-      retries: 30
+      retries: 0
   upgrade:
     remediation:
-      retries: 30
+      retries: 0
```
asekretenko, line 27: Upgrades are remiediated via rollback by default (https://fluxcd.io/docs/components/helm/api/#helm.toolkit.fluxcd.io/v2beta1.UpgradeRemediation) 
Do we know this to be (or  consider this) equally harmful?

services/kommander/0.1.0/kommander.yaml
```diff
@@ -21,10 +21,10 @@ spec:
   interval: 15s
   install:
     remediation:
-      retries: 30
+      retries: 0
```
jimmidyson, line 24: Actually, can we add a comment for this too?

jimmidyson (changes requested):

services/kommander/0.1.0/kommander.yaml
makkes: I'm 99% sure a rollback won't work with kommander. Completely untested at this point.

jimmidyson (approved):
Thanks!

asekretenko (approved):

PR #47, opened 2021-11-15
feat: add tooling for updating Flux

makkes:
This change bootstraps a Make configuration for running targets in CI
as well as a 'flux-update' target that updates the manifests of Flux
to the latest Flux version.

https://jira.d2iq.com/browse/D2IQ-81179

### DIFF
--- a/hack/flux/update-flux.sh
+++ b/hack/flux/update-flux.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+IFS=$'\n\t'
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+readonly REPO_ROOT
+LATEST_FLUX_VERSION="$(gh api -X GET "repos/fluxcd/flux2/releases" --jq '.[0].tag_name|sub("^v"; "")')"
+readonly LATEST_FLUX_VERSION
+CURRENT_FLUX_VERSION=$(find "${REPO_ROOT}/services/kommander-flux" -maxdepth 1 -regextype sed -regex '.*/[0-9]\+.[0-9]\+.[0-9]\+' -printf "%f\n" | sort -V | head -1)
+readonly CURRENT_FLUX_VERSION
+
+function update_flux() {
+    readonly BRANCH_NAME="flux-update/${LATEST_FLUX_VERSION}"
+    git checkout -b "${BRANCH_NAME}"
+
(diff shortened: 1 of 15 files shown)


aweris (commented):
LGTM :+1:, I really like how you automate the update.  I'll test this locally then approve

aweris (approved):
:+1: , We should validate with CI as well.

Just want to make sure `gh` able to authenticate using CI token passed from env

hack/flux/update-flux.sh
```diff
@@ -0,0 +1,56 @@
...
+    readonly BRANCH_NAME="flux-update/${LATEST_FLUX_VERSION}"
+    git checkout -b "${BRANCH_NAME}"
+
+    asdf install flux2 "${LATEST_FLUX_VERSION}"
+    asdf local flux2 "${LATEST_FLUX_VERSION}"
+
+    mkdir -p "$REPO_ROOT/services/kommander-flux/$LATEST_FLUX_VERSION"
+    pushd "$REPO_ROOT/services/kommander-flux/$LATEST_FLUX_VERSION"
+    ls ..
+    cp -a ../"$CURRENT_FLUX_VERSION"/* .
```
takirala, line 22: was it intentional to NOT remove the current flux version directory?

makkes: Oops, not exactly. I believe it doesn't hurt having it there, though.

takirala: i just tried this locally and, (rather untentionally) created #48 but it has both the versions of flux.

aweris:
@makkes  didn't see auto-merge.  I was asking CI test as well but it merged

makkes:
> :+1: , We should validate with CI as well.
> 
> Just want to make sure `gh` able to authenticate using CI token passed from env

Yes, that's the next subtask in the Jira.

hack/flux/update-flux.sh
takirala: if we keep multiple versions of flux, the diff is harder to review (like in #48) also not sure if https://github.com/mesosphere/kommander/blob/main/Makefile#L30 would break anytime in future (sort may behave weird if semantic versioning is not strictly followed), i ll create a follow up PR to remove the old directory and that should keep things simple.

makkes: Thanks @takirala!!

takirala: https://github.com/mesosphere/kommander-applications/pull/49/files

PR #53, opened 2021-11-15
feat: Upgrade flux to 0.23.0

d2iq-dispatch:


services/kommander-flux/0.23.0/templates/kustomization.yaml
```diff
@@ -21,6 +21,7 @@ resources:
 - rbac.authorization.k8s.io_v1_clusterrole_crd-controller-kommander-flux.yaml
 - rbac.authorization.k8s.io_v1_clusterrolebinding_cluster-reconciler-kommander-flux.yaml
 - rbac.authorization.k8s.io_v1_clusterrolebinding_crd-controller-kommander-flux.yaml
+- v1_namespace_kommander-flux.yaml
```
takirala, line 24: the namespace is created before these manifests are applied (here - https://github.com/mesosphere/kommander/blob/main/federation/pkg/controllers/kommandercluster_flux_controller.go#L142) in target clusters. not sure if having this here affects anything.

makkes: I noticed that, too, yesterday but figured it doesn't hurt having it here. But we need to try it out. Let me do that today.

makkes: We actually [do the same in the CLI](https://github.com/mesosphere/kommander-cli/blob/main/pkg/installer/manifests/base/kommander-flux-namespace.yaml) so this is not a problem.

makkes (approved):

rishabh96b (approved):

makkes:
Force-pushed a signed commit so we can merge.

PR #72, opened 2021-11-24
feat: add an audit dashboard to grafana-logging

asekretenko:
This adds an approximate feature parity with the Kibana dashboard in 1.8 (https://jira.d2iq.com/browse/D2IQ-78016)

Tested manually on top of Konvoy 2.1 on AWS:
audit_dashboard

Note that we could have been better off if we made Grafana parse out verb/username/namespace/resource from the Loki response instead of adding labels when sending the audit log into Loki. Unfortunately, this requires a newer Grafana - and we want to backport this into 2.1.1.  Thus, this PR makes use of what we have in the current version of Grafana.

### DIFF
--- a/grafana-dashboards/logging/audit.json
+++ b/grafana-dashboards/logging/audit.json
@@ -1,0 +1,492 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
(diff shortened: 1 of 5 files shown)


gracedo (approved):
🎉  looks great!

branden (approved):
Looks great, thanks!

asekretenko:
Force-pushed with a signed commit

asekretenko:

PR #73, opened 2021-11-24
feat: Bump thanos to 0.4.6

gracedo:
Bump thanos chart to latest. Instead of overwriting the old 0.4.5 thanos version, created a new 0.4.6 dir. This is for upgrades, in case we need to leave the old versions around


gracedo:
Passed CI in kommander https://github.com/mesosphere/kommander/pull/1186

gracedo:
Pushed a change to instead create a new 0.4.6 dir instead of replacing 0.4.5, in case we need to keep around older versions for upgrades... but it looks like when running this in kommander, the older version of thanos gets deployed https://github.com/mesosphere/kommander/pull/1186

gracedo:
> Pushed a change to instead create a new 0.4.6 dir instead of replacing 0.4.5, in case we need to keep around older versions for upgrades... but it looks like when running this in kommander, the older version of thanos gets deployed [mesosphere/kommander#1186](https://github.com/mesosphere/kommander/pull/1186)

Looks like this was a bug in `kommander-cli` and was fixed in https://github.com/mesosphere/kommander-cli/pull/284 🥳

makkes:

cbuto (approved):

takirala (approved):

PR #80, opened 2021-11-29
chore: Use workspaceNamespace instead of traefikNamespace

gracedo:
We originally needed a separate `traefikNamespace` substitution variable when traefik wasn't installed in the workspace namespace, but now that it is (`kommander` ns is workspace ns) we can remove the var and just use `workspaceNamespace` instead.

### DIFF
--- a/services/traefik-forward-auth/0.3.2/traefik-forward-auth.yaml
+++ b/services/traefik-forward-auth/0.3.2/traefik-forward-auth.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: ${releaseNamespace}
 spec:
   dependsOn:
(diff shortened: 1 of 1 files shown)


takirala (approved):
does this mean we can remove ALL references of `traefikNamespace` in kommander repo?

gracedo:
> does this mean we can remove ALL references of `traefikNamespace` in kommander repo?

@takirala yep! I have a branch in `kommander` queued up to do that

gracedo:
Waiting to merge this one until this bump is taken care of first https://github.com/mesosphere/kommander/pull/1201

branden (approved):

PR #84, opened 2021-12-03
chore: Add kubecost 0.20.2

gracedo:
Bumps kubecost + bumps sidecar image to address CVE
https://jira.d2iq.com/browse/D2IQ-81263

New version is added here instead of overwriting the older version in case we need to retain 2.1.0 versions for upgrades.

### DIFF
--- a/services/centralized-kubecost/0.20.2/defaults/cm.yaml
+++ b/services/centralized-kubecost/0.20.2/defaults/cm.yaml
@@ -63,7 +63,7 @@ data:
         # These values are set so that kubecost grafana dashboards are installed.
         # Grafana itself is not installed.
         sidecar:
-          image: kiwigrid/k8s-sidecar:1.3.0
+          image: kiwigrid/k8s-sidecar:1.12.3
           dashboards:
             enabled: true
             label: grafana_dashboard_kommander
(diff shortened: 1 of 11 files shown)


gracedo:
Merge this is https://github.com/mesosphere/kommander/pull/1222 goes green

services/centralized-kubecost/0.20.2/defaults/cm.yaml
```diff
@@ -0,0 +1,143 @@
...
+          vendor.kubecost.io/partner: d2iq
+
+      podSecurityPolicy:
+        enabled: false
+
+      grafana:
+        # These values are set so that kubecost grafana dashboards are installed.
+        # Grafana itself is not installed.
+        sidecar:
+          image: kiwigrid/k8s-sidecar:1.3.0
```
rishabh96b, line 66: We need this to be bumped to `1.12.3` to mitigate the CVEs.
```suggestion
          image: kiwigrid/k8s-sidecar:1.12.3
```

rishabh96b (approved):

services/centralized-kubecost/0.20.2/defaults/cm.yaml
```diff
@@ -0,0 +1,143 @@
...
+          vendor.kubecost.io/partner: d2iq
+
+      podSecurityPolicy:
+        enabled: false
+
+      grafana:
+        # These values are set so that kubecost grafana dashboards are installed.
+        # Grafana itself is not installed.
+        sidecar:
+          image: kiwigrid/k8s-sidecar:1.12.3
```
gracedo, line 66:
```suggestion
```
We should be able to just remove this override since we're including it in the kubecost chart

PR #95, opened 2021-12-14
feat: Provide Gatekeeper 3.7.0

mikolajb:
## Provide gatekeeper 3.7.0 to support kubernetes 1.22

Jira: https://jira.d2iq.com/browse/D2IQ-81722
This PR changes the gatekeeper setup, instead of using customized helm chart, I propose to use the upstream chart and move changes to a separate chart. In that way, the gatekeeper service consists of two helm releases:
1. upstream gatekeeper chart
1. `gatekeeper-setup` hosted in `kommander-applications` repository

The reason for this change is the fact that tracking changes across version bumps is very hard.

### Why can't we use the upstream chart and move the rest to kustomizations?
I moved cert creation to kustomizations but other things (webooks) depend on helm values.

### How to review?
- our gatekeeper chart: https://github.com/mesosphere/charts/tree/master/staging/gatekeeper
- upstream gatekeeper chart: https://github.com/open-policy-agent/gatekeeper/tree/master/charts/gatekeeper


takirala:
We should pull gatekeeper from `mesosphere/charts` (as we always did).

If there is some reasoning behind skipping `mesosphere/charts` (i am ooo so i may have missed some conversations), we should atleast get all the patches we have for gatekeeper chart in `mesosphere/chart`.

The mesosphere/charts gatekeeper is at an older version but I do have a PR https://github.com/mesosphere/charts/pull/1245 that bumps gatekeeper chart to 3.6.0 if we want to reuse that.

gracedo:
We also have this jira https://jira.d2iq.com/browse/D2IQ-82034 tracking gatekeeper bump if we want to reuse it here

mikolajb:
> We should pull gatekeeper from `mesosphere/charts` (as we always did).
> 
> If there is some reasoning behind skipping `mesosphere/charts` (i am ooo so i may have missed some conversations), we should atleast get all the patches we have for gatekeeper chart in `mesosphere/chart`.
> 
> The mesosphere/charts gatekeeper is at an older version but I do have a PR [mesosphere/charts#1245](https://github.com/mesosphere/charts/pull/1245) that bumps gatekeeper chart to 3.6.0 if we want to reuse that.

@takirala Excellent, thanks for the PR! I can take care of it so you can focus on catalog apps. WDYT?

makkes:

vespian (approved):

mikolajb:
Superseded by #186

PR #106, opened 2021-12-16
feat: add kaptain

mikolajb:
Allow to install kaptain as kommander's AppDeployment. [D2IQ-81777](https://jira.d2iq.com/browse/D2IQ-81777)

Changes are used here: https://github.com/mesosphere/kudo-kubeflow/pull/600

### DIFF
--- a/services/kaptain/1.3.0/kaptain.yaml
+++ b/services/kaptain/1.3.0/kaptain.yaml
@@ -1,0 +1,27 @@
+---
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
+kind: HelmRelease
+metadata:
+  name: kaptain
+  namespace: ${releaseNamespace}
+spec:
+  chart:
+    spec:
+      chart: kaptain
+      sourceRef:
+        kind: HelmRepository
+        name: kaptain-helm-mirror
+        namespace: kommander-flux
+      version: 1.3.0-dev
+  interval: 15s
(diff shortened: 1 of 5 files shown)


services/kaptain/1.3.0/kaptain.yaml
```diff
@@ -0,0 +1,27 @@
...
+  namespace: ${releaseNamespace}
+spec:
+  chart:
+    spec:
+      chart: kaptain
+      sourceRef:
+        kind: HelmRepository
+        name: kaptain-helm-mirror
+        namespace: kommander-flux
+      version: 1.3.0-dev
```
alembiewski, line 15: should we set the upcoming release version here? `1.3.0`?

alembiewski (approved):
One nit from me, the rest LGTM!

services/kaptain/1.3.0/kaptain.yaml
mikolajb: This is the values from https://github.com/mesosphere/kudo-kubeflow/blob/3c1a181f6852159ea3ef6b0d3b5d2ef45241aae0/kudo-kubeflow/helm/chart/Chart.yaml#L3

But in general, it will have to be adjusted once the release process is of the kaptain chart is provided. I would say, let's keep it as it is and revisit it when we start working on the release process.

PR #119, opened 2021-12-23
feat: add chartmuseum app

vespian:
This PR adds `chartmuseum` service which is going to be used to refactor our static helm-mirror into dynamic one that allows for uploading charts, as part of https://jira.d2iq.com/browse/D2IQ-82657


aweris (approved):
Not really liked the patching chart part however, it's still better than using a custom chart. Otherwise LGTM :+1:

PR #122, opened 2021-12-30
feat: Upgrade logging-operator and fluent-bit

branden:
https://jira.d2iq.com/browse/D2IQ-81543

This upgrades `services/logging-operator` (for pod logs), and updates default configuration to match the new upstream defaults. This also upgrades `services/fluent-bit` (for admin logs) to the same version that's deployed by logging-operator.


takirala (dismissed):

branden:
It looks like there might be a real test failure on the corresponding kommander PR, so I'll move this out of ready-for-review while I investigate.

makkes:
@mesosphere-mergebot test all

makkes:

branden:
Checks on mesosphere/kommander#1300 passed after updating a test, so I'll move this to ready for review.

gracedo (approved):
@makkes for upgrades, do you know if we need to keep the older version around? (vs directly replacing with upgraded versions which is what we've been doing)

takirala (approved):
lgtm modulo the upgrade concerns (adding vs replacing)

makkes:
Replacing here is fine as we're basically only adding files during an upgrade.

branden:
We'll need to add this commit to the kommander PR that bumps kommander-applications to pull in this change: https://github.com/mesosphere/kommander/commit/8e0700992e2e1459340b9a29b35e09647de28d3a

PR #123, opened 2022-01-04
feat: disable user registration in Gitea

makkes:
We currently don't want users to be able to register an account in
Gitea.

### DIFF
--- a/services/gitea/4.1.1/defaults/cm.yaml
+++ b/services/gitea/4.1.1/defaults/cm.yaml
@@ -31,6 +31,7 @@ data:
           HTTP_PORT: 3000
         service:
           REQUIRE_SIGNIN_VIEW: true
+          DISABLE_REGISTRATION: true
     service:
       http:
         port: 443


takirala (commented):
Our docs instruct the user to Register a new user for airgapped catalog https://docs.d2iq.com/dkp/kommander/2.1/install/air-gapped/catalog/

makkes:
> Our docs instruct the user to Register a new user for airgapped catalog https://docs.d2iq.com/dkp/kommander/2.1/install/air-gapped/catalog/ (Step 3)

Those can still be created using the admin user. This only disables self-registration.

takirala:
What do we gain be disabling self registration ? 

If the user wants to use admin creds then  they would have to fetch a secret and copy that secret into all target clusters. And during upgrades, that secret may change (right?) and they would have to do it again. The current docs instruct the user to self register and we would have to update those docs if we merge this PR.

makkes:
> What do we gain be disabling self registration ?

Security. At the moment everyone with access to the cluster's LB can just navigate the the Gitea web app and register an account. We need to make sure that (write) access to Gitea is restricted in some way.

> If the user wants to use admin creds then they would have to fetch a secret and copy that secret into all target clusters (in corresponding workspace/project namespaces). And during upgrades, that secret may change (right?) and they would have to do it again. The current docs instruct the user to self register and we would have to update those docs if we merge this PR.

I didn't say they should use admin creds, much to the contrary. It's just that opening up self-registration to everyone is not a security model we should encourage. Some kind of authentication or access control must be possible.

takirala (approved):

PR #126, opened 2022-01-05
feat: add dummy helm releases

vespian:
This PR is meant to unblock https://jira.d2iq.com/browse/D2IQ-82657 on one hand and on the other - enable pararelising the upgrades work as once it gets merged it can be worked on independently.

The idea behind this PR is to add dummy helm releases that will replace dynamic helm releases generated by:
* federation/pkg/controllers/kommandercluster_prometheus_proxy.go
* federation/pkg/controllers/kommandercluster_observer_controller.go

This way the charts parsing logic that populates helm-mirror/chartmuseum in https://github.com/mesosphere/kommander-cli/pull/298 will be able to determine versions of charts to download while still letting controllers create helm releases dynamically. The only difference is going to be that the controllers will be creating Flux kustomizations that point to these templates instead.

### DIFF
--- a/services/kommander/0.1.0/dynamic-helmreleases/README.txt
+++ b/services/kommander/0.1.0/dynamic-helmreleases/README.txt
@@ -1,6 +1,11 @@
-The purpose of these directories is to have ALL HelmRelease objects in the k-apps
-repo, including the ones that are generated dynamically.
+The purpose of these directories is to have ALL HelmRelease objects in the
+k-apps repo, including the ones that are generated dynamically.
 
-This way fetching charts and creating charts bundle can be done without any extra
-information - simply by scanning k-apps repository. The controllers that create
-these helmreleases create kustomizations instead that change these charts accordingly.
+By dynamically created, we mean HelmReleases which are created/maintained by
+controllers as opposed to simply creating Appdeployments and updating git
+repository.
+
+Thanks to this approach, fetching charts and creating charts bundle can be done
+without any extra information - simply by scanning k-apps repository. The
+controllers that create these helmreleases create kustomizations instead that
+change these charts accordingly.
--- a/services/kommander/0.1.0/dynamic-helmreleases/README.txt
+++ b/services/kommander/0.1.0/dynamic-helmreleases/README.txt
@@ -1,0 +1,6 @@
+The purpose of these directories is to have ALL HelmRelease objects in the k-apps
+repo, including the ones that are generated dynamically.
+
+This way fetching charts and creating charts bundle can be done without any extra
+information - simply by scanning k-apps repository. The controllers that create
+these helmreleases create kustomizations instead that change these charts accordingly.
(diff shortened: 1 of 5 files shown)


gracedo (commented):
> The only difference is going to be that the controllers will be creating Flux kustomizations that point to these templates instead.

So there will be a corresponding kommander pr updating the controllers? Can we review those changes (when ready) alongside this PR before merging please? I think we should be able to update the k-apps ref https://github.com/mesosphere/kommander/blob/main/Makefile#L60 to point to these changes here to test them together

makkes:
@mesosphere-mergebot test all

makkes:
@mesosphere-mergebot test all

makkes:
rebased branch to make pre-commit pass.

services/kommander/0.1.0/dynamic-helmreleases/README.txt
```diff
@@ -0,0 +1,6 @@
+The purpose of these directories is to have ALL HelmRelease objects in the k-apps
+repo, including the ones that are generated dynamically.
```
makkes, line 2: Can we add some explanation here on what "dynamically generated" HelmReleases are and why we need them, please? This will help our future selves recall what this dir is about.

vespian: Comment added.

vespian:
@makkes Comment added, can you re-review?

vespian:
@makkes Comment added, can you re-review?

makkes (approved):

vespian:
> Can we review those changes (when ready) alongside this PR before merging please? I think we should be able to update the k-apps ref https://github.com/mesosphere/kommander/blob/main/Makefile#L60 to point to these changes here to test them together

@gracedo This PR only unblocks https://jira.d2iq.com/browse/D2IQ-82657 The files themselves are not used yet, only there to make creating charts-bundles happy. So there is absolutely no risk here with merging it - just treat them as plain text files for now. 

The change to controllers themselves will happen in https://jira.d2iq.com/browse/D2IQ-83144 and it is quite possible that these files will evolve even more as there is some back-and-forth regarding the design.

services/kommander/0.1.0/dynamic-helmreleases/cluster-observer/cluster-observer.yaml
```diff
@@ -0,0 +1,29 @@
+---
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
+kind: HelmRelease
+metadata:
+  name: needsToBeOverriden
```
mikolajb, line 5: `overriden` ➡️ `overridden`

mikolajb (approved):

gracedo:
> The files themselves are not used yet, only there to make creating charts-bundles happy. So there is absolutely no risk here with merging it - just treat them as plain text files for now.

@vespian sgtm if there is no harm in merging these files without the corresponding controller, I'll unblock 👍🏻

services/kommander/0.1.0/dynamic-helmreleases/cluster-observer/cluster-observer.yaml
vespian: Done.

PR #131, opened 2022-01-11
feat: default catalog repositories in kommander

takirala:
Context  https://jira.d2iq.com/browse/D2IQ-79950
kommander PR https://github.com/mesosphere/kommander/pull/1329

### DIFF
--- a/common/defaults/project/catalog-repositories/kustomization.yaml
+++ b/common/defaults/project/catalog-repositories/kustomization.yaml
@@ -1,0 +1,2 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
--- a/common/defaults/project/kustomization.yaml
+++ b/common/defaults/project/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- catalog-repositories
--- a/common/defaults/workspace/catalog-repositories/kustomization.yaml
+++ b/common/defaults/workspace/catalog-repositories/kustomization.yaml
@@ -1,0 +1,2 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
--- a/common/defaults/workspace/kustomization.yaml
+++ b/common/defaults/workspace/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- catalog-repositories
--- a/services/gitea/4.1.1/defaults/cm.yaml
+++ b/services/gitea/4.1.1/defaults/cm.yaml
@@ -30,7 +30,7 @@ data:
           KEY_FILE: /git-tls/tls.key
           HTTP_PORT: 3000
         service:
-          REQUIRE_SIGNIN_VIEW: true
+          REQUIRE_SIGNIN_VIEW: false
           DISABLE_REGISTRATION: true
     service:
       http:


common/defaults/project/catalog-repositories/dkp-catalog-applications.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: dkp-catalog-applications
+  namespace: ${releaseNamespace}
+  labels:
+    kommander.d2iq.io/gitrepository-type: catalog
+    kommander.d2iq.io/gitapps-gitrepository-type: dkp
+spec:
+  gitImplementation: go-git
+  interval: 1m0s
+  ref:
+    branch: main
```
cbuto, line 13: 💭 are we always going to point to main? I thought we initially wanted to tag releases and try to match kommander releases. I’m not sure how we could substitute this for the release version though.

gracedo: Do we want to use a substitution variable here and set the version in kommander? (we would use the same version as kommander I guess, if we are keeping those versions coupled)

common/defaults/project/catalog-repositories/dkp-catalog-applications.yaml
```diff
@@ -0,0 +1,15 @@
...
+  timeout: 20s
+  url: https://github.com/mesosphere/dkp-catalog-applications
```
gracedo, line 15: for airgapped how would we point this to gitea server?

takirala: That is up for discussion, but the CLI (during bootstrap) will do that - https://docs.google.com/document/d/18vzWTmr3O7QNvFxbWAgwusMdT2kQ3ZVFE0YZLxT0Kc4/edit#heading=h.g6xak56mm4x3

makkes: Do we need to hold this PR until this is decided for good? I'm afraid this would break any air gap testing otherwise.

makkes (commented):
What are the impacts of having these resources unversioned on future upgrades?

common/defaults/project/catalog-repositories/dkp-catalog-applications.yaml
takirala: This will not break airgapped testing. flux will just fail to reconcile the repo thus behaving as the repo is non existent.
(in our current airgapped ci we do not perform any kuttl tests and only check if kommander is installable and this pr doesnt affect any of that)

takirala:
@mesosphere-mergebot test all

common/defaults/project/catalog-repositories/dkp-catalog-applications.yaml
takirala: Discussed this offline, decided to go with `branch: main` for now and as part of release process, we will swap this to `tag: v2.2.0`. I ll create a PR to reflect this at https://github.com/mesosphere/kommander/blob/main/CONTRIBUTING.md#publishing-a-new-release once this PR is merged.

gracedo (approved):

common/defaults/project/catalog-repositories/dkp-catalog-applications.yaml
makkes: > flux will just fail to reconcile the repo

from my point of view this is a problem and something customers will run into if not tackled. IIUC this plainly does not work in air gapped environments then?

takirala: restructured things a bit and moved the repository definitions to installer file. updated strucuture https://github.com/mesosphere/kommander-cli/pull/335#issue-1113034744 keeps things more obvious and it is upto the user to switch the `github.com/m/dkp-catalog-application` urls to local resources when installting kommander.

cbuto (approved):

cwyl02 (approved):

common/defaults/project/catalog-repositories/dkp-catalog-applications.yaml
takirala: Update the KEP and will update the release notes to change https://github.com/mesosphere/kommander-cli/pull/353/files#diff-2c390594da79f0b90f7bae6f6f8533240bb60d38c58ec8884caee53ba18aa845R10-R12 as part of release process (structure is TBD, but this will be the place to change)

PR #138, opened 2022-01-13
[release-2.1] build: Bump mergebot fixversion map

gracedo:


gracedo:
pre-commit expected to fail on release branch since checks were added after the fact, so there are some commits that don't follow conventional commits.

cbuto (approved):

PR #142, opened 2022-01-14
feat: Upgrade flux to 0.25.2

d2iq-dispatch:


takirala (approved):
looks like only a bug fix for kustomize-controller is added. no changes in other flux components.

gracedo (approved):

PR #160, opened 2022-01-24
feat: bump fluent-bit to 0.19.17

cwyl02:
https://jira.d2iq.com/browse/D2IQ-83947


services/fluent-bit/0.19.17/fluent-bit.yaml
```diff
@@ -14,7 +14,7 @@ spec:
         namespace: kommander-flux
       # This chart version should be kept in sync with logging-operator.
       # Both services should deploy the same version of fluent-bit.
-      version: 0.19.5
+      version: 0.19.17
```
branden, line 17: We want to avoid deploying multiple versions of fluent-bit in order to keep the airgapped image bundle as small as possible. Per the comment, we want to keep this fluent-bit's version in sync with the fluent-bit managed by logging-operator: https://github.com/mesosphere/kommander-applications/blob/c005334b796348d1f027a224dbbabf21811581ed/services/logging-operator/3.16.0/defaults/cm.yaml#L100

Version 0.19.17 of the fluent-bit chart deploys 1.8.11 by default:
https://github.com/fluent/helm-charts/blob/f88da97278702e48dbe57ee819c3e761a919acdc/charts/fluent-bit/Chart.yaml#L8-L9

The latest available version of logging-operator deploys fluent-bit 1.8.9 by default: https://github.com/banzaicloud/logging-operator/tree/master/charts/logging-operator-logging#configuration

We could either deploy this chart's default fluent-bit version and override logging-operator to match, or deploy logging-operator's slightly older default and override the fluent-bit chart to match. The first approach would give us the latest version of fluent-bit, and the second approach would give us a slightly older version that's more likely to have been tested with logging-operator. Unless we have a specific fix or feature we need to pull in, my preference is the second approach. Looking at release notes for fluent-bit [1.8.10](https://fluentbit.io/announcements/v1.8.10/) and [1.8.11](https://fluentbit.io/announcements/v1.8.11/), I don't see anything that seems worth upgrading past the logging-operator default. WDYT?

branden (changes requested):
Upgrade lgtm, but we need to keep the fluent-bit version deployed by this chart in sync with the fluent-bit deployed by logging-operator. Comment inlined below.

services/fluent-bit/0.19.17/fluent-bit.yaml
cwyl02: Checked out the banzaicloud repo and they use 1.8.9 for logging operator 3.17.0. I'm gonna just override to use fluentbit 1.8.9 here.

cwyl02: done

branden (approved):

vespian (approved):

cwyl02:
https://github.com/mesosphere/kommander/pull/1374 passing all checks. This is ready to be merged

PR #167, opened 2022-01-25
fix: Release istio in the workspace namespace

mikolajb:
`substitution vars` points to `kommander` namespace which doesn't exist on the atached cluster.

This changes allows to create a helmrelease in the workspace namespace.


vespian (commented):
Does `substitute` has higher priority than `substituteFrom`? 

My concern is that `substituteFrom` may contain other variables that we may want to keep, either now or in the future.

mikolajb:
> Does `substitute` has higher priority than `substituteFrom`?
> 
> My concern is that `substituteFrom` may contain other variables that we may want to keep, either now or in the future.

@vespian Docs don't say. The order seems random (I added both and they appeared in a reversed order). I don't see we use other variables, it's just `releaseName`. For the record, the value of `substitution-vars` is:
```
$ k get -n foobar configmaps substitution-vars -o yaml
apiVersion: v1
data:
  caIssuerName: kommander-ca
  certificateIssuerKind: ClusterIssuer
  certificateIssuerName: kommander-ca
  releaseNamespace: kommander
  tfaName: traefik-forward-auth
```

vespian:
Why is `releaseNamespace` set to `kommander` on attached clusters instead of workspace namespace?

mikolajb:
> Why is `releaseNamespace` set to `kommander` on attached clusters instead of workspace namespace?

It is actually some kind of a global variable set here: https://github.com/mesosphere/kommander-cli/blob/d161016fe8101bc3fc2bf34205aa639b3e329266/pkg/installer/flux_config.go#L178

vespian:
So maybe this is the problem here? Why is release namespace set on attached clusters to `kommander` instead to workspace namespace for such appdeployments?

I am approving it, with a remark that this approach has a feeling of being just a workaround for a different problem IMO.

vespian (approved):

services/istio/1.9.1/helmrelease.yaml
```diff
@@ -16,6 +16,5 @@ spec:
     name: management
     namespace: kommander-flux
   postBuild:
-    substituteFrom:
-      - kind: ConfigMap
-        name: substitution-vars
+    substitute:
+      releaseNamespace: ${releaseNamespace}
```
gracedo, line 20: are any of the other vars in the cm used? we can still include `substituteFrom` with this change, the `substitute` values would take precedence over the cm

> The var values which are specified in-line with substitute take precedence over the ones in substituteFrom.

https://fluxcd.io/docs/components/kustomize/kustomization/#variable-substitution

makkes (changes requested):
version bump, please.

makkes (approved):

services/istio/1.9.1/helmrelease.yaml
mikolajb: Thanks Grace! I couldn't find this info (about priorities). I will merge it as it is since it worked yesterday 🙂

gracedo:
@makkes @mikolajb Actually until we figure out what to do about versioning, if we decide to stick with our app versions aligning to chart versions we'll need to revert this version bump

gracedo:

PR #175, opened 2022-01-31
fix: promote HTTPS chartmuseum to Traefik

makkes:
We need to let Traefik now that chartmuseum is served over HTTPS and
not HTTP. Since the chart doesn't allow us to directly change the port
name from http to https, we add the annotation here as is documented
at
https://doc.traefik.io/traefik/routing/providers/kubernetes-ingress/#communication-between-traefik-and-pods


vespian (approved):

jimmidyson (approved):
Would be good to add a test to prevent regressions if possible.

PR #178, opened 2022-01-31
[release-2.1][BACKPORT] fix: Release istio in the workspace namespace

mesosphere-mergebot:
This is a backport of the following PR:

https://github.com/mesosphere/kommander-applications/pull/167

`substitution vars` points to `kommander` namespace which doesn't exist on the atached cluster.

This changes allows to create a helmrelease in the workspace namespace.


gracedo (approved):
As noted in https://github.com/mesosphere/kommander-applications/pull/167#issuecomment-1022671039 we may need to revert the version bump (if we do that in main we'll backport it here)

PR #179, opened 2022-02-02
test: add manifest validation tool

floridoo:
Adds a “lint” like tool that validates the Kustomize, Flux, Helm and Kubernetes manifests:

- uses Kustomize and Helm to render all manifests
- validates known Kubernetes resources by parsing them with the Kubernetes deserializer
- checks custom resources by validating them with the CRDs JSON schema

### How to run:
```console
make validate-manifests
```
or directly:
```console
cd hack/validate-manifests
go run . [--verbose=X]
```

### Verbosity levels:
0 = default (progress, errors)
1 = more details (e.g. lists all Helm & Kustomize checks)
2 = even more (lists each validated resource)

JIRA: https://jira.d2iq.com/browse/D2IQ-84293


make/test.mk
```diff
@@ -0,0 +1,3 @@
+.PHONY: test
+test:
+	cd hack/validate-manifests && go run .
```
mhrabovcin, line 3: Can we name the file `validate.mk` and target `validate-manifests` to match the tool name?

hack/validate-manifests/config.go
```diff
@@ -0,0 +1,31 @@
+package main
+
+type Config struct {
+	SkipApplications map[string]bool
```
mhrabovcin, line 4: Why are the `Skip*` config attributes maps? Wouldn't it be easier to have these as slices of strings? What purpose would be to provide `"myapp": false` ?

floridoo: It's much more convenient to look up a string in a map, than in a slice. If it's more clear we could make it a slice, but I would still make a `map[string]anything` out of it on start for faster and more convenient lookup.

mhrabovcin: Gotcha, that makes sense.

hack/validate-manifests/helm.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	client := action.NewInstall(actionConfig)
+	client.DryRun = true
+	client.ReleaseName = helmRelease.Spec.ReleaseName
+	if client.ReleaseName == "" {
+		client.ReleaseName = helmRelease.Name
+	}
+	client.Replace = true // Skip the name check
+	client.ClientOnly = true
+	client.IncludeCRDs = true
```
mhrabovcin, line 76: Should this come from the `HelmRelease` configuration?

```
	// SkipCRDs tells the Helm install action to not install any CRDs. By default,
	// CRDs are installed if not already present.
	//
	// Deprecated use CRD policy (`crds`) attribute with value `Skip` instead.
	//
	// +deprecated
	// +optional
	SkipCRDs bool `json:"skipCRDs,omitempty"`
```

hack/validate-manifests/helm.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+		client.RepoURL = helmRepo.Spec.URL
+		cliConfig := cli.New()
+		chartPath, err := client.LocateChart(chart, cliConfig)
+		if err != nil {
+			return nil, err
+		}
+		return loader.Load(chartPath)
+	case "GitRepository":
+		if helmRelease.Spec.Chart.Spec.SourceRef.Name != "management" {
```
mhrabovcin, line 120: Can we make `management` a constant and comment on why its special?

hack/validate-manifests/helm.go
floridoo: It's not used by any application currently. Do you think we would ever need to install a Helm chart without its CRDs?

floridoo:
Moved to its own repository. This PR is replaced by https://github.com/mesosphere/dkp-bloodhound/pull/4

PR #180, opened 2022-02-02
feat: Upgrade flux to 0.26.0

takirala:
Tested at https://github.com/mesosphere/kommander/pull/1411


takirala:
@cbuto good catch! looks like it has been like that for a while and since we dont override anything in the configmap it had no functional impact so far.
fixed the upgrade script to update the cm.yaml going forward in https://github.com/mesosphere/kommander-applications/pull/180/commits/e1b92e1410f51e05d03d15bc92f0563e0442cefa

gracedo (approved):
CI ✅  on https://github.com/mesosphere/kommander/pull/1411

cbuto (approved):

PR #185, opened 2022-02-07
drop the dependson of root-ca resources and change the issuer type&name

cwyl02:
This is a quick fix of the issue where attached clusters are missing the kommander root-ca bootstrap cert resources. 

https://jira.d2iq.com/browse/D2IQ-84510

kuttl test PR: https://github.com/mesosphere/kommander/pull/1422

### DIFF
--- a/services/cert-manager/0.2.7/root-ca.yaml
+++ b/services/cert-manager/0.2.7/root-ca.yaml
@@ -13,7 +13,7 @@ spec:
     kind: GitRepository
     name: management
     namespace: kommander-flux
-  timeout: 60s
+  timeout: 480s
   # passing releaseNamespace to 2nd level configuration files for able to configure namespace correctly in attached clusters
   # Using `substituteFrom` with `substitution-vars` creates 2nd level resources in `kommander` namespace instead of workspace ns
   postBuild:
--- a/services/cert-manager/0.2.7/root-ca.yaml
+++ b/services/cert-manager/0.2.7/root-ca.yaml
@@ -9,9 +9,6 @@ spec:
   prune: true
   interval: 1m0s
   path: ./services/cert-manager/0.2.7/root-ca
-  dependsOn:
-    - name: cert-manager-release
-      namespace: ${releaseNamespace}
   sourceRef:
     kind: GitRepository
     name: management
--- a/services/cert-manager/0.2.7/root-ca/root-ca.yaml
+++ b/services/cert-manager/0.2.7/root-ca/root-ca.yaml
@@ -4,6 +4,44 @@
 ## follow-up ticket for mirroring: https://jira.d2iq.com/browse/D2IQ-78135
 ---
 apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: kommander-bootstrap-ca-issuer
+  namespace: ${releaseNamespace}
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: kommander-bootstrap-root-certificate
+  namespace: ${releaseNamespace}
+spec:
+  commonName: ca.kommander-bootstrap
+  dnsNames:
+    - ca.kommander-bootstrap
+  duration: 8760h
+  isCA: true
+  issuerRef:
+    name: kommander-bootstrap-ca-issuer
+  secretName: kommander-bootstrap-root-ca
+  subject:
+    organizations:
+      - cert-manager
+---
+apiVersion: cert-manager.io/v1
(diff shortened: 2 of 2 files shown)


services/cert-manager/0.2.7/root-ca/root-ca.yaml
```diff
@@ -4,36 +4,33 @@
...
+  name: selfsigned-issuer
 spec:
   selfSigned: {}
 ---
 apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
-  name: kommander-bootstrap-root-certificate
-  namespace: ${releaseNamespace}
+  name: kommander-ca
```
cbuto, line 16: on my local cluster, I see this Certificate and ClusterIssuer already exist (not sure who creates them). Are these only created on attached clusters?

services/cert-manager/0.2.7/root-ca.yaml
```diff
@@ -9,9 +9,6 @@ spec:
   prune: true
   interval: 1m0s
   path: ./services/cert-manager/0.2.7/root-ca
-  dependsOn:
-    - name: cert-manager-release
-      namespace: ${releaseNamespace}
```
makkes, line 14: How does removing this dependency help here? The cert-manager-release will be deployed, anyway?!

makkes (commented):
I'm not sure if this is a viable way forward. Would this still work with custom certificates as described [here](https://docs.d2iq.com/dkp/kommander/2.1/install/configuration/custom-domain/#configure-a-custom-certificate)?

IMHO the better way would be to do this in code, fetching the resources that would need to be created on attached from the management cluster.

makkes:
@cwyl02 you might also want to discuss with @mhrabovcin on this as he's working on a related spike (https://jira.d2iq.com/browse/D2IQ-84025).

services/cert-manager/0.2.7/root-ca/root-ca.yaml
cwyl02: The resources in this files will only be created in attached clusters.

On the management host, these 3 resources are already created by [kommander-cli](https://github.com/mesosphere/kommander-cli/tree/main/pkg/installer/manifests/rootca).

services/cert-manager/0.2.7/root-ca.yaml
cwyl02: Currently on the attached clusters that comes with cert-manager(e.g., konvoy2 clusters), the cert-manager-release will fail and block the creation of the root-ca resources.

And yes, the cert-manager-release will be deployed and show as failed anyways and it has been like that. Figuring out how to determine whether cert-manager exists on a cluster is something we want to do but not necessarily just to get through the current state.

cwyl02:
@makkes This is definitely not an ideal solution but it is a quick solution that at least makes the attached clusters(with or without cert-manager installed) really work.

Currently, even on attached clusters without cert-manager, apps like kube-oidc-proxy and kommander-traefik's certs won't work, since the bootstrap root-ca resources are created with a name that the app certs are not expecting. 

On attached clusters with cert-manager installed, the root-ca resources are never created due to the dependsOn, thus kube-oidc-proxy and kommander-traefik's certs don't work either.

For the custom certificate part, great call on that. Let me test that one out.

services/cert-manager/0.2.7/root-ca.yaml
aweris: However, if cluster misses the cert-manager and we drop this depends, customization will try to apply nonexistent CRD. That's why this dependency exists

cwyl02:
attached cluster with custom cert doesn't work but the failure occurs at cloning the management git repo so is not related to this change. unblocking the PR.

services/cert-manager/0.2.7/root-ca.yaml
takirala: if we are going down this path, lets increase the `timeout` and `interval` to a higher value just in case.

takirala (approved):

services/cert-manager/0.2.7/root-ca.yaml
cwyl02: @takirala How about bump to the `timeout` to 300s and keep the `interval` as-is?

cwyl02: ended up changing to 480s, since cert-manager retries 30 times with 15s interval.

cbuto (approved):

mhrabovcin (approved):

takirala (approved):

cwyl02:

PR #190, opened 2022-02-10
feat: provide defaults for new kommander chart values

makkes:
This commit provides defaults for the new values in the kommander
chart that were introduced in
https://github.com/mesosphere/kommander/pull/1438.


makkes:
pre-commit is failing on commits unrelated to this PR. I don't know why that is, though. IMHO this is safe to ignore.

cwyl02 (approved):

PR #195, opened 2022-02-15
fix: Disable grafana datasource reload url

gracedo:
https://jira.d2iq.com/browse/D2IQ-84982

Newer versions of grafana added automatic reloading of datasources by hitting the Admin API. However, the API only works with basic authentication, which we disable.
https://grafana.com/docs/grafana/latest/http_api/admin

Upstream PR fix: https://github.com/grafana/helm-charts/pull/1046 we can bump to this once it is merged

### DIFF
--- a/services/centralized-grafana/32.0.2/defaults/cm.yaml
+++ b/services/centralized-grafana/32.0.2/defaults/cm.yaml
@@ -34,6 +34,7 @@ data:
           # label that the configmaps with datasources are marked with
           label: grafana_datasource_kommander
           searchNamespace: ALL
+          skipReload: true
       annotations:
         configmap.reloader.stakater.com/reload: "centralized-kubecost-grafana-datasource"
       ingress:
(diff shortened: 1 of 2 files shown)


TattdCodeMonkey (approved):

gracedo:
Tested this by manually applying the changes to the daily.. centralized-grafana is all good but that's because we have a reloader annotation to reload the pod when the kubecost datasource is modified (this was our workaround previously to get around this issue). But cluster's prometheus is now failing to pul in any datasources. Looking into this now.

gracedo:
Manually adding back the initcontainer that was removed in https://github.com/grafana/helm-charts/commit/ee0914e19e680a6cde7a5a4e9bcb43f2bef9517a - it was moved into a container utilizing reloads, which we can't take advantage of due to the API only working with basic auth. I'll look into contributing something upstream to add the initcontainer back but until then just adding it here manually. Tested on daily cluster, datasources are back ✅

takirala (approved):

PR #197, opened 2022-02-15
feat: bump cert-manager to v1.7.1

mhrabovcin:
kommander tests: https://github.com/mesosphere/kommander/pull/1446

JIRA: https://jira.d2iq.com/browse/D2IQ-84025

This is a proposal on how to replace the exiting `cert-manager-setup` helm chart https://github.com/mesosphere/charts/blob/master/stable/cert-manager-setup/README.md with newly proposed `cert-manager-release` that would only bundle upstream CRDs and official helm chart without any custom templates.

This will allow FluxCD to manage CRDs and gives us control on how to behave on upgrades between versions by setting `CRDsPolicy` on `install` and `upgrade` for the `cert-manager` `HelmRelease`.

More on how flux manages CRDs: https://fluxcd.io/docs/components/helm/helmreleases/

```
	// CRDs upgrade CRDs from the Helm Chart's crds directory according
	// to the CRD upgrade policy provided here. Valid values are `Skip`,
	// `Create` or `CreateReplace`. Default is `Create` and if omitted
	// CRDs are installed but not updated.
	//
	// Skip: do neither install nor replace (update) any CRDs.
	//
	// Create: new CRDs are created, existing CRDs are neither updated nor deleted.
	//
	// CreateReplace: new CRDs are created, existing CRDs are updated (replaced)
	// but not deleted.
	//
	// By default, CRDs are applied (installed) during Helm install action.
	// With this option users can opt-in to CRD replace existing CRDs on Helm
	// install actions, which is not (yet) natively supported by Helm.
	// https://helm.sh/docs/chart_best_practices/custom_resource_definitions.
	//
	// +kubebuilder:validation:Enum=Skip;Create;CreateReplace
	// +optional
	CRDs CRDsPolicy `json:"crds,omitempty"`
```

This is just a proposal / spike PR. Once I get preliminary approval I'll update proper versions and create `cert-manager-release` chart in `mesosphere/charts` repo.


makkes:
@mesosphere-mergebot test all

charts/cert-manager-release/values.yaml
```diff
@@ -0,0 +1,9 @@
+# Default values for cert-manager-release.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+cert-manager:
+  enabled: true
+
+  # Let flux handle CRDs installation and management
+  installCRDs: false
```
jimmidyson, line 9: What is the benefit of using flux for this?

jimmidyson:
Couldn't we instead use the CRDs without a helm chart applied via a kustomization, and then a second kustomization (dependent on the previous one) to apply the helm release using the upstream chart?

charts/cert-manager-release/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
```
jimmidyson, line 1: Why do we have a chart in this repo?

mhrabovcin: > This is just a proposal / spike PR. Once I get preliminary approval I'll update proper versions and create cert-manager-release chart in mesosphere/charts repo.

mhrabovcin:
@jimmidyson I didn't find anywhere in the flux documetation whether it can act on CRDs that are coming from kustomization VS coming from helm chart `crds` directory, which is a special place from helm perspective, where to put installed CRDs.

charts/cert-manager-release/Chart.yaml
jimmidyson: Oops sorry!

charts/cert-manager-release/Chart.yaml
```diff
@@ -0,0 +1,11 @@
...
+name: cert-manager-release
+description: A chart that bundles cert-manager CRDs with cert-manager helm chart
+type: application
+version: v1.7.1
+appVersion: v1.7.1
+
+dependencies:
+  - name: cert-manager
+    version: v1.7.1
+    repository: https://charts.jetstack.io
```
makkes, line 11: What I generally don't like about Helm chart dependencies is that they are basically invisible from outside (e.g. `helm ls` doesn't show them). Can we not make this a mere Kustomization that installs a cert-manager HelmRelease and then have another HelmRelease depend on it? That's what we do with [kubefed](https://github.com/mesosphere/kommander-applications/tree/20ca18fb41f6742c9013f88fe5bef42c1da4030d/services/kubefed/0.9.1), for example.

mhrabovcin: How does the linked kubefed install 2 depending helmreleases?

mhrabovcin:
The discussion regarding the question of using custom chart with CRDs was in slack - https://mesosphere.slack.com/archives/C0152RGV2KU/p1645097061665569

The decision was:
> I'm very much convinced that we shouldn't put the burden of CRD management onto ourselves beyond using the API machinery that Flux provides. The rest is up to chart authors.
https://mesosphere.slack.com/archives/C0152RGV2KU/p1645105855208669?thread_ts=1645097061.665569&cid=C0152RGV2KU

I think we can close this PR and use only the upstream chart with `installCRDs` value set to `true`.

mhrabovcin:
The change will be handled in https://github.com/mesosphere/kommander-applications/pull/201 

Closing this spike PR.

charts/cert-manager-release/Chart.yaml
mhrabovcin: This will be handled in new @mikolajb 's PR and he will try to not embed charts.

PR #198, opened 2022-02-16
test: add manifest validation

floridoo:
Adds a “lint” job using `dkp-bloodhound` that validates the Kustomize, Flux, Helm and Kubernetes manifests:
- uses Kustomize and Helm to render all manifests
- validates known Kubernetes resources by parsing them with the Kubernetes deserializer
- checks custom resources by validating them with the CRDs JSON schema

A few apps don't pass validation:
- `kaptain`: HelmRepository not included in `common/helm-repositories/kustomization.yaml`
- `grafana-loki`, `project-grafana-loki`, `project-logging`: they depend on other apps, but don't declare it (resulting in some CRs not being recognized, because the CRD is not loaded yet)

For now, these apps are whitelisted to let the validation pass. In the future these should be addressed by either changing the manifests  or changing the way the validation is done.

### How to run:
```console
make validate-manifests
```


mhrabovcin (approved):

makkes (approved):
quick feedback from observation: I just ran the new target and "Checking Flux resources" takes a lot of time without it making any visible progress. I suppose that's more of a feature request for bloodhound itself, though.

floridoo:
> quick feedback from observation: I just ran the new target and "Checking Flux resources" takes a lot of time without it making any visible progress. I suppose that's more of a feature request for bloodhound itself, though.

The different Flux checks (HelmReleases, Kustomizations) run in parallel, which makes it harder to have a meaningful output before completion. There are some improvements possible though.
The whole execution took 15s on my machine, so I didn't mind (cloud machine, so the fast network might help here). I am curious: how long did the validation take for you?

PR #218, opened 2022-02-28
feat: Bump grafana and kps to latest

gracedo:
Bump grafana to latest (6.22.0) and kps to latest which also bumps the grafana subchart to 6.22.0 so we can align grafana versions as well to use the same images. Latest grafana also includes the [bug fix I upstreamed](https://github.com/grafana/helm-charts/pull/1046) so we can remove the workaround now.

While testing these bumps, noticed a few changes to account for - dashboard discovery labels must have value `"1"` now, and I also noticed that there is a `dashboards.default_home_dashboard_path` grafana.ini configuration option now that we can use instead of our cronjobs, so I've disabled those cronjobs (🥳 ).

https://jira.d2iq.com/browse/D2IQ-77877

Also opened up charts pr https://github.com/mesosphere/charts/pull/1294 to disable cronjob by default there as well. If that lands first I'll bump the chart here.


cbuto (approved):

cwyl02 (approved):

gracedo:
https://github.com/mesosphere/kommander/pull/1485 ✅

PR #221, opened 2022-03-01
feat: Add default images and charts tags

mikolajb:
Jira: https://jira.d2iq.com/browse/D2IQ-85626


services/kommander-appmanagement/0.2.0/defaults/cm.yaml
```diff
@@ -11,6 +11,6 @@ data:
         manager:
           replicas: "${kommanderAppManagementReplicas}"
           image:
-            tag: "${appManagementImageTag}"
+            tag: "${appManagementImageTag:=v2.2.0-dev-amd64}"
```
vespian, line 14: As discussed, we need to include the outcomes of https://mesosphere.slack.com/archives/C02Q9AP62NN/p1646213450807939 here as we need to introduce new variables. Old ones can't be changed in 2.2.

vespian (approved):

services/kommander-appmanagement/0.2.0/defaults/cm.yaml
mhrabovcin, line 14: Will we need to change all these values once we release `v2.2.0` and the next development version will be `v2.3.0` ?

mikolajb: [Yes, exactly](https://jira.d2iq.com/browse/D2IQ-85575): 

> the tag above will become the default version used by k-cli
once we release v2.2.0 , we bump this tag to v2.3.0-dev

jimmidyson: The tags are defaulted in the various templates right now to use `Chart.version` if unspecified. I don't changing this is necessary, especially with the amd64 suffix. I think I can extract from the mindthegap pr the removal of the amd64 suffixes and no changes will be required here. This will also future proof us for when mindthegap is introduced.

jimmidyson (changes requested):

makkes (approved):

PR #223, opened 2022-03-02
chore: adds spec.wait to Kustomizations

farhan5900:
Resolves https://jira.d2iq.com/browse/D2IQ-85656

### DIFF
--- a/services/centralized-kubecost/0.22.1/post-install-jobs.yaml
+++ b/services/centralized-kubecost/0.22.1/post-install-jobs.yaml
@@ -6,6 +6,7 @@ metadata:
 spec:
   force: false
   prune: true
+  wait: true
   interval: 1m0s
   path: ./services/centralized-kubecost/0.22.1/post-install-jobs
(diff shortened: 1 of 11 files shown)


makkes (commented):
I'm not sure this list is exhaustive. We would need to add the field to the controller creating the Kustomizations from AppDeployments, too, right?

Here: https://github.com/mesosphere/kommander/blob/f3ce175023bd99d6945ebf6d30d91d48f4bc1d26/federation/pkg/controllers/appdeployment_controller.go#L767

Not sure if there's more controllers that generate Kustomizations, e.g. for projects or catalog apps.

vespian:
>We would need to add the field to the controller creating the Kustomizations from AppDeployments, too, right?

This^^

farhan5900:
Here is the PR addressing that: https://github.com/mesosphere/kommander/pull/1494

vespian (approved):

gracedo (approved):

PR #227, opened 2022-03-03
feat: Add pre-upgrade strategy for jaeger

gracedo:
https://jira.d2iq.com/browse/D2IQ-85305
testing in https://github.com/mesosphere/kommander/pull/1501

A new selector label was added in 1.21.4 which caused backwards-incompatibility with existing deployments. The recommended action is to delete the deployment before upgrading https://github.com/jaegertracing/helm-charts/issues/271. I've added a "pre-upgrade" step (job backed by flux kustomization) to delete the deployment that the helmrelease flux kustomization (newly created) now depends on.

Due to the HR moving behind a new flux kustomization, we now also see a race when the `jaegar` top-level kustomization updates to point to the new service version path, where the HR no longer exists and the HR flux kustomization is not yet applied. Since we are already deleting the deployment, however, I will accept this race for now, as a "reinstall" strategy works just as well as the deletion of the deployment.

I also added the `istio-helmrelease` flux kustomization back (v2.1.1 also had this flux kustomization)

In hindsight, we probably should have created flux kustomizations for all helmreleases in these services to allow for more flexibility, but given that it's acceptable for jaeger to reinstall, I'll punt that for a later time.

But just to note our thoughts around this, if we need to resolve this now in 2.2, we discussed a one-time patch of the 2.1.1 k-apps services in the bootstrap gitrepo, to add flux kustomizations for all the HRs, and update the top-level kustomizations (created by appdeployment controller) to set `prune` to `false` to avoid deleting HRs in case there is a race or deps between the new flux kustomizations potentially causing uninstall, then setting `prune` back to `true` after the patch of services. This would also involve adding flux kustomizations behind all HRs in this 2.2 version of k-apps for all `services/` (this change is straightforward). But if we can delay this to post-2.2, let's delay it (😢). Added a ticket https://jira.d2iq.com/browse/D2IQ-85865 to follow-up on this after the release


services/istio/1.11.5/helmrelease.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
+kind: Kustomization
+metadata:
+  # jaeger-pre-upgrade kustomization depends on this istio-helmrelease kustomization.
+  # if this name is updated, the dependsOn must be bumped in ./services/jaeger/2.28.0/pre-upgrade.yaml
```
takirala, line 5:
```suggestion
  # if this name is updated, the dependsOn must be bumped in ./services/jaeger/2.29.0/pre-upgrade.yaml
```

takirala (approved):

cbuto (approved):

cwyl02 (approved):

PR #231, opened 2022-03-04
fix: Use label selector for jaeger pre-upgrade

gracedo:
Better to use label selectors to select for the deployment to delete in pre-upgrade, to avoid deleting the newer deployment in any races. Newer jaeger versions added a label to the deployment that wasn't in the previous version, so using that to select against newer deployments.

### DIFF
--- a/services/jaeger/2.29.0/pre-upgrade/delete-jaeger-deployment.yaml
+++ b/services/jaeger/2.29.0/pre-upgrade/delete-jaeger-deployment.yaml
@@ -1,6 +1,10 @@
 # For upgrades from <2.21.4
-# The job will also be run pre-install but the kubectl --ignore-not-found flag should
-# protect against failing on nonexistent deployments (if jaeger is not installed yet)
+# https://github.com/jaegertracing/helm-charts/pull/257/files#diff-d9ac456c418c74b7ba25a05ec2cc077cd8dd45ddb7a3008c2dee25e4d347a23eR48
+# New "app.kubernetes.io/instance" label selector was added to the jaeger
+# deployment; here we select for deployments with the "app.kubernetes.io/name"
+# label but without the newly added "app.kubernetes.io/instance" label, to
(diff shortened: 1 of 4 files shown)


cbuto (approved):

jimmidyson (approved):

services/jaeger/2.29.0/pre-upgrade/delete-jaeger-deployment.yaml
```diff
@@ -51,4 +51,4 @@ spec:
             - sh
             - -c
             - |
-              kubectl delete deployment jaeger-jaeger-operator -n istio-system --ignore-not-found
+              kubectl delete deployments.apps -l app.kubernetes.io/name=jaeger-operator,app.kubernetes.io/instance!=jaeger-operator --cascade=orphan -n istio-system --ignore-not-found
```
takirala, line 54: is this label present in older version of jaeger and NOT in newer version of jaeger?
can we add a comment on how/where this label comes from? (Thinking of the next jaeger bump in future).

takirala (approved):

services/jaeger/2.29.0/pre-upgrade/delete-jaeger-deployment.yaml
gracedo: https://github.com/jaegertracing/helm-charts/pull/257/files#diff-d9ac456c418c74b7ba25a05ec2cc077cd8dd45ddb7a3008c2dee25e4d347a23eR48 `app.kubernetes.io/instance` label was added in newer version, so selecting against here. I'll add a comment (I was thinking we would probably be removing - or updating - 2.2 upgrade strategies in the next 2.3 release?)

takirala: > I was thinking we would probably be removing - or updating - 2.2 upgrade strategies in the next 2.3 release?)

:crossed_fingers:

PR #235, opened 2022-03-08
refactor: rename kommander version variables

vespian:
This is a followup for following PRs:
* https://github.com/mesosphere/kommander-cli/pull/381
* https://github.com/mesosphere/kommander-applications/pull/228
* https://github.com/mesosphere/kommander-applications/pull/221

There were two problems that this PR solves:
* we need to define a default version for kommander-licensing subchart as it does not define sane defaults by itself. Cleaning it up is beyond the scope of the UniversalUpdater work.
* the substitute variable names in kommander v2.2.0 need to be different than the ones in kommander v2.1.0

The second bulletpoint requires some clarification - in v2.2.0 we are changing the way we supply the default version of kommander and kommander-appmanagement charts. 

Instead of storing it in k-cli, we moved it to k-apps repo. The substitute variables we were using so far may contain version information that is incompatible with v2.2.0 and we can't change them before upgrading the appdeployment as this will cause Flux to start rolling out changes in kommander v2.1.0 appdeployment as the version of the chart in this version is taken from these substitute variables. Changing the names of these vars and relying on their defaults solves this.

### DIFF
--- a/services/kommander/0.2.0/defaults/cm.yaml
+++ b/services/kommander/0.2.0/defaults/cm.yaml
@@ -56,11 +56,11 @@ data:
           manager:
             rootCertSecretName: ${caSecretName}
             image:
-              tag: ${kommanderLicensingControllerManagerImageTag:=v2.2.0-dev}
+              tag: ${kommanderLicensingControllerManagerImageTag}
               repository: ${kommanderLicensingControllerManagerImageRepository}
       webhook:
         image:
-          tag: ${kommanderLicensingControllerWebhookImageTag:=v2.2.0-dev}
+          tag: ${kommanderLicensingControllerWebhookImageTag}
           repository: ${kommanderLicensingControllerWebhookImageRepository}
       defaultEnterpriseApps:
         centralized-kubecost: "0.22.1"
--- a/services/kommander/0.2.0/defaults/cm.yaml
+++ b/services/kommander/0.2.0/defaults/cm.yaml
@@ -10,8 +10,8 @@ data:
       enabled: ${airgappedEnabled}
     authorizedlister:
       image:
-        tag: ${authorizedlisterImageTag}
-        repository: ${authorizedlisterImageRepository}
+        tag: ${kommanderAuthorizedlisterImageTag}
+        repository: ${kommanderAuthorizedlisterImageRepository}
     certificates:
       ca:
         issuer:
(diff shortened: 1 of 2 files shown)


mikolajb (approved):

services/kommander/0.2.0/defaults/cm.yaml
```diff
@@ -56,12 +56,12 @@ data:
...
             image:
-              tag: ${licensingControllerManagerImageTag}
-              repository: ${licensingControllerManagerImageRepository}
+              tag: ${kommanderLicensingControllerManagerImageTag:=v2.2.0-dev}
+              repository: ${kommanderLicensingControllerManagerImageRepository}
       webhook:
         image:
-          tag: ${licensingControllerWebhookImageTag}
-          repository: ${licensingControllerWebhookImageRepository}
+          tag: ${kommanderLicensingControllerWebhookImageTag:=v2.2.0-dev}
```
jimmidyson, line 63: This is fixed in https://github.com/mesosphere/kommander/pull/1512.

services/kommander/0.2.0/defaults/cm.yaml
```diff
@@ -56,12 +56,12 @@ data:
           manager:
             rootCertSecretName: ${caSecretName}
             image:
-              tag: ${licensingControllerManagerImageTag}
-              repository: ${licensingControllerManagerImageRepository}
+              tag: ${kommanderLicensingControllerManagerImageTag:=v2.2.0-dev}
```
jimmidyson, line 59: This is fixed in https://github.com/mesosphere/kommander/pull/1512.

jimmidyson (changes requested):
Let's drop the licensing image changes please as they are not required once https://github.com/mesosphere/kommander/pull/1512 Is merged.

services/kommander/0.2.0/defaults/cm.yaml
vespian: I double-checked the PR you linked and I do not understand how it solves the issue I described in PR descrption:

```
we need to define a default version for kommander-licensing subchart as it does not define sane defaults by itself. Cleaning it up is beyond the scope of the UniversalUpdater work.
```
The code in question is here:

https://github.com/mesosphere/kommander/blob/aa264652835ce100995e53cf7e338ffae59e6e1f/chart/kommander/charts/kommander-licensing/values.yaml#L17
https://github.com/mesosphere/kommander/blob/aa264652835ce100995e53cf7e338ffae59e6e1f/chart/kommander/charts/kommander-licensing/values.yaml#L46

and the changes in the PR you pointed are basically doing:
```diff
-appVersion: "dev"
+appVersion: v0.0.0-dev
description: Kommander Licensing
-version: 0.2.0-dev
+version: v0.0.0-dev
```

Also, please note that we need to make the image tags overridable, hence we can't simply change appVersion to `v2.2.0-dev`. We probably could clean this up in a better way but we are already pushing the scope of the `UniversalMigrator` change and this work is blocking other work. Let's please merge what we have and iterate/prioritize remaining work.

services/kommander/0.2.0/defaults/cm.yaml
vespian: See my comment above.

services/kommander/0.2.0/defaults/cm.yaml
jimmidyson: The chart packaging changes the version of the commander-licensing subchart when it is packaged (see https://github.com/mesosphere/kommander/pull/1512/commits/33ab8ae6dc9999c3374368c73b63ce235b4d3aa1). This means that the defaulting of the image tag in kommander-licensing works properly.

jimmidyson: So the image tags in the kommander-licensing chart use the chart version, which now matches the parent kommander chart version by virtue of how we package the chart. You can still override the tags by providing relevant values, but generally this is not necessary, the same as it is not necessary for the kommander chart.

services/kommander/0.2.0/defaults/cm.yaml
jimmidyson: I've removed this now.

services/kommander/0.2.0/defaults/cm.yaml
jimmidyson: I've removed this now.

jimmidyson (approved):

jimmidyson:
@mesosphere-mergebot test all

PR #237, opened 2022-03-11
fix: Restore grafana metrics

branden:
https://jira.d2iq.com/browse/D2IQ-84979

This restores grafana metrics. Without this change, the default ServiceMonitor selector omits the grafana ServiceMonitor, and its metrics are not scraped.

### DIFF
--- a/services/kube-prometheus-stack/33.1.1/defaults/cm.yaml
+++ b/services/kube-prometheus-stack/33.1.1/defaults/cm.yaml
@@ -39,6 +39,8 @@ data:
         # **NOTE** Any changes here need to be copied to kube-prometheus-stack-overrides.yaml
         # This is because arrays in values are replaced, not appended.
         - name: dkp-service-monitor-metrics
+          additionalLabels:
+            prometheus.kommander.d2iq.io/select: true
           selector:
             matchLabels:
               servicemonitor.kommander.mesosphere.io/path: "metrics"
@@ -53,6 +55,8 @@ data:
             - targetPort: 7979
               interval: 30s
         - name: dkp-service-monitor-metrics-http
+          additionalLabels:
+            prometheus.kommander.d2iq.io/select: true
           selector:
             matchLabels:
               servicemonitor.kommander.mesosphere.io/path: "metrics"
@@ -64,6 +68,8 @@ data:
             - targetPort: http
               interval: 30s
         - name: dkp-service-monitor-api-v1-metrics-prometheus
+          additionalLabels:
+            prometheus.kommander.d2iq.io/select: true
           selector:
             matchLabels:
               servicemonitor.kommander.mesosphere.io/path: "api__v1__metrics__prometheus"
@@ -74,6 +80,8 @@ data:
               port: metrics
               interval: 30s
         - name: dkp-service-monitor-api-v1-metrics-prometheus-http-10s
+          additionalLabels:
+            prometheus.kommander.d2iq.io/select: true
           selector:
             matchLabels:
               servicemonitor.kommander.mesosphere.io/path: "api__v1__metrics__prometheus"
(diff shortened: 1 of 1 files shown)


gracedo (commented):
Innnnnteresting. Looked into how servicemonitors are selected a bit more and found 

> By default, Prometheus discovers PodMonitors and ServiceMonitors within its namespace, that are labeled with the same release tag as the prometheus-operator release. Sometimes, you may need to discover custom PodMonitors/ServiceMonitors, for example used to scrape data from third-party applications. An easy way of doing this, without compromising the default PodMonitors/ServiceMonitors discovery, is allowing Prometheus to discover all PodMonitors/ServiceMonitors within its namespace, without applying label filtering. To do so, you can set prometheus.prometheusSpec.podMonitorSelectorNilUsesHelmValues and prometheus.prometheusSpec.serviceMonitorSelectorNilUsesHelmValues to false.

https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#prometheusioscrape

> set prometheus.prometheusSpec.podMonitorSelectorNilUsesHelmValues and prometheus.prometheusSpec.serviceMonitorSelectorNilUsesHelmValues to false

Looking at the last sentence, maybe that's what we should do anyways. I don't think we would want to filter out any servicemonitors, and I recall in the past there were some hiccups getting other apps' servicemonitors pulled in due to that `release` label. It would be nice to not have to remember to include that label if we were to define servicemonitors in the context of other apps (vs defining them out of this chart, which automatically adds that label).

^ as a matter of fact, I recall this happening in the istio chart and sure enough, I just took a look and it's still using an outdated `release` label in its [servicemonitors](https://github.com/mesosphere/charts/blob/master/staging/istio/charts/prometheus-operator/templates/servicemonitors.yaml#L9) (pointing to `prometheus-kubeaddons`, from 1.x). If we make the config change to select all service monitors, we can just remove that label from the istio chart (otherwise we need to update it).

That said, the missed grafana servicemonitor issue is still a bug in the upstream chart that should be fixed. KPS recently made a change to use the subchart's servicemonitor directly instead of defining their own in KPS chart, so I imagine this side-effect was missed/unintended.

gracedo:
I just tested updating the `Prometheus` CR on the daily to set `serviceMonitorSelector: {}` and saw it pull in the missing service monitors including istio's ✅

cbuto (approved):

takirala (dismissed):

branden:
One reason we may want to filter servicemonitors is to make sure we aren't pulling metrics from workloads that haven't opted in to it. Prometheus-operator currently monitors all namespaces, so removing the helm label selector would mean we're picking up every servicemonitor on the cluster. I could imagine a COPS ticket where prometheus is full/slow/down/etc. because the customer declared a bunch of servicemonitors for a different prometheus instance without realizing that the DKP prometheus was picking them up.

I propose that we add our own label to our servicemonitors and filter based on it, to prevent inadvertently ingesting metrics that we haven't tuned our prometheus deployment to handle. If a customer wants their metrics scraped by the platform prometheus, they can apply the label. I'll take a look at what it takes to label the servicemonitors outside kube-prometheus-stack, but reaching parity with what we have now will be easy. LMK what you think about that approach.

gracedo:
I can go either way - I'm not sure what the common customer use case is with deploying separate prometheus stacks, but do see how scraping all servicemonitors can be problematic/unexpected for the customer. I was going to say we should include that selector label in our docs but it looks like we already do 😅 https://docs.d2iq.com/dkp/kommander/2.1/monitoring/#monitor-applications so since it looks like this is indeed the intention to select servicemonitors on a label vs all, we can stick with it. Doing an audit of where we may not be using that label in other apps would be great 🙏🏻

gracedo (approved):

branden:
I pushed a change that selects pod and service monitors by a custom label. I also made it explicit in the config that we search all namespaces.

I searched through kommander-applications and mesosphere/charts and found no existing examples of anything setting the label `release: kube-prometheus-stack`. I would've added the label for istio's servicemonitor but that will require a change to its chart, so I'll leave it for another PR.

cbuto (approved):

PR #238, opened 2022-03-11
feat: Add kubecost as a default enterprise app

cbuto:
https://jira.d2iq.com/browse/D2IQ-85872

Kubecost will now be considered an enterprise app (and will be removed from the default set of apps)

### DIFF
--- a/services/kommander/0.2.0/defaults/cm.yaml
+++ b/services/kommander/0.2.0/defaults/cm.yaml
@@ -64,6 +64,8 @@ data:
           repository: ${kommanderLicensingControllerWebhookImageRepository}
       defaultEnterpriseApps:
         centralized-kubecost: "0.22.1"
+        kubecost: "0.22.1"
+        kubecost-thanos-traefik: "0.0.1"
         centralized-grafana: "33.1.1"
         karma: "2.0.1"
         karma-traefik: "0.0.1"


services/kommander/0.2.0/defaults/cm.yaml
```diff
@@ -64,6 +64,8 @@ data:
           repository: ${kommanderLicensingControllerWebhookImageRepository}
       defaultEnterpriseApps:
         centralized-kubecost: "0.22.1"
+        kubecost: "0.22.1"
```
takirala, line 67: Is having `kubecost` ins `defaultApps` few lines below https://github.com/mesosphere/kommander-applications/pull/238/files#diff-50b1ab93c1d40ed8185f0a8ff4d8f44b3b348f75ca3c43d9b8d069d24d9e2782L91 still fine ?

gracedo: I _think_ this should be fine for the mgmt cluster because we set that "default apps deployed" label to true on kommander workspace so the default apps controller ignores deploying those default apps altogether to the mgmt cluster (since we use cli to deploy those, not controller)

gracedo (approved):

cwyl02 (approved):

services/kommander/0.2.0/defaults/cm.yaml
takirala: so the field is nested as
```
attached:
      prerequisites:
        defaultApps
```
so its fine to deploy `kubecost` to attached clusters ?!

takirala: <discussed offline> this is fine as multi cluster is an enterprise only feature.

takirala (approved):

PR #239, opened 2022-03-11
fix: grafana dashboard view only

cwyl02:
https://jira.d2iq.com/browse/D2IQ-84281

charts repo PR #1: https://github.com/mesosphere/charts/pull/1299
charts repo PR #2: https://github.com/mesosphere/charts/pull/1302
kommander PR: https://github.com/mesosphere/kommander/pull/1522


services/centralized-grafana/33.1.3/centralized-grafana.yaml
```diff
@@ -38,6 +38,12 @@ rules:
     verbs:
       - get
       - head
+  - nonResourceURLs:
```
gracedo, line 41: 🤔 can you double check why we define these roles separately here vs using the ones created in the chart? (It was very likely me who did this, I can’t recall if there was a reason why or if it was done mistakenly)

cwyl02: good catch! I took a look at the values of centralized and found: https://github.com/mesosphere/kommander-applications/blob/main/services/centralized-grafana/33.1.1/defaults/cm.yaml#L13
Let me try flipping that to true..

services/centralized-grafana/33.1.3/defaults/cm.yaml
```diff
@@ -2,15 +2,16 @@
...
   namespace: ${releaseNamespace}
 data:
   values.yaml: |
     ---
+    fullnameOverride: "centralized-grafana"
     mesosphereResources:
       create: false
       ingressRBAC:
-        enabled: false
+        enabled: true
```
cwyl02, line 14: FYI dkp-centralized-grafana-admin/view/edit will become

centralized-grafana-dkp-grafana-admin/view/edit

@gracedo

branden (approved):

cbuto (approved):

services/centralized-grafana/33.1.3/defaults/cm.yaml
gracedo: I think that’s fine as long as we aren’t referencing the clusterroles anywhere (I don’t think so? But maybe worth searching docs)

gracedo (approved):

services/centralized-grafana/33.1.3/defaults/cm.yaml
cwyl02: oops, forgot about the docs. They are actually mentioned in docs: https://docs.d2iq.com/dkp/kommander/2.1/operations/access-control/rbac/

I vote for using the new name but also open to change it back. I will open a docs PR if the new name is OK

gracedo: Thanks for finding that page! It actually looks like the kps role names might be the odd ones out as all others are dkp- prefixed. I wonder if we instead want to modify the kps chart clusterrole templates to use the prefix? So instead of kube-prometheus-stack-dkp-alertmanager-view, we could template it to dkp-kube-prometheus-stack-alertmanager-view instead? (dkp-<release name>-…)

cwyl02: opened docs PR anyways for now: https://github.com/mesosphere/dcos-docs-site/pull/4180.
We can still rollback tomorrow to the old way if we determined that's the way to go.

PR #250, opened 2022-03-21
feat: Bump kubecost to 0.23.1

gracedo:
bump kubecost to latest patch version
testing in https://github.com/mesosphere/kommander/pull/1556


makkes (approved):

mhrabovcin:
Can we remove this custom ingress and turn on the one that is in the chart https://github.com/mesosphere/kommander-applications/blob/main/services/kubecost/0.23.0/kubecost.yaml#L87-L115

gracedo:
> Can we remove this custom ingress and turn on the one that is in the chart https://github.com/mesosphere/kommander-applications/blob/main/services/kubecost/0.23.0/kubecost.yaml#L87-L115

@mhrabovcin I don't think so :/ it doesn't look like they've updated the subcharts so I think it's still an issue

mhrabovcin (approved):

PR #254, opened 2022-03-24
fix: Restore missing metrics scrape targets

branden:
This restores some missing scrape targets by applying a label to their servicemonitors that the servicemonitor selector matches against.


cwyl02 (approved):

gracedo (approved):

takirala (approved):

branden:
I tested this by deploying kommander to a kind cluster. When I open the prometheus targets page I see targets for centralized-grafana, kube-state-metrics, and node-exporter, and all are up. I think we're good to merge!

PR #279, opened 2022-03-30
feat: Bump gitea to 5.0.4

mikolajb:
Bumping gitea to satisfy CVEs


services/gitea/5.0.4/defaults/cm.yaml
```diff
@@ -52,3 +53,9 @@ data:
       readOnly: true
       mountPath: "/git-tls"
     clusterDomain: cluster.local.
+    memcached:
+      image:
+        tag: 1.6.15-debian-10-r0
+    postgresql:
+      image:
+        tag: 10.20.0-debian-10-r48
```
cbuto, line 61: do we want to stick with the same Postgres version if possible? 

I.e 11.11.0-debian-10-r62 which is Postgres 11 I think?

mikolajb: Sounds good, thanks @cbuto !

cwyl02 (approved):

cbuto:
CI is ✅ 🎉

mikolajb:
> @mikolajb can you kick off the upgrade e2e tests before merging this, please? We need to make sure upgrade works with this new version.

@makkes I tested the upgrades locally

PR #296, opened 2022-04-01
feat: Bump knative, use chart version as app version

gracedo:
I noticed that the knative app version we're adding here is set to the appversion, but we are actually aligning our app version with the chart version - so fixing it there as well as bumping to the latest version.

Depends on https://github.com/mesosphere/charts/pull/1309
Once ^ merges, I will test this branch on `kommander` and once that goes green, then we can merge this PR.

Testing in https://github.com/mesosphere/kommander/pull/1636


alexeygorobets (approved):

branden (approved):

gracedo:
https://github.com/mesosphere/kommander/pull/1636 is green ✅

PR #298, opened 2022-04-02
chore: Set istio's kubectl image to the latest release

branden:
https://jira.d2iq.com/browse/D2IQ-84836

Tested at https://github.com/mesosphere/kommander/pull/1639

### DIFF
--- a/services/istio/1.11.6/defaults/cm.yaml
+++ b/services/istio/1.11.6/defaults/cm.yaml
@@ -24,3 +24,6 @@ data:
       serviceMonitor:
         labels:
           prometheus.kommander.d2iq.io/select: "true"
+    global:
+      image: "${kubetoolsImageRepository:=bitnami/kubectl}"
+      tag: "${kubetoolsImageTag:=1.23.5}"


services/istio/1.11.6/defaults/cm.yaml
```diff
@@ -24,3 +24,6 @@ data:
       serviceMonitor:
         labels:
           prometheus.kommander.d2iq.io/select: "true"
+    global:
+      image: "${kubetoolsImageRepository:=bitnami/kubectl}"
+      tag: "${kubetoolsImageTag:=1.23.5}"
```
branden, line 29: The default is 1.22.6: https://github.com/mesosphere/charts/blob/c7a2ddf604907d408f0fedcc41dd44af3693191d/staging/istio/values.yaml#L1-L3

gracedo (approved):
woops sorry, I just tagged kapps for rc.6 so this won't make it, but it can go into rc.7

branden:
The do-not-merge label can be removed once https://github.com/mesosphere/kommander/pull/1639 is green.

branden:
https://github.com/mesosphere/kommander/pull/1639 is green 🎉

gracedo:
blocked by https://jira.d2iq.com/browse/D2IQ-87555

branden:
This only modifies jobs that are managed by helm, so it should be unaffected by [D2IQ-87555](https://jira.d2iq.com/browse/D2IQ-87555).

gracedo:
@branden my bad! jumped the gun. I guess we reverted the other kubectl bumps in jobs defined in kapps, only those changes are probably affected by [D2IQ-87555](https://jira.d2iq.com/browse/D2IQ-87555). I'll add a note in that ticket that it should including getting the image bumps back in

cbuto (approved):

cwyl02 (approved):

services/istio/1.11.6/defaults/cm.yaml
takirala: where is `kubetoolsImageRepository` set?

branden: I looked a little and couldn't find where we set it. But we already reference it a lot.

branden:
Checks on https://github.com/mesosphere/kommander/pull/1639 passed, except for a flake on the airgapped k8s 1.23 test: https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_Kommander2_HighCpu_AirgappedValidationTestsKubernetesV123/3417390?buildTab=log&focusLine=1516&linesState=1516

services/istio/1.11.6/defaults/cm.yaml
gracedo: I can't find it anywhere, I thought maybe it would be in k-cli in one of the substitution var files but I can't find it there either. maybe we're not actually setting it?? cc @mikolajb do you know 😅

gracedo: I can't find it anywhere, I thought maybe it would be in k-cli in one of the substitution var files but I can't find it there either. maybe we're not actually setting it?? cc @mikolajb do you know 😅

mikolajb: Yeah, it looks like we don't set it 😮 
https://github.com/search?q=org%3Amesosphere+kubetoolsImageRepository&type=code

takirala: _Lost in translation_ looks like - https://github.com/mesosphere/kommander/blob/v2.0.0/chart/kommander-bootstrap/templates/helmReleases/common/kommander-vars.yaml#L28-L29

gracedo: We should probably get that added back in :) https://github.com/mesosphere/kommander-cli/blob/ab8e8b2a3287265f8d05a3957dcd8950626e1d70/pkg/installer/flux_config.go#L193

gracedo: Created https://jira.d2iq.com/browse/D2IQ-87946 to track ^

PR #301, opened 2022-04-02
fix: Add ttl to jobs

gracedo:
~Revert the kubectl revert (so, this PR brings them back) and~ set `ttlSecondsAfterFinished` on all jobs defined in kapps to ensure they are deleted after completion to avoid upgrade "immutable field" errors due to updated images etc.

Let's wait to bump images (except for CVE migitations) till after 2.2 GA.

### DIFF
--- a/services/centralized-kubecost/0.23.3/post-install-jobs/post-install-jobs.yaml
+++ b/services/centralized-kubecost/0.23.3/post-install-jobs/post-install-jobs.yaml
@@ -34,6 +34,7 @@ metadata:
   name: copy-kubecost-grafana-datasource-cm
   namespace: kubecost
 spec:
+  ttlSecondsAfterFinished: 100
   template:
     metadata:
       name: copy-kubecost-grafana-datasource-cm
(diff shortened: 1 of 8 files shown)


mikolajb (approved):

cwyl02 (approved):

cbuto:
have we ran this through the upgrade tests? the jobs that are being upgraded are from 2.1, so I’m not 100% certain this will resolve the upgrade issues but we should test if possible!

gracedo:
Nope haven’t tested was going to come back it after the weekend; thinking about it some more we should hold off on bumping (kubectl and other) images for this release (except to mitigate CVEs). We should probably still add ttl to these jobs (to get them cleaned up before the next release !) but not really comfortable bumping images in all these different apps until we have the time to properly test (which we don’t have the luxury of for 2.2 which needs to be out the door). We would also need to do a sweep of other jobs we define in our forked/maintained charts at least, and thoroughly test all other upstream charts we use to make sure bumped images won’t cause issues elsewhere. I think so far we haven’t been hit with the immutable error in the jobs defined in our maintained charts because they are hooks which we clean up after a successful run.

gracedo:
Removed the revert commit (which brought back kubectl bumps) and only included setting TTL to all jobs defined in apps

cbuto (approved):

gracedo:

PR #309, opened 2022-04-04
fix: Update appversions

gracedo:
While updating appversions table for 2.2 RN, noticed we also missed updating some app versions in kapps. hopefully got them all here

### DIFF
--- a/services/kubernetes-dashboard/5.1.1/kubernetes-dashboard.yaml
+++ b/services/kubernetes-dashboard/5.1.1/kubernetes-dashboard.yaml
@@ -77,4 +77,4 @@ data:
   name: "Kubernetes"
   dashboardLink: "/dkp/kubernetes/"
   docsLink: "https://kubernetes.io/docs/home/"
-  version: "5.1.1"
+  version: "2.4.0"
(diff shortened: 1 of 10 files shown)


cbuto (approved):

services/kubernetes-dashboard/5.1.1/kubernetes-dashboard.yaml
```diff
@@ -77,4 +77,4 @@ data:
   name: "Kubernetes"
   dashboardLink: "/dkp/kubernetes/"
   docsLink: "https://kubernetes.io/docs/home/"
-  version: "5.1.1"
+  version: "2.4.0"
```
cbuto, line 80: just for my reference, this is the app version right?

gracedo: yep, app version (vs chart version which we also use as our "kommander application version" 😬 )

cbuto: cool cool thanks!

takirala (approved):

PR #315, opened 2022-04-05
chore: Bump kubefed's kubectl image

branden:
https://jira.d2iq.com/browse/D2IQ-84836

Tested at https://github.com/mesosphere/kommander/pull/1653

### DIFF
--- a/services/kubefed/0.9.1/defaults/cm.yaml
+++ b/services/kubefed/0.9.1/defaults/cm.yaml
@@ -30,3 +30,7 @@ data:
       service:
         labels:
           servicemonitor.kommander.mesosphere.io/path: metrics
+      postInstallJob:
+        repository: bitnami
+        image: kubectl
+        tag: 1.23.5


gracedo:
blocked by https://jira.d2iq.com/browse/D2IQ-87555

branden:
This only modifies jobs that are managed by helm, so it should be unaffected by [D2IQ-87555](https://jira.d2iq.com/browse/D2IQ-87555).

cbuto (approved):

cwyl02 (approved):

gracedo (approved):

services/kubefed/0.9.1/defaults/cm.yaml
```diff
@@ -30,3 +30,7 @@ data:
       service:
         labels:
           servicemonitor.kommander.mesosphere.io/path: metrics
+      postInstallJob:
+        repository: bitnami
+        image: kubectl
+        tag: 1.23.5
```
takirala, line 36: can we add a comment here that has the string `0.9.1` so that it comes up in editor search when kubefed is bumped. we can decide to remove the override (cuz its already bumped upstream) or bump to an even newer version if it comes up in search.

branden: I'm not sure if I follow. I don't think we'd ever want to remove the override, since the intention is to use the same kubectl image across all services defined in this repo, to prevent including multiple kubectl images in the airgap bundle. Part of the difficulty with this task was tracking down all the kubectl references that were in chart defaults or hardcoded in templates. I imagine we'd want to regularly bump the kubectl image everywhere all at once, and this override helps make that easy.

PR #322, opened 2022-04-07
Revert "fix: Add ttl to jobs (#301)"

gracedo:
This reverts commit 20a81bb8670e72fdd36f28d7045853972221ffbb.

Reverting adding ttl to our jobs defined in kapps since they should not get re-run (specifically, the gatekeeper cleanup job getting re-run would end up deleting secrets that may not get recreated in the helm chart is already deployed)

```
	// Force instructs the controller to recreate resources
	// when patching fails due to an immutable field change.
	// +kubebuilder:default:=false
	// +optional
	Force bool `json:"force,omitempty"`
```
Post release, we will revisit setting these kustomizations `Force` to `true` to avoid the `immutable field` errors if we want to update the images in the jobs (instead of setting ttl).


cbuto (approved):

takirala (approved):

takirala:

PR #325, opened 2022-04-07
upgrade: bump metallb version to 0.10.3

palexster:
jira: https://jira.d2iq.com/browse/D2IQ-87268


vespian (approved):

gracedo:
we are actually getting rid of metallb in kommander, so we don't need to bump it :)

gracedo (changes requested):
just blocking this one. I think we're going to remove the app entirely in 2.3

PR #335, opened 2022-04-11
fix: use correct envvar name in traefik init container

makkes:
The envvar name before has never been consumed so I'm changing it back to the value that's [actually consumed](https://github.com/mesosphere/kubeaddons-extrasteps/blob/v0.5.5/pkg/traefik/traefik.go#L103).


services/kommander/0.2.0/defaults/cm.yaml
```diff
@@ -88,7 +88,7 @@ data:
       prerequisites:
         defaultApps:
           reloader: "0.0.104"
-          traefik: "10.9.1"
+          traefik: "10.9.2"
```
aweris, line 91: [nit] it would be great if we create another PR for rest of changes since they seems not related with renaming

aweris (approved):

services/kommander/0.2.0/defaults/cm.yaml
makkes: I'm not sure I understand what you'd like me to do.

makkes: I'm not changing the version of traefik itself, only the App version so as to make it upgradeable.

aweris: Ohh, it makes sense. Confused because of the version schema.

aweris: Just realized this change is going to break e2e tests. In some parts of apps, we're still assuming the app version equal to the chart version. 

However, we need to change those assumptions anyway

makkes:
force-pushed after rebase to resolve merge conflicts

mikolajb (approved):

vespian (approved):

makkes:
[all checks ran and passed](https://teamcity.mesosphere.io/project/ClosedSource_Kommander2_KommanderApplications?branch=pull%2F335&buildTypeTab=overview&mode=trends). Admin-merging.

PR #347, opened 2022-04-14
fix: Set CreateReplace upgrade policy on kps

gracedo:
kube-prometheus-stack has been updating CRDs across upgrades, but we didn't ste the upgrade policy to `CreateReplace` so we're not actually updating the CRDs during upgrade. We'll need to backport this to get it into the next 2.2 patch.

Also saw that kubetunnel CRDs policy was only set on `install` not `upgrade`, I think we probably need both?

testing in https://github.com/mesosphere/kommander/pull/1687


cbuto (approved):

takirala (approved):

gracedo:

PR #348, opened 2022-04-14
feat: Remove metallb

gracedo:
As of 2.2 we are no longer supporting metallb installation via kommander. There are steps in the 2.2 upgrade docs on how to detach and remove metallb appdeployment.

tested in https://github.com/mesosphere/kommander/pull/1688


cbuto (approved):

gracedo:
Not as simple as removing, CI in at least kommander repo would need to be adjusted.
https://github.com/mesosphere/kommander/pull/1688 for example of failing kommander CI

PR #350, opened 2022-04-15
fix: Prevent mutating minio-operator job during upgrade

branden:
https://jira.d2iq.com/browse/D2IQ-87960

This PR directs Flux to recreate the job that creates Loki's bucket in minio when mutations are disallowed. This prevents an error that can block upgrades. (I actually intend to replace the job with the Tenant CR's support for declaring buckets in an upcoming PR.)

The e2e upgrade test for 2.2.0-dev shows that minio pods may be repeatedly OOMkilled with the existing configuration, which can contribute to test flakes. This PR also bumps minio's memory requests and limits by 50% to address this.

This will need to be backported to release-2.2.

Tested against kommander at https://github.com/mesosphere/kommander/pull/1693.

Here's a passing upgrade test that uses this PR branch: https://teamcity.mesosphere.io/buildConfiguration/MesosphereOnly_ClosedSource_E2E_Kommander_E2eUpgradeK21ClusterKinD/3422495?buildTab=tests&status=passed&expandedTest=build%3A%28id%3A3422495%29%2Cid%3A1949 (The kommander app version test fails because it expects to be running against 2.2.0-dev, not a PR branch based off main.)


gracedo (approved):

cbuto (approved):
💭 are these jobs safe to be re-ran? Now that they will be removed after the TTL, I think they will be ran again at some point (we ran into this with the gatekeeper jobs for example).

cwyl02 (approved):

services/grafana-loki/0.33.2/minio.yaml
```diff
@@ -68,9 +68,11 @@ data:
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: grafana-loki-minio-make-bucket
+  name: grafana-loki-minio-create-bucket
   namespace: ${releaseNamespace}
 spec:
+  # If this job lingers then upgrades may fail on attempting to modify its spec, which is immutable.
+  ttlSecondsAfterFinished: 60
```
gracedo, line 75: What @cbuto said, I'm not sure we want to be re-running this job, we should probably instead set `force` to `true` on the flux kustomization, more background in https://jira.d2iq.com/browse/D2IQ-87555

gracedo (changes requested):

branden:
The bucket creation script is idempotent, it'll do nothing and exit with success if the bucket already exists. I'll try setting `force` on the kustomization though.

cwyl02 (approved):

services/grafana-loki/0.33.2/jobs.yaml
```diff
@@ -0,0 +1,21 @@
...
+  timeout: 1m
+  path: ./services/grafana-loki/0.33.2/jobs
+  sourceRef:
+    kind: GitRepository
+    name: management
+    namespace: kommander-flux
+  postBuild:
+    substituteFrom:
+      - kind: ConfigMap
+        name: substitution-vars
```
gracedo, line 21:
```suggestion
  # passing releaseNamespace to 2nd level configuration files for ability to configure namespace correctly in attached clusters
  # Using `substituteFrom` with `substitution-vars` creates 2nd level resources in `kommander` namespace instead of workspace ns
  postBuild:
    substitute:
      releaseNamespace: ${releaseNamespace}
```
I think we need this (`substitution-vars` only exists on mgmt cluster)

services/project-grafana-loki/0.33.2/jobs.yaml
```diff
@@ -0,0 +1,21 @@
...
+  timeout: 1m
+  path: ./services/project-grafana-loki/0.33.2/jobs
+  sourceRef:
+    kind: GitRepository
+    name: management
+    namespace: kommander-flux
+  postBuild:
+    substituteFrom:
+      - kind: ConfigMap
+        name: substitution-vars
```
gracedo, line 21:
```suggestion
  # passing releaseNamespace to 2nd level configuration files for ability to configure namespace correctly in attached clusters
  # Using `substituteFrom` with `substitution-vars` creates 2nd level resources in `kommander` namespace instead of workspace ns
  postBuild:
    substitute:
      releaseNamespace: ${releaseNamespace}
```
I think we need this (`substitution-vars` only exists on mgmt cluster)

services/grafana-loki/0.33.2/jobs.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
+kind: Kustomization
+metadata:
+  name: grafana-loki-jobs
+  namespace: ${releaseNamespace}
+spec:
```
gracedo, line 6: Does this need to depend on anything? e.g. does it need to be run after loki HR is up, or after minio is up? And I know not directly related to this PR but I saw `Tenant` which is a minio CR is defined in these charts as well, did we not run into issues of that potentially getting applied before the minio CRDs were actually created? Since I don't think we have any dependencies set there either 🤔  (should we set them?)

branden: This job won't succeed until after Minio is up and accepting requests. There's an inner retry loop that reattempts failed connections, and then the job itself will restart on failure. Not setting an explicit dependency between the job and Tenant hasn't caused any problems, so I don't see a reason to change that here.

gracedo (approved):

branden:

PR #364, opened 2022-04-25
added mergebot workflow file

shaeed-noordeen:


.github/workflows/mergebot.yml
```diff
@@ -0,0 +1,53 @@
...
+          path: mergebot_code/mergebot
+          ref: sn/kommander-cli-test
+          persist-credentials: false
+          token: ${{ secrets.MERGEBOT_TOKEN }}
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+          cache: 'pip'
+      - name: Install mergebot Python dependencies
+        run: pip install -r ./mergebot_code/mergebot/requirements.txt
```
jimmidyson, line 42: Why don't we build a docker image for mergebot and run it directly e.g. https://docs.github.com/en/enterprise-server@3.3/actions/using-jobs/running-jobs-in-a-container ?

.github/workflows/mergebot.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Run Mergebot
+        env:
+          LOG_STDOUT: "true"
+          EVENT_TYPE: ${{ github.event_name }}
+          RUN_ID: ${{ github.run_id }}
+          MERGEBOT_TC_USERNAME: ${{secrets.MERGEBOT_TC_USERNAME}}
+          MERGEBOT_TC_PASSWORD: ${{secrets.MERGEBOT_TC_PASSWORD}}
+          MERGEBOT_TC_TOKEN: ${{secrets.MERGEBOT_TC_TOKEN}}
+          JIRA_USERNAME: ${{secrets.JIRA_USERNAME}}
+          JIRA_PASSWORD: ${{secrets.JIRA_PASSWORD}}
```
jimmidyson, line 52: I'd personally prefer the TC and JIRA actions to be separated out into separate workflows. They're logically different actions, with different triggers, etc so I feel it would be more idiomatic to separate these out.

PR #366, opened 2022-04-25
feat: notify in slack about repo tarball release

farhan5900:
Resolves: https://jira.d2iq.com/browse/D2IQ-88067

### DIFF
--- a/make/release.mk
+++ b/make/release.mk
@@ -4,9 +4,14 @@ S3_ACL ?= "bucket-owner-full-control"
 
 .PHONY: release
 release: ARCHIVE_NAME = kommander-applications-$(GIT_TAG).tar.gz
+release: PUBLISHED_URL = https://downloads.d2iq.com/dkp/$(GIT_TAG)/kommander-applications-$(GIT_TAG).tar.gz
 release: install-tool.awscli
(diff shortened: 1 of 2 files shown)


gracedo (approved):

rishabh96b (approved):

make/release.mk
```diff
@@ -4,9 +4,14 @@ S3_ACL ?= "bucket-owner-full-control"
 
 .PHONY: release
 release: ARCHIVE_NAME = kommander-applications-$(GIT_TAG).tar.gz
+release: PUBLISHED_URL = https://downloads.d2iq.com/dkp/$(GIT_TAG)/kommander-applications-$(GIT_TAG).tar.gz
```
makkes, line 7:
```suggestion
release: PUBLISHED_URL = https://downloads.d2iq.com/dkp/$(GIT_TAG)/$(ARCHIVE_NAME)
```

makkes (changes requested):

make/release.mk
farhan5900: In this case, the curl was complaining about the invalid payload since `ARCHIVE_NAME` is itself using `GIT_TAG`. That's why I kept it like that.

makkes (approved):

PR #380, opened 2022-05-05
feat: add validation against multiple Kubernetes versions

floridoo:
Validates the manifests against different Kubernetes versions, configurable in `.bloodhound.yml`.

Jira: https://jira.d2iq.com/browse/D2IQ-88980


vespian (approved):
:heart:

make/validate.mk
```diff
@@ -8,4 +8,4 @@ $(DKP_BLOODHOUND_BIN):
 
 .PHONY: validate-manifests
 validate-manifests: $(DKP_BLOODHOUND_BIN)
-	$(DKP_BLOODHOUND_BIN) . --skip-applications=kaptain,grafana-loki,project-grafana-loki,project-logging
+	$(DKP_BLOODHOUND_BIN)
```
takirala, line 11: (nit) - its fine to make CI verbose.
```suggestion
	$(DKP_BLOODHOUND_BIN) -v 4
```

takirala (approved):

make/validate.mk
floridoo: In this case it lists all checked resources, which makes it pretty hard to find the failing ones (while not providing much value IMO).

PR #395, opened 2022-05-16
feat: Disable minio inplace upgrades

cbuto:
testing in https://github.com/mesosphere/kommander/pull/1779

`MINIO_UPDATE` is `on` by default, which allows for inplace upgrades of minio (aka download the latest binary and trigger a restart of minio). We don't want inplace upgrades happening within the container, so this commit disables that functionality.


cbuto:
apparently this is required for minio to work 🤔

PR #404, opened 2022-05-20
build: add 0.0.2 cert-federation chart for airgapped bundle

takirala:
needed for https://github.com/mesosphere/kommander/pull/1798 

Because of how the CLI `chart-bundle` command works, we only pull those helm charts which are referred in HelmReleases. By creating a HelmRelease in common/build/kommander-cert-federation.yaml we are ensuring the the kommander-cert-federation-0.0.2-tgz is always downloaded. 

ideally, we can create an app in `services` directory and then create an appdeployment to deploy that but : 
- it gets more complicated as we need to deploy the appdeployment thrice and need to populate different values each time
- it ll create a ClusterApp and we dont want that (but we can live with it)
- it ll display in ui and we definitely dont want that. previously you could mark it as type: internal and we would hide those but that has been changed lately.

for non-airgapped deployments, this file is ignored and is pulled directly from m/charts repo. for airgapped deployments, looking at this file makes the chart-bundle command pull the helm chart when building k-apps bundle.

we cant simply amend kommander repo tooling to include the `kommander-cert-federateion-0.0.2.tgz` because even though that would satisfy the case where a user is solely relying on the bundles we ship for airgapped deployments, it doesn't satisfy the case where a user does `dkp install kommander -v6 --airgapped` which downloads charts from internet. (and writing this message today, i realize that it is a broken (or expected?) behavior today in CLI :shrug: )


cbuto (approved):
is it worth including a comment explaining why this is needed? at first glance, someone might be confused since its not really referenced anywhere 😓

takirala:
> is it worth including a comment explaining why this is needed? at first glance, someone might be confused since its not really referenced anywhere sweat

good idea, will update with a comment.

.github/workflows/backport.yaml
```diff
@@ -25,7 +25,3 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           auto_backport_label_prefix: backport-to-
-
```
takirala, line 28: the action is working fine, but it fails when it tries to print the log. https://github.com/mesosphere/kommander-applications/runs/6562935492?check_suite_focus=true

branden (approved):

PR #423, opened 2022-06-02
feat: disable remote bases in kustomize overlays

takirala:
kustomize-controller 0.25.0+ introduced a flag to support remote
kustomization overlays by default. We don't use remote bases for
kustomizations (yet!) so we can safely disable it (as per recommendation
upstream).

See https://github.com/fluxcd/kustomize-controller/pull/638


gracedo (approved):

cwyl02 (approved):

takirala:
merged in #427

PR #433, opened 2022-06-10
ci: add make targets to execute e2e tests

aweris:
Add e2e testing for kommander-applications

Tests:

- install, 
- upgrade - single cluster

Make commands:

- `make test.e2e.install VERBOSE=true`
- `make test.e2e.upgrade.singlecluster VERBOSE=true`

Remaining work:

- [x] TC jobs with pull request triggers

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -21,20 +21,6 @@ kommander-e2e:
 			git checkout main ; \
 	fi
 
-.PHONY: kommander-e2e
-kommander-e2e: ## Clones the kommander-e2e repo locally or updates the clone
-kommander-e2e:
-	@if [ -d $(KOMMANDER_E2E_DIR) ] ; then \
-		cd $(KOMMANDER_E2E_DIR) && \
-			git fetch origin && \
-			git reset --hard origin/main ; \
-	else \
-		mkdir -p $(KOMMANDER_E2E_DIR) && \
-			git clone -q https://github.com/mesosphere/kommander-e2e.git $(KOMMANDER_E2E_DIR) && \
-			cd $(KOMMANDER_E2E_DIR) && \
-			git checkout main ; \
-	fi
-
 .PHONY: test.e2e.install
 test.e2e.install: kommander-e2e ; $(info $(M) running end-to-end kommander install test from kommander-e2e)
 	cd $(KOMMANDER_E2E_DIR) && \
--- a/make/test.mk
+++ b/make/test.mk
@@ -3,10 +3,10 @@ KOMMANDER_E2E_DIR  = $(REPO_ROOT)/.tmp/kommander-e2e
 # E2E configurations
 E2E_TIMEOUT       ?= 120m
 E2E_KINDEST_IMAGE ?= "kindest/node:v1.23.5"
+UPGRADE_FROM_VERSION ?= "release-2.2"
 
 # (aweris): This should be a temporary workaround for v2.3.0 development. If you're still see clone test in v2.4.0
 # it means "a temporary workaround" actually means "permanent solution".
-
 .PHONY: kommander-e2e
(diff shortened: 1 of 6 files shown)


cbuto:
going to spend sometime on this during a slow point in on-call

cbuto:
running into this in the upgrade tests

```
err="failed to ensure "upgrading conditions are met": existing kommander version does not meet the minimum upgradable criteria : v2.3.0-dev"
```

cbuto:
@d2iq-mergebot test kommander-applications/install-e2e-tests

gracedo:
> The [install e2e tests](https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_Kommander2_KommanderApplications_InstallE2eTests/3502850) are passing but I ran into this in the [upgrade tests](https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_Kommander2_KommanderApplications_UpgradeE2eTests/3502773):
> 
> ```
> err="failed to ensure "upgrading conditions are met": existing kommander version does not meet the minimum upgradable criteria : v2.3.0-dev"
> ```
> 
> I'm not sure if we want to bump that constraint to allow upgrading to 2.3.0-dev ([mesosphere/kommander-cli#593](https://github.com/mesosphere/kommander-cli/pull/593))

@cbuto @aweris do we want the upgrades test here to install the latest 2.2 version instead of off main? That might be the better test scenario, we probably don't need to test upgrades against the latest in main as we wouldn't undergo that upgrade path, it'd only be from the version in the last kommander release

make/test.mk
```diff
@@ -0,0 +1,44 @@
...
+		VERBOSE=$(VERBOSE) \
+		make test.e2e
+
+.PHONY: test.e2e.upgrade.singlecluster
+test.e2e.upgrade.singlecluster: kommander-e2e ; $(info $(M) running end-to-end kommander upgrade main to $(GIT_COMMIT) test from kommander-e2e)
+	cd $(KOMMANDER_E2E_DIR) && \
+		E2E_TEST_PATH="feature/upgrade/suites/kind/singlecluster" \
+		E2E_TIMEOUT=$(E2E_TIMEOUT) \
+		E2E_KINDEST_IMAGE=$(E2E_KINDEST_IMAGE) \
+		E2E_KOMMANDER_APPLICATIONS_REPOSITORY="github.com/mesosphere/kommander-applications.git?ref=main" \
```
gracedo, line 40: going off my other comment, perhaps we want to point this to the last release? (for 2.2, 2.2.2-dev?)

aweris: My intention here is to test only branch changes in e2e tests. If we use previous release, all other changes will be included in this test as well

make/test.mk
```diff
@@ -0,0 +1,44 @@
+KOMMANDER_E2E_DIR  = $(REPO_ROOT)/.tmp/kommander-e2e
+
+# E2E configurations
+E2E_TIMEOUT       ?= 120m
+E2E_KINDEST_IMAGE ?= "kindest/node:v1.23.5"
+UPGRADE_FROM_VERSION ?= "release-2.2"
```
cbuto, line 6: we'll need to update some release automation tickets to make sure this gets updated to the previous release in each branch

make/test.mk
cbuto: yeah I think using the last release definitely adds some value. I updated that in https://github.com/mesosphere/kommander-applications/pull/433/commits/c8b7677e21861f642e3ff54c5e9cee742819bcdd

cbuto:
I'm looking into the upgrade failures. It looks like it might be installing the wrong kustomization for 2.2 `./clusters/kommander_host-cluster` 🤔

palexster (approved):

PR #440, opened 2022-06-15
Fix kiali references to istio and jaeger

branden:
Fixes https://jira.d2iq.com/browse/D2IQ-89452.

Tested against kommander at https://github.com/mesosphere/kommander/pull/1901.

This needs to be backported to release-2.2.


services/kiali/1.49.0/defaults/cm.yaml
```diff
@@ -25,7 +26,7 @@ data:
           prometheus:
             url: http://kube-prometheus-stack-prometheus.${workspaceNamespace}.svc.cluster.local:9090
           tracing:
-            in_cluster_url: http://jaeger-kubeaddons-jaeger-operator-jaeger-query:16686
+            in_cluster_url: http://jaeger-jaeger-operator-jaeger-query.istio-system.svc.cluster.local:16686
```
gracedo, line 29: yikes, good catch

gracedo (approved):
I think we'll need to bump the application version since this is a config-only change :/ for 2.3 we can upgrade kiali to the latest (followup pr, looks like 1.52.0 from https://kiali.org/helm-charts/index.yaml)

branden:
@gracedo Thanks, I forgot that the helm controller won't pick up changes to values configmaps. I pushed 8b2f694a17c22328080dffb6517fba651ea9186f to change the configmap name so that we can leave the application version where it is. I grepped through kommander and kommander-cli and didn't find any assumptions about the defaults configmap resource name, but lmk if you think it could cause a problem.

services/kiali/1.49.1/kiali.yaml
```diff
@@ -42,8 +42,8 @@ data:
   name: "Kiali"
   dashboardLink: "/dkp/kiali"
   docsLink: "https://istio.io/docs/tasks/telemetry/kiali/"
-  # Chart version matches app version
-  version: "1.49.0"
+  # App version is ahead of chart version due to config updates.
+  version: "1.49.1"
```
gracedo, line 46: So I think here we actually want to display the "true" AppVersion of kiali, which is 1.49.0 (1.49.1 is now our kommander application version, which is still deploying kialia appversion v1.49.0).

```suggestion
  version: "1.49.0"
```

takirala (approved):

branden:
@d2iq-mergebot backport release-2.2

branden:

PR #455, opened 2022-06-27
feat: traefik-forward-auth-mgmt: move init logic to controller

floridoo:
Remove init container, because the configuration is handled by the `TFAController`.

Must be merged together with https://github.com/mesosphere/kommander/pull/1974

Jira: https://jira.d2iq.com/browse/D2IQ-90588


services/traefik-forward-auth-mgmt/0.3.9/defaults/cm.yaml
```diff
@@ -18,15 +18,13 @@ data:
       port: 4181
     traefikForwardAuth:
       enabled: true
-      # oidcUri will be overridden by the init-container
+      # oidcUri, clientID, clientSecret will be overridden by the TFAController
       oidcUri: "https://dex.${releaseNamespace}.svc.cluster.local:8080/dex"
-      caSecretName: kommander-traefik-certificate
       clientId: traefik-forward-auth
       clientSecret:
+        value: "placeholder"
```
aweris, line 25: [Question] Why do we add `placeholder` as value? Shouldn't we skip this if we don't have proper value for it?

aweris (approved):

services/traefik-forward-auth-mgmt/0.3.9/defaults/cm.yaml
floridoo: It is analog to the existing definition in `traefik-forward-auth`. I guess it makes it clearer that this is a value that is later overwritten by the controller.

mhrabovcin:
The `kommander-applications/install-e2e-tests` and `kommander-applications/upgrade-e2e-tests` are expected to fail on this PR as the changes in the PR must be accompanied with https://github.com/mesosphere/kommander/pull/1974

PR #459, opened 2022-06-29
Automate service/component versions

bcmendoza:
https://jira.d2iq.com/browse/D2IQ-77062
Also lays some groundwork for https://jira.d2iq.com/browse/D2IQ-90395


hack/static/versions.yaml
```diff
@@ -0,0 +1,473 @@
+- AppID: centralized-grafana
+  AppName: centralized-grafana-34.9.3
+  AppVersion: 0.55.0
```
gracedo, line 3: so this would be the prometheus-operator version, I believe.
the thing that makes this significantly more complex (kube-prometheus-stack is probably the most complicated case) is that within kps there are many components -- prometheus-operator (the chart's appversion in this case), grafana, prometheus, alertmanager.. so maybe for kps we would need to parse the values file https://github.com/mesosphere/charts/blob/master/staging/kube-prometheus-stack/values.yaml to get the various component images?

e.g. alertmanager https://github.com/mesosphere/charts/blob/master/staging/kube-prometheus-stack/values.yaml#L421

You can take a look at the 2.2.1 release notes table https://docs.d2iq.com/dkp/kommander/2.2/release-notes/2.2.1/#component-updates at `Component Versions` column. looks like hopefully just kps where we have to add some special handling to get the other application versions

hack/static/versions.yaml
```diff
@@ -0,0 +1,473 @@
...
+  AppVersion: v0.3.0
+  Chart: kube-oidc-proxy
+  DisplayName: kube-oidc-proxy
+  Enabled: false
+  LatestVersion: 0.3.1
+  Path: /services/kube-oidc-proxy
+  Repo: mesosphere.github.io-charts-staging
+  Type: internal
+  Version: 0.3.1
+- AppID: kube-prometheus-stack
```
gracedo, line 243: That comment is more accurate actually here for the kps app where we are deploying all the other components

make/repo.mk
```diff
@@ -24,3 +24,7 @@ ifeq ($(GIT_CURRENT_BRANCH),main)
 else
 	svu patch --pattern 'v[0-9].[0-9]{[0-9],}.[0-9]{[0-9],[0-9]-rc.*,-rc.*,}' --suffix dev --no-metadata
 endif
+
+.PHONY: repo.gen.versions
+repo.gen.versions: ## Generates a file with all current kommander-apps versions
```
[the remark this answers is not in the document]
bcmendoza: @cbuto thanks for the find. I changed a line to force downloading the chart from remote and then showing values

hack/static/cmd/versions.go
```diff
@@ -0,0 +1,174 @@
...
+
+		sort.Sort(kApps)
+
+		k, err := cursor.GetByID("kommander-application://kommander")
+		if err != nil {
+			return err
+		}
+
+		fmt.Println("Determining which applications are installed by default...")
+		defaultApps, err := getKommanderDefaultApps(k)
```
cbuto, line 40: this will get the enterprise apps and default workspace apps, but those aren't necessarily the entire set of apps that are enabled by default when installing Kommander, maybe we can rename the `Enabled` field to be more precise? 

Here are the default apps deployed by the CLI -> https://github.com/mesosphere/kommander-cli/blob/c283173ea1f3cafdcdfef2ed55ba252d92b2b90c/pkg/installer/config/config.go#L175.

bcmendoza: Ah, thanks for this. Any idea on how I can get this list of apps by analyzing the kApps repo?

bcmendoza: @cbuto Is this list reliable for current and future versions? https://github.com/mesosphere/kommander-cli/blob/c283173ea1f3cafdcdfef2ed55ba252d92b2b90c/pkg/installer/config/config.go#L212 if so, I'll just hardcode into this tool.

cbuto: that list could definitely change from release to release, you can see above that where we have a [different set of defaults ](https://github.com/mesosphere/kommander-cli/blob/c283173ea1f3cafdcdfef2ed55ba252d92b2b90c/pkg/installer/config/config.go#L175) for 2.1.

bcmendoza: It’s hacky and non-trivial, but I could parse the AST. But maybe better to export that map, import it here, and then bump the kommander-cli dependency in this program.

bcmendoza: Note from today's standup: We'll just not include default apps info here in this tool, and instead that information should be updated in the docs page manually.

cbuto (approved):

cwyl02 (approved):

PR #537, opened 2022-08-10
test: Test against Kubernetes v1.24

jimmidyson:
Test against Kubernetes v1.24.

```release-note

```

- [x] No License Change (or NA).

### DIFF
--- a/.bloodhound.yml
+++ b/.bloodhound.yml
@@ -4,6 +4,7 @@ k8s_versions:
   - 1.21.0
   - 1.22.0
   - 1.23.6
+  - 1.24.3
 
 # paths to load (in that order)
 paths:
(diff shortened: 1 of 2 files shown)


jimmidyson:
@aweris @mhrabovcin The upgrade test failure is actually a failure in the install of the previous kommander version on kubernetes 1.24 prior to upgrading to latest. This is a known issue and exactly what we're fixing in dkp 2.4 - support for kubernetes 1.24. What should we do here? Only run upgrade test on kubernetes 1.23? I'd like install test to run on 1.24 so maybe I should separate out the version of kubernetes to run upgrade test on? Wdyt?

jimmidyson:
I've changed this to test upgrade on Kubernetes v1.23 in b7fe85e779b304b236a790d6fa284fddba95e952.

Azhovan (approved):

mhrabovcin (approved):

mikolajb (approved):

PR #540, opened 2022-08-10
fix(loki): set flush_op_timeout to 10m

cbuto:
As a part of https://jira.d2iq.com/browse/D2IQ-91707, it was found that the `flush_op_timeout` is defaulted to 10s in Loki <2.5.0 and that was causing issues with Loki failing to startup as the object storage was down for a long time and the WAL built up so the initial flushing of the WAL was taking longer than 10s.

We should set the default here to avoid issues in the future with respect to changing Loki versions.

```release-note

```

### DIFF
--- a/services/grafana-loki/0.48.4/defaults/cm.yaml
+++ b/services/grafana-loki/0.48.4/defaults/cm.yaml
@@ -37,6 +37,7 @@ data:
           chunk_encoding: snappy
           chunk_retain_period: 1m
           max_transfer_retries: 0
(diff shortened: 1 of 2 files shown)


cwyl02 (approved):

gracedo (approved):

gracedo:
does this need to go into release-2.3?

bcmendoza (approved):

cwyl02:
@d2iq-mergebot backport release-2.3

PR #546, opened 2022-08-11
feat: Add 120GB capacity to Loki storage by default

cbuto:

```release-note

```


cneth:
Just fyi there is a page in the docs that (tries to) document the size and number of PV's necessary for all the platform applications, so if we tweak the config, that probably needs an update.

https://d2iq.atlassian.net/wiki/spaces/DENT/pages/29919792/Workspace+Platform+Application+Configuration+Requirements

cbuto:
we are no longer planning on increasing the default storage size

PR #548, opened 2022-08-12
revert: karma config map change of `warn` level

takirala:

```release-note

```


cbuto (approved):

mikolajb (approved):

services/karma/2.0.1/defaults/cm.yaml
```diff
@@ -76,7 +76,7 @@ data:
           prefix: /dkp/kommander/monitoring/karma/
         log:
           config: true
-          level: warn
+          level: info
```
dkoshkin, line 79: Is `warn` not a real level?

DreadPirateShawn: i wonder if it's `warning`

cbuto: it does look like `warning` is the right setting 

https://github.com/prymitive/karma/blob/main/docs/CONFIGURATION.md#log

gracedo: we will just revert to info, these were not generating so much noise to cause disk issues

gracedo (approved):

services/fluent-bit/0.19.21/kustomization.yaml
```diff
@@ -2,3 +2,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - fluent-bit.yaml
+  - grafana-dashboards-logging
```
gracedo, line 5: image
After manually pushing these changes to a cluster, the audit dashboard disappeared (since fluent-bit was not installed) (ignore the other dashboards, those were created ad hoc)

services/karma/2.0.1/defaults/cm.yaml
takirala: fixed in #550 for `main` branch.

PR #575, opened 2022-08-24
feat: Bump KPS to v39.9.0

gracedo:
Bump KPS to latest upstream

```release-note

```


gracedo:
This bump did not fix the prometheus-adapter node metrics issue. I am going to open a PR instead to downgrade KPS to the last working version to fix this for now. We should continue to track the issue and pull in a KPS upgrade with the fix if one is found.

PR #598, opened 2022-09-07
build: Add workflow to create kommander branch

gracedo:

This workflow is triggered on PRs with a label `open-kommander-pr`- the action triggers (and pushes) a branch (prefixed with `kapps/`) to the kommander repo, which includes an update to the `Makefile` to point the kommander-applications ref to the branch that the k-apps PR is based off of. There is then another workflow added to the kommander repo (PR with some fixes to that workflow: https://github.com/mesosphere/kommander/pull/2284) which triggers a PR creation automatically for branches that are prefixed with `kapps/`.

I tested this workflow on my fork with this PR: https://github.com/gracedo/kommander-applications/pull/1 which created this branch on the kommander repo: https://github.com/mesosphere/kommander/tree/kapps/gracedo/test-workflow with this commit https://github.com/mesosphere/kommander/commit/e6b43b85efbc947bdb37a5c0074ed83ea9e93709

The kommander workflow to create the PR is currently failing but the fixes are up in https://github.com/mesosphere/kommander/pulls and on a branch `kapps/grace_test` that I created which successfully triggered the workflow to create the PR here: https://github.com/mesosphere/kommander/pull/2285

Overall workflow steps from both k-apps and kommander:
1. k-apps PR (with e.g. branch name `feat/a-branch`) is labeled with `open-kommander-pr`
2. A branch named `kapps/feat/a-branch` is created and pushed to the kommander repo, which contains an update in the Makefile that changes the k-apps ref from `main` to `feat/a-branch`
3. A draft PR is created in the kommander repo off `kapps/feat/a-branch` branch.

The user that opened the k-apps PR is then responsible for checking that tests pass in the kommander PR and making any additional required changes in that PR. The commit that updated the Makefile then need to be reverted prior to merging the PR into kommander (this user workflow stays the same). Maybe we can look at adding more worrkflows to ease this process as well.

https://jira.d2iq.com/browse/D2IQ-92487

```release-note

```

### DIFF
--- a/.github/workflows/kommander-branch.yaml
+++ b/.github/workflows/kommander-branch.yaml
@@ -1,0 +1,64 @@
+name: "Create Kommander Branch"
+
+on:
+ pull_request:
+    types: [opened, reopened, synchronize, labeled, unlabeled]
+
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  get-kapps-branch-name:
+    runs-on: ubuntu-latest
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'open-kommander-pr') }}
+    outputs:
+      branch_name: ${{ steps.branch-name.outputs.branch_name }}
+      escaped_branch_name: ${{ steps.escaped-branch-name.outputs.escaped_branch_name }}
+      base_branch_name: ${{ steps.base-branch-name.outputs.base_branch_name }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+      - id: branch-name
+        run: echo "::set-output name=branch_name::${{ github.head_ref }}"
+      - id: escaped-branch-name
+        run: echo "::set-output name=escaped_branch_name::$(echo ${{ github.head_ref }} | sed -e 's/\//\\\//g')"
+      - id: base-branch-name
+        run: echo "::set-output name=base_branch_name::${{ github.base_ref }}"
+      - name: Check output branch-name
+        run: echo ${{ steps.branch-name.outputs.branch_name }}
+      - name: Check output escaped-branch-name
+        run: echo ${{ steps.escaped-branch-name.outputs.escaped_branch_name }}
+      - name: Check output base-branch-name
+        run: echo ${{ steps.base-branch-name.outputs.base_branch_name }}
+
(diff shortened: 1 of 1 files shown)


bcmendoza (approved):

takirala (approved):

.github/workflows/kommander-branch.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - id: base-branch-name
+        run: echo "::set-output name=base_branch_name::${{ github.base_ref }}"
+      - name: Check output branch-name
+        run: echo ${{ steps.branch-name.outputs.branch_name }}
+      - name: Check output escaped-branch-name
+        run: echo ${{ steps.escaped-branch-name.outputs.escaped_branch_name }}
+      - name: Check output base-branch-name
+        run: echo ${{ steps.base-branch-name.outputs.base_branch_name }}
+
+  create-kommander-branch:
```
cbuto, line 36: we may be able to create the branch and PR in this action using a PAT that has access to the Kommander repo, but I think this workflow should be good as well!

cbuto (approved):

.github/workflows/kommander-branch.yaml
gracedo: > [Deploy keys](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys) can be set per repository and so are arguably more secure than using a repo scoped [Personal Access Token (PAT)](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token). Allowing the action to push with a configured deploy key will trigger on: push workflows. This makes it an alternative to using a PAT to trigger checks for pull requests. Note that you cannot use deploy keys alone to [create a pull request in a remote repository](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#creating-pull-requests-in-a-remote-repository) because then using a PAT would become a requirement. This method only makes sense if creating a pull request in the repository where the workflow is running.

ref: https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#push-using-ssh-deploy-keys

Using a PAT would work but since this is a public repo I thought it would be safer to not use a PAT here

cbuto: gotcha yeah, makes sense to me

PR #606, opened 2022-09-15
feat: Add new nvidia-gpu-operator app

gracedo:
Adds new `nvidia-gpu-operator` app which will be used instead of `nvidia` starting in 2.4

**The metadata.yaml file needs to be updated with new copy** --> https://jira.d2iq.com/browse/D2IQ-92963

https://jira.d2iq.com/browse/D2IQ-92632

I referenced https://docs.google.com/document/d/1Jj5SU1cz4TBZ0sUnczSCxq2cZtCeWMwdmR0I9LJ9PNk/edit and used the values passed into the `helm install` command there to set our defaults cm

```release-note

```

### DIFF
--- a/services/nvidia-gpu-operator/1.11.1/defaults/cm.yaml
+++ b/services/nvidia-gpu-operator/1.11.1/defaults/cm.yaml
@@ -13,4 +13,4 @@ data:
     toolkit:
       version: v1.10.0-centos7
     gfd:
-      enabled: false
+      enabled: true
--- a/services/nvidia-gpu-operator/1.11.1/defaults/cm.yaml
+++ b/services/nvidia-gpu-operator/1.11.1/defaults/cm.yaml
@@ -1,0 +1,16 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nvidia-gpu-operator-1.11.1-d2iq-defaults
+  namespace: ${releaseNamespace}
+data:
+  values.yaml: |-
+    nfd:
+      enabled: false
+    driver:
+      enabled: false
+    toolkit:
+      version: v1.10.0-centos7
+    gfd:
+      enabled: false
--- a/services/nvidia-gpu-operator/metadata.yaml
+++ b/services/nvidia-gpu-operator/metadata.yaml
@@ -1,0 +1,33 @@
+displayName: NVIDIA GPU Operator
+description: NVIDIA Data Center GPU Manager (DCGM) manages and monitors NVIDIA datacentre GPUs in cluster environments.
+category:
+  - monitoring
+type: platform
+scope:
+  - workspace
+certifications:
+  - certified
+  - supported
+  - airgapped
+overview: |-
+  # Overview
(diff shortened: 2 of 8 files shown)


services/nvidia-gpu-operator/metadata.yaml
```diff
@@ -0,0 +1,33 @@
+displayName: NVIDIA GPU Operator
+description: NVIDIA Data Center GPU Manager (DCGM) manages and monitors NVIDIA datacentre GPUs in cluster environments.
```
gracedo, line 2: TODO this metadata needs to be rewritten -> https://jira.d2iq.com/browse/D2IQ-92963

gracedo:
https://github.com/mesosphere/kommander/pull/2334 is green ✅

services/nvidia-gpu-operator/1.11.1/defaults/cm.yaml
```diff
@@ -0,0 +1,16 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nvidia-gpu-operator-1.11.1-d2iq-defaults
+  namespace: ${releaseNamespace}
+data:
+  values.yaml: |-
```
cbuto, line 8: maybe as a follow-up, can we define a security context for the operator? things like run as non-root, read only filesystem, disable privilege escalation?

cbuto (approved):

services/nvidia-gpu-operator/1.11.1/defaults/cm.yaml
gracedo: added a note to https://d2iq.atlassian.net/browse/D2IQ-92880

services/nvidia-gpu-operator/1.11.1/defaults/cm.yaml
```diff
@@ -0,0 +1,16 @@
...
+    nfd:
+      enabled: false
+    driver:
+      enabled: false
+    toolkit:
+      version: v1.10.0-centos7
+    gfd:
+      enabled: false
```
faiq, line 16:
```suggestion
```

faiq: we can tweak the values in the configuration task. this is just a value we already have some insights on.

as it turns out, v1.11.1 of the gpu operator app doesn't even support turning it off. 

i also thought we agreed to have gpu operator deploy its own gfd from this thread https://mesosphere.slack.com/archives/C01MPKVBB5E/p1663188128968139

faiq (approved):

services/nvidia-gpu-operator/1.11.1/defaults/cm.yaml
gracedo: Ah I see, I didn't realize this was the field we wanted to leave on or that it's not even supported in this release (I just carried over the values from the doc). And if I'm being completely honest, I also didn't want to make all these PRs run through CI again 😆  but given that `enabled: false` isn't even disabling anything now, I'll fix it to at least read `true` so when we upgrade to a version that supports it, there aren't any surprises there since we do want it to be turned on. I'll update + rebase

PR #624, opened 2022-09-21
feat: add `rook-ceph` and `rook-ceph-cluster` services

takirala:

Adds a no-op CephCluster managed by rook-ceph operator. Exhaustive research at https://docs.google.com/document/d/14jCCRI9kcEXYkEVTVCt9i-xYApIgXLcVFpBb-zGFu9c/edit#heading=h.ltksup2t7o57

Future PRs to follow up with integration for velero and Loki. Initial PR to add ceph on kind tests at https://github.com/mesosphere/kommander/pull/2380

- Adds a `rook-ceph` service which is the operator that creates a deployment and a bunch of ceph related CRDs along with some rbac required for operator to function properly.
- Adds a `rook-ceph-cluster` service which creates a `CephCluster` and `CephObjectStorage` CRs which creates a ceph cluster and then an object storage which relies on the ceph cluster. 
- Velero and Loki (in follow up PRs) will create ObjectBucketClaims to initialize their own buckets and use ceph storage.
- Running ceph on KinD is unsupported and I think we are the first ones to do it, see https://github.com/mesosphere/kommander/pull/2380 for details. This means the way e2e tests are run also needs to be changed (will be in follow up PRs when integrating velero and loki with ceph).
- Please see https://docs.google.com/document/d/14jCCRI9kcEXYkEVTVCt9i-xYApIgXLcVFpBb-zGFu9c/edit#heading=h.ltksup2t7o57 for some helpful links to understand how ceph is deployed and configured.

https://d2iq.atlassian.net/browse/D2IQ-92930

```release-note

```

Yes it does, we are adding LGPL-2.1 or LGPL-3 licensed code. This was cleared with legal and attributions will follow. see https://github.com/ceph/ceph/blob/main/COPYING for detailed licensing breakdown


common/helm-repositories/rook-ceph.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: source.toolkit.fluxcd.io/v1beta1
+kind: HelmRepository
```
gracedo, line 3: I think we're missing adding this file into kustomization.yaml https://github.com/mesosphere/kommander-applications/blob/tga/add-ceph-operator/common/helm-repositories/kustomization.yaml
(I only noticed because I added a new app recently and forgot to add it too 😆)

services/rook-ceph/1.10.1/rook-ceph-helmrelease.yaml
```diff
@@ -0,0 +1,30 @@
+---
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
+kind: HelmRelease
```
gracedo, line 3: Can we move this into a dir and put it behind another flux kustomization just in case we need to define jobs on top of this in the future? (e.g. upgrade strategies)

common/helm-repositories/rook-ceph.yaml
takirala: I was confused why it didnt got applied and i was punting debugging it while i was deploying it manually, thanks for catching it!

cwyl02 (approved):

cwyl02:
the icons still look weird but I think that can be fixed later.

takirala:
Icons and metadata will be cleaned up by UI team in https://d2iq.atlassian.net/browse/D2IQ-93345 in a follow up PR.

services/rook-ceph/1.10.1/defaults/cm.yaml
```diff
@@ -0,0 +1,36 @@
...
+  values.yaml: |
+    ---
+    crds:
+      # CRDs are not installed with "crds" directory and If this flag isdisabled post-install, the cluster may be DESTROYED.
+      # If the CRDs are deleted in this case, see the disaster recovery guide to restore them.
+      # https://rook.io/docs/rook/latest/Troubleshooting/disaster-recovery/#restoring-crds-after-deletion
+      enabled: true
+
+    # Whether rook watches its current namespace for CRDs or the entire cluster, defaults to false
+    currentNamespaceOnly: true
```
cbuto, line 17: why not watch all namespaces by default?

services/rook-ceph-cluster/1.10.1/defaults/cm.yaml
```diff
@@ -0,0 +1,139 @@
...
+                      values:
+                        - rook-ceph-osd
+                        - rook-ceph-osd-prepare
+            volumeClaimTemplates:
+              - metadata:
+                  name: data
+                spec:
+                  resources:
+                    requests:
+                      storage: 20Gi
```
cbuto, line 83: so this means 3 PVs at 20 GI each? but that is split between all the services that use this storage aka Loki and Velero? 

we might need to bump the default but I think the plan was to revisit this cluster config anyway in another ticket

(sorry if we discussed this already 🙈)

services/rook-ceph-cluster/1.10.1/defaults/cm.yaml
```diff
@@ -0,0 +1,139 @@
+---
```
cbuto, line 1: insights has a lot of critical Polaris findings that we should address if possible

services/rook-ceph/1.10.1/defaults/cm.yaml
takirala: we can test deploying another cephcluster in another namespace or even same namespace and once everything works, we can flip this boolean.

services/rook-ceph-cluster/1.10.1/defaults/cm.yaml
takirala: yep, we ll test and increase this once we actually integrate loki, velero and once we have some concrete data we can adjust this. nothing is set in stone with this PR, we just dont have enough data to come up with an ideal value.

services/rook-ceph-cluster/1.10.1/defaults/cm.yaml
takirala: created https://d2iq.atlassian.net/browse/D2IQ-93353 to resolve this

services/rook-ceph/1.10.1/defaults/cm.yaml
cbuto: sounds good to me

cbuto (approved):

gracedo (approved):

PR #631, opened 2022-09-26
fix: patch gatekeeper deployment to support FIPS clusters

cbuto:
For components, like webhooks, that interact with FIPS enabled Kubernetes components must use TLS 1.2 as the min supported version (TLS 1.3 does not work).

https://d2iq.atlassian.net/browse/D2IQ-92981

```release-note

```


services/gatekeeper/3.8.2/release/release.yaml
```diff
@@ -26,10 +26,24 @@ spec:
...
     - kind: ConfigMap
-      name: gatekeeper-3.8.1-d2iq-defaults
+      name: gatekeeper-3.8.2-d2iq-defaults
     - kind: ConfigMap
       name: gatekeeper-overrides
       optional: true
+  postRenderers:
+    - kustomize:
+        # Gatekeeper set the min TLS version to 1.3 in 3.8.0 but FIPS enabled clusters do not support TLS 1.3,
+        # In 3.8.1, the --tls-min-version flag is not configurable so we
```
msdolbey, line 36: I found the tiniest nit here imaginable. BUT, I do think this is a comma situation here.

```suggestion
        # In 3.8.1, the --tls-min-version flag is not configurable, so we
```

cbuto: my IDE even told me to do that but I ignored it lol, fixed in 62361137b9a23dcca432a792ab7b12d5aa682ed9

msdolbey (approved):
Testing thus far has been looking positive 
I think good right

cbuto:
looks like bloodhound is not happy with using postRenders, going to look to see if that's an easy fix

```
      e2e_suite_test.go:48: failed to load runtime config: failed get application versions for github.com/mesosphere/kommander-applications.git?ref=10989af81a2687a95a8cd3df21eaf84b97d6c682: *helm.Parser: postRenderers not yet implemented
```

gracedo (approved):
It's too bad we can't have this change only apply on FIPS enabled clusters.. (probably not an easy way to do this) only other note is to make sure upgrades work, nothing weird would happen with downgrading tls in an existing cluster, for example (this change would apply to all clusters, not just FIPS)

Do we want to maybe just apply this to the `release-2.3` branch for now? There's an [automated PR](https://github.com/mesosphere/kommander-applications/pull/634) open to bump gatekeeper to 3.9.0, we can maybe try testing this with that version first to see if the issue still exists in the latest version. [RN](https://github.com/open-policy-agent/gatekeeper/releases/tag/v3.9.0) don't call out anything that seems to fix this, but we should probably try the bump without this post-renderer "hack" first. We have some time until 2.4

msdolbey:
> It's too bad we can't have this change only apply on FIPS enabled clusters.. (probably not an easy way to do this) only other note is to make sure upgrades work, nothing weird would happen with downgrading tls in an existing cluster, for example (this change would apply to all clusters, not just FIPS)
> 
> Do we want to maybe just apply this to the `release-2.3` branch for now? There's an [automated PR](https://github.com/mesosphere/kommander-applications/pull/634) open to bump gatekeeper to 3.9.0, we can maybe try testing this with that version first to see if the issue still exists in the latest version. [RN](https://github.com/open-policy-agent/gatekeeper/releases/tag/v3.9.0) don't call out anything that seems to fix this, but we should probably try the bump without this post-renderer "hack" first. We have some time until 2.4

How hard would it be to automatically detect if something is FIPS-ey and then solely apply this if it's fips-enabled?
If there's something that's relatively easy, then there are all sorts of places that say if the cluster if FIPS-ey or not from provisioning. Or would that be a big thing to do?
The other way is to just say - if a user has a FIPS enabled cluster, they will need to 
- use such and such a configuration that points to this app + version in particular, but I know we're trying to shy away from "workarounds"

cbuto:
> It's too bad we can't have this change only apply on FIPS enabled clusters.. (probably not an easy way to do this) only other note is to make sure upgrades work, nothing weird would happen with downgrading tls in an existing cluster, for example (this change would apply to all clusters, not just FIPS)

Agreed, if it was exposed via the Helm chart we could just document that configuration change for FIPS clusters.

> Do we want to maybe just apply this to the `release-2.3` branch for now? There's an [automated PR](https://github.com/mesosphere/kommander-applications/pull/634) open to bump gatekeeper to 3.9.0, we can maybe try testing this with that version first to see if the issue still exists in the latest version. [RN](https://github.com/open-policy-agent/gatekeeper/releases/tag/v3.9.0) don't call out anything that seems to fix this, but we should probably try the bump without this post-renderer "hack" first. We have some time until 2.4

yeah we can apply it to 2.3 first then maybe before 2.4 we will have a new chart that supports setting the TLS min version (I opened an upstream issue to track that). 

In 3.9.0, the [min version](https://github.com/open-policy-agent/gatekeeper/blob/v3.9.0/pkg/webhook/common.go#L60) is still defaulted to TLS 1.3 and the chart doesn't support adding additional args or setting the tls-min-version flag.

cbuto:
recreated this PR against 2.3 - https://github.com/mesosphere/kommander-applications/pull/642

takirala (approved):

cbuto:
holding for https://d2iq.atlassian.net/browse/D2IQ-93143

gracedo:
superseded by https://github.com/mesosphere/kommander-applications/pull/668

PR #646, opened 2022-09-27
feat: Bump gitea to 6.0.1

mesosphere-teamcity:
This upgrades gitea to 6.0.1 to use the latest upstream chart gitea 6.0.1.


gracedo:
@mesosphere-mergebot test all

gracedo:
@mesosphere-mergebot test all

gracedo (approved):

services/gitea/6.0.1/defaults/kustomization.yaml
```diff
@@ -1,5 +1,4 @@
----
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-- cm.yaml
+  - cm.yaml
```
mikolajb, line 4: Why? 🙂

mikolajb (approved):

services/gitea/6.0.1/defaults/kustomization.yaml
gracedo: I think the automation is doing some reformatting but it'll probably just happen this first bump 😅

PR #652, opened 2022-09-27
feat: Bump reloader to 0.0.119

mesosphere-teamcity:
This upgrades reloader to 0.0.119 to use the latest upstream chart reloader v0.0.119.


gracedo:
https://github.com/mesosphere/kommander/pull/2389 is ✅

services/reloader/0.0.119/defaults/cm.yaml
```diff
@@ -0,0 +1,8 @@
+---
```
cwyl02, line 1: There seems to be some difference on the value of `values.yaml` here

gracedo:
superseded by https://github.com/mesosphere/kommander-applications/pull/675

services/reloader/0.0.119/defaults/cm.yaml
gracedo: it looks like the only difference is that the cm is no longer broken up into newlines

PR #657, opened 2022-09-27
feat: Bump traefik to 10.25.0

mesosphere-teamcity:
This upgrades traefik to 10.25.0 to use the latest upstream chart traefik 10.25.0.


services/kommander/0.4.0/defaults/cm.yaml
```diff
@@ -67,11 +67,11 @@ data:
       defaultEnterpriseApps:
         centralized-kubecost: "0.27.0"
         kubecost: "0.27.0"
-        kubecost-thanos-traefik: "0.0.1"
+        kubecost-thanos-traefik: "10.25.0"
```
bcmendoza, line 70: need to prevent these types of changes

bcmendoza (changes requested):

bcmendoza:
Need to change how kommander defaults are set

PR #678, opened 2022-10-03
feat: Bump traefik to 10.30.1

mesosphere-teamcity:
This upgrades traefik to 10.30.1 to use the latest upstream chart traefik 10.30.1.


bcmendoza:
Need to change how kommander defaults are set

PR #684, opened 2022-10-04
feat: use velero from upstream chart backed by ceph

takirala:
**What problem does this PR solve?**:

- Remove velero chart reference from `m/charts` and replaces it with upstream velero chart.
- Bumps velero app version from 3.2.3 to 3.3.0
   - This bumps velero chart from version to `2.31.8` (from [`1.5.2`](https://github.com/mesosphere/charts/blob/582cbf47a04b9f244e34caf4a558710b2eb97d24/staging/velero/Chart.yaml#L2)  we use in m/charts currently).
   - This is a new chart and has Apache License https://github.com/vmware-tanzu/helm-charts (No attribution change).

TODO:

- [ ] e2e tests needs to be updated. See https://github.com/mesosphere/kommander-e2e/pull/519

Depends on 
- #685 
- https://github.com/mesosphere/kommander-e2e/pull/519

**Which issue(s) does this PR fix?**:

https://d2iq.atlassian.net/browse/D2IQ-92897

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**If the PR adds a version bump, does it add a breaking change in License**:

- [x] No License Change (or NA).

PR #685, opened 2022-10-04
feat: `ObjectBucketClaim`s from `rook-ceph-cluster` service

takirala:

Creates `ObjectBucketClaims` for workspace level loki and velero services by default. The buckets are consumed by loki and velero in their configurations.

TODO:

- [x] Depends on https://github.com/mesosphere/charts/pull/1356
- [x] remove reference to https://takirala.github.io/charts/stable after merging above

 

https://d2iq.atlassian.net/browse/D2IQ-92897

```release-note

```

### DIFF
--- a/services/rook-ceph-cluster/1.10.3/objectbucketclaims/helmrelease.yaml
+++ b/services/rook-ceph-cluster/1.10.3/objectbucketclaims/helmrelease.yaml
@@ -12,7 +12,7 @@ spec:
         kind: HelmRepository
         name: mesosphere.github.io-charts-stable
         namespace: kommander-flux
-      version: 0.1.0
+      version: 0.1.1
   interval: 15s
   install:
     crds: CreateReplace
--- a/services/rook-ceph-cluster/1.10.3/objectbucketclaims/helmrelease.yaml
+++ b/services/rook-ceph-cluster/1.10.3/objectbucketclaims/helmrelease.yaml
@@ -1,0 +1,33 @@
+---
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
+kind: HelmRelease
+metadata:
+  name: object-bucket-claims
+  namespace: ${releaseNamespace}
+spec:
+  chart:
+    spec:
+      chart: object-bucket-claim
+      sourceRef:
+        kind: HelmRepository
+        name: mesosphere.github.io-charts-stable
+        namespace: kommander-flux
+      version: 0.1.0
+  interval: 15s
+  install:
+    crds: CreateReplace
+    remediation:
+      retries: 30
+    createNamespace: true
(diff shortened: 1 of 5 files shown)


services/rook-ceph-cluster/1.10.3/objectbucketclaims/helmrelease.yaml
```diff
@@ -0,0 +1,33 @@
...
+  upgrade:
+    crds: CreateReplace
+    remediation:
+      retries: 30
+  releaseName: object-bucket-claims
+  valuesFrom:
+    - kind: ConfigMap
+      name: rook-ceph-cluster-1.10.3-d2iq-defaults
+    - kind: ConfigMap
+      name: rook-ceph-cluster-overrides
```
cbuto, line 31: would users be able to configure the OBC with cluster-scoped configuration? (I don't think so unless we can hardcode another configmap name here, right?)

gracedo: hmm.. I guess we should add it here then 😅 it'll be called `rook-ceph-cluster-cluster-overrides` will have to just deal with that stutter :p

takirala: good catch, add another optional entry.

cwyl02 (approved):

services/rook-ceph-cluster/1.10.3/objectbucketclaims/helmrelease.yaml
```diff
@@ -0,0 +1,36 @@
...
+  namespace: ${releaseNamespace}
+spec:
+  chart:
+    spec:
+      chart: object-bucket-claim
+      sourceRef:
+        kind: HelmRepository
+        name: mesosphere.github.io-charts-stable
+        namespace: kommander-flux
+      version: 0.1.1
```
gracedo, line 15: Bump

gracedo (approved):

services/rook-ceph-cluster/1.10.3/objectbucketclaims/helmrelease.yaml
takirala: will do this in a follow up PR

takirala:
Merging this will temporarily break CI. This PR is part of the following dependency chain:

https://github.com/mesosphere/kommander/pull/2411
https://github.com/mesosphere/kommander-cli/pull/739
https://github.com/mesosphere/kommander-applications/pull/684
https://github.com/mesosphere/kommander-applications/pull/685
https://github.com/mesosphere/kommander-e2e/pull/519

PR #687, opened 2022-10-05
chore: clean up kubetools vars

cbuto:
Cleans up some unused substitution variables.

We may want to bump some of these images in the future but I intentionally left that out for now 

https://d2iq.atlassian.net/browse/D2IQ-87946

```release-note

```

### DIFF
--- a/services/istio/1.14.1/defaults/cm.yaml
+++ b/services/istio/1.14.1/defaults/cm.yaml
@@ -25,5 +25,5 @@ data:
         labels:
           prometheus.kommander.d2iq.io/select: "true"
     global:
-      image: "${kubetoolsImageRepository:=bitnami/kubectl}"
-      tag: "${kubetoolsImageTag:=1.23.6}"
+      image: bitnami/kubectl
+      tag: 1.23.6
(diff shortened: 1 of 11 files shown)


cwyl02 (approved):

services/istio/1.14.1/defaults/cm.yaml
```diff
@@ -25,5 +25,5 @@ data:
         labels:
           prometheus.kommander.d2iq.io/select: "true"
     global:
-      image: "${kubetoolsImageRepository:=bitnami/kubectl}"
-      tag: "${kubetoolsImageTag:=1.23.6}"
+      image: bitnami/kubectl
+      tag: 1.23.6
```
gracedo, line 29: we should bump this to be the same everywhere to avoid multiple versions of kubectl, however it's not strictly related to this change so maybe we can do it in a follow-up and see if it breaks istio in any way

cbuto: yeah definitely, I can follow up with that change

cbuto: actually.... our CI doesn't actually deploy Istio, so if we do update it I will have to do some manual testing 😓

PR #711, opened 2022-10-13
feat: adds dependencies in metadata

farhan5900:
It adds a new `dependencies` field in the `metadata.yaml` file for all the services that have dependencies on other services.

https://d2iq.atlassian.net/browse/D2IQ-93535

```release-note

```

### DIFF
--- a/services/centralized-kubecost/metadata.yaml
+++ b/services/centralized-kubecost/metadata.yaml
@@ -1,3 +1,5 @@
 type: internal
 scope:
   - workspace
+dependencies:
+  - traefik
--- a/services/dex-k8s-authenticator/metadata.yaml
+++ b/services/dex-k8s-authenticator/metadata.yaml
@@ -1,3 +1,7 @@
 type: internal
 scope:
   - workspace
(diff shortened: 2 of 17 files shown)


mstruebing (approved):
I can not judge the dependencies itself, but the format/way of describing them looks good to me.

Azhovan (approved):

gracedo (approved):
Fyi @clintonmedbery

gracedo:
@clintonmedbery @mstruebing some of these have kommander app as a dep, the UI can maybe choose whether or not to display that particular app, it might be extraneous to since of the user is able to view the UI, then kommander is obviously installed. These deps also include what we have been calling “internal” apps, which I’m not sure if the UI has decided to expose. I remember there was a story to expose them for the mgmt cluster but I also recall that story might have been reverted 🤔 anyways, might be worth a discussion whether it makes sense for the UI to display certain apps as dependencies in the info/warning, if those apps aren’t exposed in the UI in the first place to enable or disable

PR #722, opened 2022-10-17
[release-2.2] refactor: move loki HR and tenant behind kustomizations

cbuto:
**What problem does this PR solve?**:
This PR moves the Loki HR and tenant behind a kustomization to solve a few upgrade issues

**Which issue(s) does this PR fix?**:

https://d2iq.atlassian.net/browse/D2IQ-93696

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**If the PR adds a version bump, does it add a breaking change in License**:

- [ ] No License Change (or NA).

PR #727, opened 2022-10-18
revert: add upgrade job for grafana-loki

cbuto:
Since we need https://github.com/mesosphere/kommander-applications/pull/722 anyways, that will also take care of the issue this PR was meant to solve so lets revert it


cwyl02 (changes requested):
amend the commit message to conventional

cwyl02 (approved):

PR #735, opened 2022-10-19
fix: make sure to reload upon s3 secret refresh

cwyl02:
not sure why the looks like the s3 bucket secret can change after kommander install.

https://d2iq.atlassian.net/browse/D2IQ-92639

```release-note

```

### DIFF
--- a/services/grafana-loki/0.48.6/defaults/cm.yaml
+++ b/services/grafana-loki/0.48.6/defaults/cm.yaml
@@ -8,6 +8,8 @@ data:
   values.yaml: |
     loki:
       ingesterFullname: loki-ingester
+      annotations:
+        secret.reloader.stakater.com/reload: dkp-loki
 
       config: |
         auth_enabled: false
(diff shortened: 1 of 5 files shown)


services/grafana-loki/0.48.5/defaults/cm.yaml
```diff
@@ -8,6 +8,8 @@ data:
   values.yaml: |
```
cbuto, line 8: a bit unrelated, but I realized that we probably have to bump the Loki k-apps version before 2.4 is released to avoid upgrading the workspace/project Loki when the management cluster is upgraded

cbuto (approved):

takirala (approved):

takirala:
Created https://d2iq.atlassian.net/browse/D2IQ-93778 to bump loki version

takirala:
nvm, pushed 0d31d6f6ef7cd0ab1e50af20824466b7161b049d to add reloader annotation for velero :)

PR #737, opened 2022-10-20
chore: erasure coded replication for ceph object store

takirala:

Add erasure coded config for object store dataPool - See https://ceph.io/en/news/blog/2015/ceph-erasure-coding-overhead-in-a-nutshell/ to understand more about storage overhead.

With this PR we have parity with 2.3.x:
- Default capacity of storage is 160G. (we have 33% storage overhead, so this is ~106G usable space.)
  - For reference, In 2.3.x we shipped with 80G x 3 (1xvelero, 1xgrafana-loki, 1xproject-grafana-loki) = 240G total space. And storage overhead was 50% so we had 120G usable space.
- Bumps the number of replicas for rados gateway pods. these are responsible for providing the S3 API on top of OSDs(object storage daemons that are responsible for raw storage).
- Caps the project logging at 5G. This is easily overridable by user when deploying project logging stack. I picked this value so that we should be able to deploy multiple projects easily.

```release-note

```

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -26,12 +26,12 @@ kommander-e2e:
 	@if [ -d $(KOMMANDER_E2E_DIR) ] ; then \
 		cd $(KOMMANDER_E2E_DIR) && \
 			git fetch origin && \
-			git reset --hard tga/ceph-defaults ; \
+			git reset --hard origin/main ; \
 	else \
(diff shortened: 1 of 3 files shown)


cbuto (approved):

takirala:
CI was :heavy_check_mark: at a8220eaebc30355c8c37162e7ed62b24ccf42b3e admin merging after reverting the custom branch ref.

PR #746, opened 2022-10-21
feat: remove deprecated metallb

jimmidyson:
Cleans up technical debt.

```release-note

```


gracedo:
`metallb` is technically deprecated, the only reason we haven't removed it from k-apps yet is that we hadn't been able to prioritize fixing our CI https://d2iq.atlassian.net/browse/D2IQ-87137 - so maybe this is a good time to prioritize it :)

jimmidyson:
@gracedo I love that idea! That would mean changing the existing kuttl tests - do we want to do that or wait until they become kommander-e2e tests and remove kuttl tests completely?

gracedo:
> That would mean changing the existing kuttl tests - do we want to do that or wait until they become kommander-e2e tests and remove kuttl tests completely?

Unfortunately I don't think moving the kuttl tests out into e2e will be prioritized anytime soon, but maybe we can discuss more during 2.5 planning to see which EDIs (I think that's the right acronym 😆 ) we decide to pull in

jimmidyson:
Metallb <0.13.5 does not support Kubernetes v1.25, which is required for DKP 2.5. Our kuttl tests deploy metallb in attached clusters. Metallb 0.13 does not support configmap based configuration which is what we use in our kuttl tests on attached clusters. One option would be to create a new helm chart (and associated kommander application) that would be deployed on attached clusters, but we do not want to support metallb any longer via a kommander application. I guess the easiest way to handle this is to actually deploy metallb as part of the cluster provisioning in the kuttl tests instead of deploying it via a kommander appdeployment.

takirala (approved):

gracedo (approved):

PR #755, opened 2022-10-25
feat: flux kustomizations for helm repositories

takirala:

Consider the 2.3.x to 2.4.x upgrade scenario:

- MinIO would be running in 2.3.x
- We start upgrade to 2.4.x
- app_repo_upgrader in CLI would "merge" 2.4.x k-apps repo into the gitea repo of 2.3.x cluster
- This would override the `common/helm-repositories/kustomization.yaml` file to remove `minio.yaml`
- This causes the user's MinIO deployment to be "stuck" cuz the `HelmRepository` disappears from the Cluster (it is still present in gitea repo)

To solve above, this PR replaces the `kustomization.yaml` with flux's Kustomization as it can intelligently load all the contents of a directory dynamically. 

Alternative of https://github.com/mesosphere/kommander-cli/pull/795

https://d2iq.atlassian.net/browse/D2IQ-93885

```release-note

```

**Checklist**


common/airgapped/kustomization.yaml
```diff
@@ -1,8 +1,26 @@
...
-      kind: HelmRepository
+metadata:
+  name: kommander-helm-repositories
+  namespace: kommander-flux
+spec:
+  prune: true
+  wait: true
+  interval: 6h
+  retryInterval: 1m
+  path: ../helm-repositories
```
gracedo, line 11: in case something doesn't quite work, I think I remember relative paths not working for flux kustomizations, you might have to use the root path

takirala: thanks, updated!

takirala:
@d2iq-mergebot test all

takirala:
@mesosphere-mergebot test all

takirala:
closing in favor of https://github.com/mesosphere/kommander-cli/pull/795

PR #758, opened 2022-10-26
fix: Override KPS tlsMinVersion to 1.2

gracedo:
KPS prometheus-operator was using tls1.3 by default, set `tlsMinVersion` to 1.2

https://d2iq.atlassian.net/browse/D2IQ-93914

```release-note

```

**Checklist**


msdolbey (approved):
👍 
Yes yes, quite good

takirala (approved):

PR #760, opened 2022-10-27
fix: increase requests to flux avoid misfunctioning

palexster:

This PR increases CPU and Memory requests for `kustomize` and `source` controllers to avoid observed throttling and protect those controllers from OOM kills.

https://d2iq.atlassian.net/browse/D2IQ-93893


Azhovan (approved):

mikolajb (approved):

cbuto (changes requested):
we'll need to add these patches here: https://github.com/mesosphere/kommander-applications/blob/main/hack/flux/flux-update-kustomization.yaml#L5-L29 so they are retained after a Flux upgrade

takirala (changes requested):
These changes needs to be written in the patch file we apply after auto-updating flux. Otherwise, these changes will get overwritten in a future auto flux update.
https://github.com/mesosphere/kommander-applications/blob/main/hack/flux/flux-update-kustomization.yaml

palexster:
@takirala @cbuto got it, thanks a lot!

takirala (commented):
What i wanted to say was - you need to add changes in the patch file && the actual manifests of source controller and kustomize controller.

In the future automated flux upgrade PRs, the script ll apply the patch to the manifests but we need to do it manually just this once in the current PR. Sorry for the confusion :D

gracedo:
@takirala @cbuto @palexster I pushed the changes to the manifests, double check?

palexster:
Thanks, @gracedo ! :pray:

PR #761, opened 2022-10-27
feat: use an official ngc image for the validator

faiq:
Removes the use of a "forked" container for an official container instead 

```release-note

```

**Checklist**


services/nvidia-gpu-operator/1.11.1/defaults/cm.yaml
```diff
@@ -28,11 +28,9 @@ data:
       enabled: true
       version: 2.4.6-2.6.10-ubuntu20.04
     validator:
-      # we are using our own validator pod that
+      # we are newer version of the validator pod that
```
gracedo, line 31:
```suggestion
      # we are using a newer version of the validator pod that
```

gracedo (approved):

cwyl02 (approved):

PR #762, opened 2022-10-27
fix: use default value for useSecret configuration

cbuto:
This PR solves a potential upgrade problem if Velero is configured to use AWS S3 for storage. In 2.3, `useSecret` was set to `true` (the default in the chart). In 2.4, we set it to `false` so an existing configuration that relied on that being `true` would break upon upgrade. 

We can keep the chart default value of `true` and also still be able to leverage `extraSecretRef` to load the Ceph creds into Velero via env vars. 

```release-note

```

**Checklist**

### DIFF
--- a/services/velero/3.3.0/defaults/cm.yaml
+++ b/services/velero/3.3.0/defaults/cm.yaml
@@ -20,7 +20,6 @@ data:
         config:
           region: "fallback"
     credentials:
-      useSecret: false
       # This is created by rook-ceph-cluster service. A ConfigMap and a Secret with same name as bucket are created.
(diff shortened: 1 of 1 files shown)


cbuto:
the Velero kuttl test passed at https://github.com/mesosphere/kommander/pull/2521

I also validated the `backupstoragelocation` object is in the `Available` state after install/upgrade which means the Ceph credentials are valid and Velero can access the Ceph bucket.

takirala (approved):

cwyl02 (approved):

gracedo (approved):

PR #764, opened 2022-10-28
fix: Update license mapping for kommander UI

jimmidyson:

```release-note

```

**Checklist**


vacas5:
will we need to keep this in sync with every kommander-ui version?

PR #773, opened 2022-10-31
feat: Bump rook-ceph to 1.10.4

mesosphere-teamcity:
This upgrades rook-ceph to 1.10.4 to use the latest upstream chart rook-ceph v1.10.4.


takirala:
loopback device support has been merged and will be part of next release, we can update to v1.10.6 once its out

PR #782, opened 2022-11-03
fix: use 8.5.14 for grafana-logging

cwyl02:
to workaround upstream issue for crash when looking at audit dashboard

https://d2iq.atlassian.net/browse/D2IQ-94038

```release-note

```

**Checklist**

### DIFF
--- a/services/grafana-logging/6.38.1/defaults/cm.yaml
+++ b/services/grafana-logging/6.38.1/defaults/cm.yaml
@@ -8,6 +8,12 @@ metadata:
 data:
   values.yaml: |
     ---
+    # overriding the default image due to this issue
+    # https://d2iq.atlassian.net/browse/D2IQ-94038
+    # cause by upstream bug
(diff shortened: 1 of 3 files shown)


services/grafana-logging/6.38.1/defaults/cm.yaml
```diff
@@ -8,6 +8,12 @@ metadata:
 data:
   values.yaml: |
     ---
+    # overriding the default image due to this issue
+    # https://d2iq.atlassian.net/browse/D2IQ-94038
+    # cause by upstream bug
+    # https://github.com/grafana/grafana/issues/56087
+    image:
+      tag: 8.5.14
```
gracedo, line 16: think we need to make the same change in `project-grafana-logging` too

cwyl02: done

gracedo (approved):

PR #783, opened 2022-11-03
fix: Fix cluster label var for rook ceph dashboard

gracedo:
`label_values(cluster)` would be grabbing every possible value for the label `cluster` across all metrics. The label that ceph metrics use is actually called `rook_cluster`. We should be looking at an actual ceph metric and grabbing the possible values of the `rook_cluster` label instead. (I'm not sure how this worked in the past, maybe some other metrics somwhere happened to set a `cluster` label to `kommander` at some point?)

From https://grafana.com/docs/grafana/latest/datasources/prometheus/#query-variable
image

Also, noticed that duplicate metrics were getting scraped - looks like we were already scraping ceph metrics via our `kubernetes-pods` scrapeconfig which is scraping pods with the deprecated prometheus scrape annotations. We should disable that scrape config but in interest of time for this release, we will leave it and instead disable the service monitor deployed by the ceph chart, to avoid dup metrics.

image
image

```release-note

```

**Checklist**

### DIFF
--- a/services/rook-ceph-cluster/1.10.3/defaults/cm.yaml
+++ b/services/rook-ceph-cluster/1.10.3/defaults/cm.yaml
@@ -177,7 +177,7 @@ data:
           reclaimPolicy: Delete
 
     monitoring:
-      enabled: true
+      enabled: false
       createPrometheusRules: true
 
     #################################################################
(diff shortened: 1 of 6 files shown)


msdolbey (approved):
👍 
I suppose this is a rook

takirala (approved):
Good catch!

gracedo:
Added change to disable ceph service monitor

cwyl02 (approved):

PR #791, opened 2022-11-08
fix: Update dashboard app versions, bump Kiali img

gracedo:
- Just noticed that we are overriding kiali image for some reason, and that the chartbumper missed updating the image. I've removed the image override, not sure why it was there but we should always be using the version deployed with the chart (chart version looks to be aligned with img version).
- Also noticed that we missed bumping most of the app versions we are included in dashboard configmaps.

We'll need to cut another RC with these fixes.

```release-note

```

**Checklist**

### DIFF
--- a/licenses.d2iq.yaml
+++ b/licenses.d2iq.yaml
@@ -337,7 +337,7 @@ resources:
       - url: https://github.com/jetstack/kube-oidc-proxy
         ref: ${image_tag}
         license_path: LICENSE
-  - container_image: quay.io/kiali/kiali-operator:v1.52.0
+  - container_image: quay.io/kiali/kiali-operator:v1.57.1
     sources:
       - url: https://github.com/kiali/kiali-operator
         ref: ${image_tag}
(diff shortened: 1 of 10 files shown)


takirala (approved):

cwyl02 (approved):

takirala:
Kinda related, what should be the version here https://github.com/mesosphere/kommander-applications/blob/main/services/rook-ceph-cluster/1.10.3/helmrelease/dashboard-cm.yaml#L13 ? Should we update that to be in sync with helm chart version (or keep it as is to reflect the base tech - that is - ceph version)

gracedo:
> Kinda related, what should be the version here https://github.com/mesosphere/kommander-applications/blob/main/services/rook-ceph-cluster/1.10.3/helmrelease/dashboard-cm.yaml#L13 ? Should we update that to be in sync with helm chart version (or keep it as is to reflect the base tech - that is - ceph version)

@takirala Should be the actual application / base tech version, which is not necessarily the same as helm chart (e.g. KPS is the best example, it has prom, alertmanager and grafana dashboards, which are versioned by the actual app versions)
image

gracedo:
I'll hold on merging until it's agreed that this is a blocker

cbuto (approved):

gracedo:
@d2iq-mergebot backport main

PR #832, opened 2022-11-23
build(deps): use go 1.19

aweris:

Bump go version to 1.19 for release hack

--

--

--

**Checklist**

Depends on #831


mikolajb:
> The go test failure related with `make go-lint` target executed prerequisite of the test. It was resolved by #833 with bumping proper versions to work with go 1.19

Why not reversing the order?

mikolajb (approved):

cbuto (approved):

aweris:
> Why not reversing the order?

No particular reason.  This is my resolve order. Maybe I should try to reverse the order when merging it

PR #833, opened 2022-11-23
build(deps): bump tool versions

aweris:

Bump tool versions 

--

--

--

**Checklist**

Depends on #832


mikolajb (approved):

.tool-versions
```diff
@@ -1,9 +1,9 @@
-awscli 2.4.24
-flux2 0.36.0
-github-cli 2.13.0
+awscli 2.9.0
+flux2 0.37.0
```
cbuto, line 2: The `flux2` is handled by this script https://github.com/mesosphere/kommander-applications/blob/main/hack/flux/update-flux.sh, maybe we should leave it at the current version until that automation bumps it

PR #848, opened 2022-12-01
chore: update Kubernetes versions to check against

floridoo:
Updates the versions we want to be compatible with for DKP v2.5. Kubernetes 1.25 is disabled for now because some apps need to be bumped before being compatible.

https://d2iq.atlassian.net/browse/D2IQ-94278

```release-note

```

**Checklist**


cwyl02 (approved):

cbuto (approved):

floridoo:
Fix for the install-e2e-tests: https://github.com/mesosphere/kommander-applications/pull/849

PR #865, opened 2022-12-16
fix: use default fluentd image

cbuto:
The logging-operator [switch to using a new repository](https://github.com/banzaicloud/logging-operator/pull/1108/files#diff-911a398d9218f00dc5749daedc4d64b64f16f2f8b010c898dc28393b3454fd0dR134) for the config-reloader image so we should use that by default. 

We bumped the logging-operator which changed the default repo but since we override the tag, its trying to pull the wrong image.

More context: https://mesosphere.slack.com/archives/C03P4F14M7E/p1671204117585399

```release-note

```

**Checklist**

### DIFF
--- a/services/logging-operator/3.17.10/defaults/cm.yaml
+++ b/services/logging-operator/3.17.10/defaults/cm.yaml
@@ -57,8 +57,6 @@ data:
     tls:
       enabled: true
     fluentd:
-      configReloaderImage:
-        tag: v0.7.1
       image:
         # Explicitly use the default image. This should be updated when logging-operator is upgraded.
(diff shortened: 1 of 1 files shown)


cwyl02 (approved):

gracedo (commented):
Can we just switch to using the default and not override it at all now? (I don't remember why we overrode the image in the past)

cbuto:
> Can we just switch to using the default and not override it at all now? (I don't remember why we overrode the image in the past)

yeah I can do that instead 👍

gracedo (approved):

PR #904, opened 2023-01-23
feat: bump kustomize-controller concurrency to 15

cbuto:
Increases the `kustomize-controller` concurrency to 15, mainly to speed up the installation process. We have a ton of kustomizations (half of which are just ConfigMaps), so an easy win for speeding up the installation process is to increase the concurrency from 4 to 15. In a few tests, this saved 10-15 minutes while waiting for all the platform applications to be ready.

I made this change on the daily and saw some CPU spikes so I've bumped that as a extra precaution (memory seemed fine):

Screen Shot 2023-01-23 at 2 13 01 PM
Screen Shot 2023-01-23 at 2 13 09 PM

https://d2iq.atlassian.net/browse/D2IQ-95122

```release-note

```

**Checklist**

### DIFF
--- a/hack/flux/flux-update-kustomization.yaml
+++ b/hack/flux/flux-update-kustomization.yaml
@@ -12,6 +12,9 @@ patches:
       - op: add
         path: /spec/template/spec/containers/0/args/-
         value: --no-remote-bases=true
+      - op: add
+        path: /spec/template/spec/containers/0/args/-
+        value: --concurrent=15
       - op: replace
         path: /spec/template/spec/containers/0/resources/requests/memory
         value: 150Mi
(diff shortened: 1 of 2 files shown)


cwyl02 (approved):

palexster (approved):
This is great! 
Do you think a similar boost for AppManagement?

cbuto:
> This is great! It's a vast speed-up... Would a similar boost for AppManagement help too, in your opinion?

The AppDeployment controller (the main part of AppManagement) actually reconciles all AppDeployments at once, so we really only want to allow 1 concurrent reconciliation for that controller but I have considered if we should rethink how that works (aka reconcile one AppD at a time and allow many concurrent reconciliations).

takirala (approved):

PR #907, opened 2023-01-24
feat: add new chart for lifecycle controller

Azhovan:
As a part of cluster expansion project, a new controller is created that will be responsible for the lifecycle of the KommanderCore resource.
https://github.com/mesosphere/dkp-platform/pull/376

https://d2iq.atlassian.net/browse/D2IQ-95249

```release-note

```

**Checklist**


Azhovan:
Im re-opening it when I fix a problem i found out in the directory structure

PR #926, opened 2023-01-27
feat: Add Loki pre-install job

cbuto:
This PR adds a pre-install job that the grafana-loki HR will depend on.

The pre-install job is meant to improve the installation process by waiting for the CephCluster to become ready before installing the Loki HelmRelease. By waiting for the CephCluster to be ready, the Loki HelmRelease won't show as "failed" because we wait longer to start the initial install. 

https://d2iq.atlassian.net/browse/D2IQ-95122

```release-note

```

**Checklist**


services/rook-ceph-cluster/1.10.8/objectbucketclaims/helmrelease.yaml
```diff
@@ -12,6 +12,7 @@ spec:
         name: mesosphere.github.io-charts-stable
         namespace: kommander-flux
       version: 0.1.4
+  timeout: 20m
```
cbuto, line 15: I bumped this because the this HR was timing out while running the health check jobs (waiting for the OBCs to be ready). We won't really need this once we move away from the obc health check jobs

services/grafana-loki/0.48.7/grafana-loki-pre-install-jobs/pre-install.yaml
```diff
@@ -0,0 +1,99 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: grafana-loki-pre-install
+  namespace: ${releaseNamespace}
+rules:
+  - apiGroups: ["kustomize.toolkit.fluxcd.io"]
+    resources: ["kustomizations"]
+    verbs: ["get", "patch"]
+  - apiGroups: ["ceph.rook.io"]
```
takirala, line 17: (just making sure) i am assuming this will not fail when ceph crds are not installed on the cluster. correct me if that assumption is wrong.

services/rook-ceph-cluster/1.10.8/defaults/cm.yaml
```diff
@@ -257,7 +257,7 @@ data:
         enabled: true
         bucketName: dkp-loki
         storageClassName: dkp-object-store
-        enableOBCHealthCheck: true
```
takirala, line 260: why is this being flipped ?

takirala (approved):

sonofgibs (approved):

services/rook-ceph-cluster/1.10.8/defaults/cm.yaml
cbuto: This enables the job that waits for the OBC to be ready then reconciles the HR. 

We don’t really need this anymore since we don’t expect the Loki HR to fail. But as a safety net, I think I might keep it enabled (it won’t do any harm from what I can tell)

services/grafana-loki/0.48.7/grafana-loki-pre-install-jobs/pre-install.yaml
cbuto: yep these apply cleanly without the Ceph CRDs installed

PR #944, opened 2023-02-02
feat: Remove allowcrossnamespace from traefik configuration

mikolajb:
I found a comment that mentions to remove the settings after the migration. Because migration is not supported in the current version, let's try to remove it.

```release-note

```

**Checklist**


mikolajb:
@d2iq-mergebot test all

gracedo (approved):

takirala (approved):

PR #955, opened 2023-02-03
fix: Add necessary permissions to kommander-operator chart

mikolajb:
It was missing a few permissions

```release-note

```

**Checklist**

### DIFF
--- a/charts/kommander-operator/templates/deployment.yaml
+++ b/charts/kommander-operator/templates/deployment.yaml
@@ -29,11 +29,3 @@ spec:
             - name: http
               containerPort: 80
               protocol: TCP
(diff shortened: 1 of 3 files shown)


palexster (approved):

Azhovan (approved):

charts/kommander-operator/templates/deployment.yaml
```diff
@@ -29,11 +29,3 @@ spec:
             - name: http
               containerPort: 80
               protocol: TCP
-          livenessProbe:
```
palexster, line 32: In a future PR: should we add again a liveness probe?

PR #963, opened 2023-02-06
feat: bump Loki version to 0.48.20

cbuto:
Bumps Loki to a newer version than what will be in v2.4.1 (after https://github.com/mesosphere/kommander-applications/pull/942 is backported)

```release-note

```

**Checklist**


cbuto:
test bump in https://github.com/mesosphere/kommander/pull/2933

gracedo (approved):

cwyl02 (approved):

takirala:
Hmm, what happens if we ship another fix in 2.4.3 for loki and want to upgrade from DKP 2.4.3 -> 2.5.0 ? I dont think its an overkill to think about such a situation now :thinking: 

One option could be this PR can bump loki version to `0.48.20` or something similar. We already digressed from upstream version anyway. Just thinking out loud.

cbuto:
> Hmm, what happens if we ship another fix in 2.4.3 for loki and want to upgrade from DKP 2.4.3 -> 2.5.0 ? I dont think its an overkill to think about such a situation now 🤔
> 
> One option could be this PR can bump loki version to `0.48.20` or something similar. We already digressed from upstream version anyway. Just thinking out loud.

yeah that's a really good point, i think we have this situation for other apps already too. I can bump to `0.48.20` so we have a bunch of versions to use. This also reminds me of the k-apps versioning convo we've had in the past.

cwyl02 (approved):

takirala (approved):

PR #965, opened 2023-02-06
fix: restore admission controller enabled status in ceph

takirala:

https://d2iq.atlassian.net/browse/D2IQ-95736

```release-note

```

**Checklist**


takirala:
This approach does not work because ceph is still flaking on how long does it take for the configmaps to be propagated.

PR #968, opened 2023-02-06
feat: wait for ceph operator in ceph cluster preupgrade

takirala:

Alternative of #965 

With current ceph upgrade of 1.10.6 -> 1.10.10 which is flaking, i think the problem is : 

```
t=0 | everything was healthy, both rook-ceph and rook-ceph-cluster were running at 1.10.6
t=1 | kicked off upgrade to 1.10.10 for both rook-ceph and rook-ceph-cluster  HelmReleases
t=2 | rook-ceph was in process of deleting the webhook.
t=2 | (same time as above) rook-ceph-cluster helm upgrade tried to mutate CephCluster and other resources. The ValidatingWebhookConfiguration is still present but probably the pod/service is being terminated.

The problem is last two actions happen at t=2 . instead rook-ceph-cluster should be upgraded AFTER completing the upgrade of rook-ceph.
```

Based on how the last two actions happen, we hit the flake (or not). This PR attempts to resolve the race by adding an optional dependency check. Note that its important for this to be optional because we have to support those cases where ceph cluster is managed by a ceph operator that is not owned/managed by DKP.

https://d2iq.atlassian.net/browse/D2IQ-95736

Sample logs of new container:

```
NAME        AGE   READY     STATUS
rook-ceph   30m   Unknown   Reconciliation in progress
Waiting for ceph operator to complete its upgrade...
Ceph operator is at v1.10.3 version and desired version is v1.10.10. Continuing to wait...
Ceph operator is at v1.10.3 version and desired version is v1.10.10. Continuing to wait...
Ceph operator is at same version as desired cluster version (v1.10.10). Exiting...
```

https://teamcity.mesosphere.io/repository/download/ClosedSource_Kommander2_KommanderApplications_UpgradeE2eTests/3819191:id/feature/upgrade/kind/singlecluster/support-bundle-2023-02-06T22_19_49.tar.gz!/support-bundle-2023-02-06T22_19_49/pod-logs/kommander/check-dkp-ceph-crd-qshjh-pre-upgrade.log

Or during fresh installs : 

```
NAME        AGE   READY     STATUS
rook-ceph   3s    Unknown   Reconciliation in progress
Waiting for ceph operator to complete its upgrade...
Ceph operator is at <no value> version and desired version is v1.10.10. Continuing to wait...
Ceph operator is at <no value> version and desired version is v1.10.10. Continuing to wait...
Ceph operator is at <no value> version and desired version is v1.10.10. Continuing to wait...
Ceph operator is at same version as desired cluster version (v1.10.10). Exiting...
```

Above log might look differently based on the order of how `rook-ceph` and `rook-ceph-cluster` gets installed.

https://teamcity.mesosphere.io/repository/download/ClosedSource_Kommander2_KommanderApplications_InstallE2eTests/3819193:id/feature/install/kindcluster/support-bundle-2023-02-06T22_00_07.tar.gz!/support-bundle-2023-02-06T22_00_07/pod-logs/kommander/check-dkp-ceph-crd-m76g9-pre-upgrade.log

```release-note

```

**Checklist**

### DIFF
--- a/services/rook-ceph-cluster/1.10.10/pre-install.yaml
+++ b/services/rook-ceph-cluster/1.10.10/pre-install.yaml
@@ -18,3 +18,5 @@ spec:
   postBuild:
     substitute:
       releaseNamespace: ${releaseNamespace}
+      # Update the following version whenever ceph-cluster service is bumped.
+      desiredCephOperatorVersion: v1.10.10
--- a/services/rook-ceph-cluster/1.10.10/pre-install/ceph-crd-check.yaml
+++ b/services/rook-ceph-cluster/1.10.10/pre-install/ceph-crd-check.yaml
@@ -14,6 +14,9 @@ rules:
   - apiGroups: ["apiextensions.k8s.io"]
     resources: ["customresourcedefinitions"]
     verbs: ["get", "watch", "list"]
+  - apiGroups: ["helm.toolkit.fluxcd.io"]
+    resources: ["helmreleases"]
+    verbs: ["get", "watch", "list"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
@@ -42,7 +45,7 @@ spec:
       serviceAccountName: check-dkp-ceph-crd
       restartPolicy: OnFailure
       containers:
-        - name: kubectl
+        - name: pre-install
           image: bitnami/kubectl:1.24.6
           command:
             - sh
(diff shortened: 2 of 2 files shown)


cwyl02 (approved):

gracedo (approved):
I wonder if this would have been easier if we created an umbrella helm chart with all the ceph charts 🥲

takirala:
> I wonder if this would have been easier if we created an umbrella helm chart with all the ceph charts

Supporting ceph with operator and ceph without operator (and multiple ceph clusters in both cases) would be the trickiest thing i can think of when desigining a single helm chart. But I guess thats another battle for another day :D

takirala:
@d2iq-mergebot backport release-2.4

PR #969, opened 2023-02-06
ci: Only run pr labeler on chartbump prs

gracedo:
Avoid creating kommander branch/pr for PRs that are not automated chart bump PRs

```release-note

```

**Checklist**


takirala (approved):

gracedo:
Had pushed a test commit bumping dir of `cert-manager` to test that the workflow skips adding the PR labels (`open-kommander-pr` and `do-not-merge/testing`) with the update to the workflow def. Verified that the step was skipped because the PR is not prefixed with `chartbump/` which all automated chart bump prs are prefixed with.

image

PR #972, opened 2023-02-07
fix: include kommander operator in airgap kz

cwyl02:
this should fix the problem where in airgap mode kommander-operator resources are missing

```release-note

```

**Checklist**


cwyl02:
after discussing with @mikolajb , this is not a viable solution

PR #974, opened 2023-02-07
feat: updated application required dependencies

nataliepina:
Closes D2IQ-95624

This updates applications that require dependencies to be enabled on them. Some applications already have the correct required dependencies set, however, others were missing these. 
These changes follow the required dependencies listed with applications here [Platform Application Dependencies](https://docs.d2iq.com/dkp/2.4/platform-application-dependencies)  and here [Project Platform Application Dependencies](https://docs.d2iq.com/dkp/2.4/project-platform-application-dependencies).

https://d2iq.atlassian.net/browse/D2IQ-95624

```release-note

```

**Checklist**

### DIFF
--- a/services/project-grafana-loki/metadata.yaml
+++ b/services/project-grafana-loki/metadata.yaml
@@ -5,6 +5,10 @@ category:
 type: platform
 scope:
   - project
+dependencies:
+  - project-logging
+  - grafana-loki
+  - logging-operator
 certifications:
   - certified
   - supported
--- a/services/grafana-logging/metadata.yaml
+++ b/services/grafana-logging/metadata.yaml
@@ -5,6 +5,8 @@ category:
 type: platform
 scope:
   - workspace
+dependencies:
+  - grafana-loki
 certifications:
   - certified
   - supported
--- a/services/grafana-loki/metadata.yaml
+++ b/services/grafana-loki/metadata.yaml
@@ -5,6 +5,8 @@ category:
 type: platform
 scope:
   - workspace
+dependencies:
+  - rook-ceph-cluster
 certifications:
   - certified
   - supported
--- a/services/kube-oidc-proxy/metadata.yaml
+++ b/services/kube-oidc-proxy/metadata.yaml
@@ -5,6 +5,9 @@ category:
 type: internal
 scope:
   - workspace
+dependencies:
+  - cert-manager
+  - traefik
 certifications:
   - certified
   - supported
--- a/services/kubernetes-dashboard/metadata.yaml
+++ b/services/kubernetes-dashboard/metadata.yaml
@@ -5,6 +5,8 @@ category:
 type: platform
 scope:
   - workspace
+dependencies:
+  - traefik
 certifications:
   - certified
   - supported
--- a/services/project-grafana-logging/metadata.yaml
+++ b/services/project-grafana-logging/metadata.yaml
@@ -5,6 +5,8 @@ category:
 type: platform
 scope:
   - project
+dependencies:
+  - project-grafana-loki
 certifications:
   - certified
   - supported
--- a/services/project-logging/metadata.yaml
+++ b/services/project-logging/metadata.yaml
@@ -5,6 +5,8 @@ category:
 type: platform
 scope:
   - project
+dependencies:
+  - logging-operator
 certifications:
   - certified
   - supported
--- a/services/rook-ceph-cluster/metadata.yaml
+++ b/services/rook-ceph-cluster/metadata.yaml
@@ -6,6 +6,8 @@ category:
 type: platform
 scope:
   - workspace
+dependencies:
+  - rook-ceph
 certifications:
   - certified
   - supported
--- a/services/traefik/metadata.yaml
+++ b/services/traefik/metadata.yaml
@@ -5,6 +5,9 @@ category:
 type: platform
 scope:
   - workspace
+dependencies:
+  - cert-manager
+  - reloader
 certifications:
   - certified
   - supported


gracedo:
Is this purely just for displaying metadata in the UI? We had originally meant for the field to indicate "hard" dependencies among our apps, that actually block other apps from deploying altogether if these dependencies are not deployed. What we have in the docs include these but also more "soft" dependencies where they may not block the dependent apps from being deployed, but more that it may not make sense or work as expected/properly without some of these dependencies.

We never came up with a design on how to use the field though, so I just want to make sure we understand how it is being used in the UI and whether we need to distinguish between "hard" dependencies vs "suggestions". There are also dependencies that only apply in the default config, but the config can also be changed to get rid of the dependency (I see in the docs we have some "optional" dependencies).

Basically, there is more to how we define these dependencies and I'm not sure how it might change the plan for however the UI will be using this field

nataliepina:
@gracedo 
Yes, to my knowledge, this is mainly for displaying the metadata in the UI. Though with the changes related to this epic we are going to prevent the user from enabling an app if they do not have these required dependencies enabled. We currently have some issues where the user is able to enable the app and then it will come back with an error later if it doesn’t have the required deps enabled. 
Is there anything in the documentation to distinguish between the hard requirements for dependencies and the suggestions? I do see there are some that have a message stating that it is optional, would that relate to the ones that are more of a suggestion? I haven’t included any of those optional ones in these changes, and only the ones that did not state it was optional.

@vacas5 do you have any additional input with your knowledge of the epic and the work that should be done here?

vacas5:
We were trying to use this list to prevent a user from enabling an application. IMO we should probably only list hard dependencies here. Maybe docs could change the table to have a separate column for dependency "nice to haves"?

gracedo:
@nataliepina If we want to only specify "hard" dependencies (defined in the backend) then the current state of the metadata files are all up to date

vacas5:
so this PR should be closed, @gracedo ?

services/project-grafana-loki/metadata.yaml
```diff
@@ -5,6 +5,10 @@ category:
 type: platform
 scope:
   - project
+dependencies:
+  - project-logging
+  - grafana-loki
+  - logging-operator
```
takirala, line 11: we can also add `rook-ceph-cluster` , but we can choose not to if we want to exclude transitive dependencies.

gracedo:
yes @vacas5 , sorry for the extra work @nataliepina !

services/project-grafana-loki/metadata.yaml
gracedo: I think we're sticking to keeping the dependencies defined here as "hard" dependencies where in the backend we actually prevent an app from being deployed altogether if the deps are not deployed (e.g. whenever we define `dependsOn`). Because for the other "soft" dependencies, they may only fail to deploy successfully on the default config. If we need both, I would vote to have them in separate fields

vacas5:
Okay so I think we are way confused at this point. Our goal in the UI is to prevent a user from getting into a difficult state if they try and enable an app before it's dependencies are enabled. Are you saying that the backend already has validation for this? Or is it only in the CLI?

gracedo:
@vacas5 there is no backend validation at all! But there are two different types of error states a user can find themselves in.
With "hard" dependencies, they will see an error in the helmrelease that it is waiting for a dependency to be ready first.
With "soft" dependencies, an app might not deploy successfully with the default config (e.g. loki and ceph are configured by default to be dependent on rook-ceph-cluster to up, but a user can configure them to NOT use ceph as their storage). Or maybe the app will come up "healthy" but some parts of the app may not work as expected (e.g. all apps with dashboards basically rely on traefik to be up and healthy in order to access their dashboards and any UIs, but otherwise the apps  in the backend are probably functioning fine).

I'm not sure how we want to be presenting these error scenarios to the user on our UI. The "hard" dependencies are just easier to work off of and restrict in the UI (at least for first iteration) because we know for sure that these apps will always fail to become ready unless their deps are deployed. For the other cases we probably need to spend more time to think about the possible scenarios and what the best UX would look like for those

nataliepina:
I think some of the confusion may be due to what is in the documentation. Perhaps we'll want to make the distinction there for users to reference. 

Are there actually 3 categories when it comes to dependencies?  Hard, soft, and optional? 
Screenshot 2023-02-09 at 1 52 13 PM
In the docs here, it states that `kube-prometheus-stack` is an optional dependency.  What does that mean for 'rook-ceph', is it hard?


Screenshot 2023-02-09 at 1 52 23 PM
Comparing to this section above, I see that in the matadata.yaml for kube-prometheus-stack it has traefik listed as a dependency
```dependencies:
  - traefik
  ```
Which correlates with the documentation. Maybe there is some distinction that can be used there. Just some thoughts. 

Totally fine if I need to close this out also.

vacas5:
Okay I should have asked for clarification on hard/ soft. From my perspective the UI is there to make the k8s experience easier. We should try and help users from getting into sticky situations where something doesn't work as expected. The CLI is always there for advanced use cases.

If the backend isn't using this field for anything else, I think then we should include soft dependencies here so we can help our users get the best experience that we would recommend. Left these thoughts in the epic too

takirala:
@nataliepina @vacas5 is this targeted for 2.5.0 ?

nataliepina:
> @nataliepina @vacas5 is this targeted for 2.5.0 ?
@takirala 
I believe that is the intent. We are still waiting for some final clarification on this, and if this work here should be closed or added. There is some conversation happening on this epic around this https://d2iq.atlassian.net/browse/D2IQ-87215, we need a final confirmation from Sonali there.

PR #988, opened 2023-02-14
fix: patch mutating/validating gatekeeper webhooks

cbuto:
This PR works around a recent change to the gatekeeper mutating and validation webhooks in `v3.11.0`. 

The release namespace is now being ignored in https://github.com/open-policy-agent/gatekeeper/blob/master/charts/gatekeeper/templates/gatekeeper-mutating-webhook-configuration-mutatingwebhookconfiguration.yaml#L29-L32. 

To work around this, we can patch the hardcoded matchExpressions that ignores the release namespace until we can move it to a different namespace (hopefully next release).

https://d2iq.atlassian.net/browse/D2IQ-95922

```release-note

```

**Checklist**

### DIFF
--- a/services/gatekeeper/3.11.0/release/release.yaml
+++ b/services/gatekeeper/3.11.0/release/release.yaml
@@ -29,6 +29,30 @@ spec:
     - kind: ConfigMap
       name: gatekeeper-overrides
       optional: true
+  postRenderers:
+    - kustomize:
+        patchesJson6902:
+          - target:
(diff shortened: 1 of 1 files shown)


cbuto:
kuttl tests are passing at https://github.com/mesosphere/kommander/pull/2999

takirala (approved):

cwyl02 (approved):

PR #1002, opened 2023-02-22
feat: Add controller-runtime Grafana dashboard

cbuto:
Adds a very basic dashboard to visualize some controller-runtime metrics grabbed from https://book.kubebuilder.io/plugins/grafana-v1-alpha.html. 

image

https://d2iq.atlassian.net/browse/D2IQ-94330

Open to discussing where to put this dashboard but this seems like a neutral place, let me know if there's a better spot!

```release-note

```

**Checklist**

### DIFF
--- a/services/kube-prometheus-stack/44.2.1/grafana-dashboards/controller-runtime.json
+++ b/services/kube-prometheus-stack/44.2.1/grafana-dashboards/controller-runtime.json
@@ -1,0 +1,710 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "Prometheus",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    }
+  ],
+  "__requires": [
+    {
+      "type": "datasource",
+      "id": "prometheus",
+      "name": "Prometheus",
+      "version": "1.0.0"
+    }
+  ],
+  "annotations": {
+    "list": [
+      {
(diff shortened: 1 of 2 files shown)


services/kube-prometheus-stack/44.2.1/grafana-dashboards/controller-runtime.json
```diff
@@ -0,0 +1,710 @@
...
+      "pluginName": "Prometheus"
+    }
+  ],
+  "__requires": [
+    {
+      "type": "datasource",
+      "id": "prometheus",
+      "name": "Prometheus",
+      "version": "1.0.0"
+    }
```
takirala, line 18: these differ from existing dashboards - https://github.com/mesosphere/kommander-applications/blob/cbuto/add-controller-runtime-dash/services/kube-prometheus-stack/44.2.1/grafana-dashboards/flux-cluster-stats.json
like there is no mention of `version: 1.0.0` and datasource is always referred to as `"datasource": "$datasource",`

I believe you might be just copy-pasting and dont know what needs to be done differently to get the right template though.

cbuto: ah yeah, let me drop those fields

cbuto: done in 4d94fd7d2048a1fbd6f85536410e779aa62493eb

sonofgibs (approved):

takirala (approved):

PR #1003, opened 2023-02-22
feat: automate image bump for licenses.d2iq.yaml

sonofgibs:
Automates the process of bumping images in licenses.d2iq.yaml by using the new 'update' flag for image mapping. 

https://d2iq.atlassian.net/browse/D2IQ-95738 

The last commit in this PR is and example run of an automated image bump. Two images were manually changed to outdated versions. Mergebot created a pr to remedy this.

As comments are not included in new versions of licenses.d2iq.yaml, there will be a follow-up pr which updates the documentation in this repo's README.

```release-note

```

**Checklist**

### DIFF
--- a/licenses.d2iq.yaml
+++ b/licenses.d2iq.yaml
@@ -210,7 +210,7 @@ resources:
       - url: https://github.com/knative-sandbox/net-istio
         ref: knative-v1.8.0
         license_path: LICENSE
-  - container_image: gcr.io/knative-releases/knative.dev/net-istio/cmd/webhook:v1.7.0
+  - container_image: gcr.io/knative-releases/knative.dev/net-istio/cmd/webhook:v1.8.0
     sources:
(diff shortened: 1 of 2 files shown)


cbuto (approved):

licenses.d2iq.yaml
```diff
@@ -1,27 +1,12 @@
 ignore:
   - docker.io/mesosphere/grafana-plugins:v0.0.1
-
-  # The image is set of tools that is not built from source code.
```
takirala, line 4: i remember you were talking about this - have we determined that having comments is not possible in the yaml generator ?

sonofgibs: @takirala hypothetically it could be possible but it was decided it'd be more effort than benefit given so decided to just add to README

cwyl02 (approved):

takirala (approved):

PR #1019, opened 2023-03-01
feat: Bump up alertmanager resource requests

gracedo:
This bumps it up to match what we were originally setting in the KPS overrides (for mgmt cluster). The diff in resources here is small and doesn't hurt to bump here to be the default. This allows us to also remove this override from the KPS overrides.

goes with https://github.com/mesosphere/kommander-cli/pull/986

https://d2iq.atlassian.net/browse/D2IQ-96187

```release-note

```

**Checklist**

### DIFF
--- a/services/kube-prometheus-stack/44.2.1/defaults/cm.yaml
+++ b/services/kube-prometheus-stack/44.2.1/defaults/cm.yaml
@@ -361,8 +361,8 @@ data:
             cpu: 200m
             memory: 250Mi
           requests:
-            cpu: 10m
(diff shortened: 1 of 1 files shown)


takirala (approved):

cbuto (approved):

gracedo:
I also updated https://d2iq.atlassian.net/wiki/spaces/DENT/pages/68124702/Workspace+Platform+Application+Defaults+and+Resource+Requirements table to account for these updates in KPS resources

PR #1065, opened 2023-03-16
feat: Add optional mgmt overrides cm to KPS HR

gracedo:
Always add optional CM for KPS HR for mgmt specific overrides that will be created in the kcli. This is so that we can allow users to use the workspace configOverrides for their own custom configuration. Right now, users would edit and existing `kube-prometheus-stack-overrides` CM with our mgmt override values, and because of that, we have not been able to automatically update that CM for existing versions if we had any updates, and also results in a bug during expansion, since the mgmt overrides are not compatible with an attached cluster setup.

This work is done in tandem with changes in the CLI to create a new `kube-prometheus-stack-mgmt-overrides` CM that also gets updated during upgrades. To assist with upgrades from previous versions, we will also implement an upgrader step to read in the existing `kube-prometheus-stack-overrides` CM and remove our mgmt override values from that CM so that the result is user custom config (or if no custom config was added, we can remove the config override altogether from the appdeployment).

https://d2iq.atlassian.net/browse/D2IQ-96172

```release-note

```

**Checklist**

### DIFF
--- a/services/kube-prometheus-stack/44.2.1/helmrelease/kube-prometheus-stack.yaml
+++ b/services/kube-prometheus-stack/44.2.1/helmrelease/kube-prometheus-stack.yaml
@@ -31,6 +31,9 @@ spec:
   valuesFrom:
     - kind: ConfigMap
       name: kube-prometheus-stack-44.2.1-d2iq-defaults
+    - kind: ConfigMap
+      name: kube-prometheus-stack-mgmt-overrides
+      optional: true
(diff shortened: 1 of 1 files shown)


cbuto (approved):

cwyl02 (approved):

sonofgibs (approved):

gracedo:
@d2iq-mergebot backport release-2.5

PR #1106, opened 2023-03-31
feat: update dependencies to match docs

clintonmedbery:
Closes [D2IQ-96389](https://d2iq.atlassian.net/browse/D2IQ-96389)

Ok back to the drawing board. So here is how this works. We have three types of dependencies. Please let me know if I am incorrect... 😅 

I made a spreadsheet to help:
https://docs.google.com/spreadsheets/d/1Z9nzVdcACXk3nZx7eEQw2Zf3tuZAXogUl-ATU4ifO9Q/edit#gid=0

And the docs:
https://beta-docs.d2iq.com/dkp/2.5/platform-application-dependencies
https://beta-docs.d2iq.com/dkp/2.5/project-platform-application-dependencies

**Hard dependencies**
You CANNOT deploy an app if these are deployed. These are under `requiredDependencies` in this repo. They are under required dependencies in docs. 

**Soft dependencies** 
You CAN deploy an app if these are deployed, but you must confirm. They are under `dependencies in this PR/repo. They are under required dependencies in the docs, which is a bit confusing.

**Optional dependencies**
There are no special considerations for these. They are under optional dependencies in the docs.

~Originally this was just meant to add soft dependencies, but apps seemed out of date so this attempts to bring kommander-app dependencies up to speed with docs. https://beta-docs.d2iq.com/dkp/2.5/platform-application-dependencies#PlatformApplicationDependencies-Backup~

https://d2iq.atlassian.net/browse/D2IQ-96389

**Checklist**

[D2IQ-96389]: https://d2iq.atlassian.net/browse/D2IQ-96389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/services/grafana-loki/metadata.yaml
+++ b/services/grafana-loki/metadata.yaml
@@ -2,6 +2,8 @@ displayName: Grafana Loki
 description: Horizontally-scalable, highly-available, multi-tenant log aggregation system inspired by Prometheus. Requires Logging Operator.
 category:
   - logging
+dependencies:
+  - rook-ceph-cluster
 type: platform
 scope:
   - workspace
--- a/services/kube-oidc-proxy/metadata.yaml
+++ b/services/kube-oidc-proxy/metadata.yaml
@@ -2,6 +2,9 @@ displayName: kube-oidc-proxy
 description: Proxy to consistently authenticate to managed Kubernetes clusters, on multi-cloud, using OIDC. Requires cert-manager and Traefik.
 category:
   - sso
+dependencies:
+  - cert-manager
+  - traefik
 type: internal
 scope:
   - workspace
--- a/services/kubernetes-dashboard/metadata.yaml
+++ b/services/kubernetes-dashboard/metadata.yaml
@@ -2,6 +2,8 @@ displayName: Kubernetes Dashboard
 description: General purpose, web-based UI for Kubernetes clusters. It allows users to manage applications running in a cluster and troubleshoot them, as well as manage the cluster itself.
 category:
   - monitoring
+dependencies:
+  - traefik
 type: platform
 scope:
   - workspace
--- a/services/project-grafana-loki/metadata.yaml
+++ b/services/project-grafana-loki/metadata.yaml
@@ -2,6 +2,10 @@ displayName: Grafana Loki
 description: Horizontally-scalable, highly-available, multi-tenant log aggregation system inspired by Prometheus. Requires Logging Operator.
 category:
   - logging
+dependencies:
+  - project-logging
+  - grafana-loki
+  - logging-operator
 type: platform
 scope:
   - project
--- a/services/velero/metadata.yaml
+++ b/services/velero/metadata.yaml
@@ -2,6 +2,8 @@ displayName: Velero
 description: Open source tool to safely backup and restore, perform disaster recovery, and migrate Kubernetes cluster resources and persistent volumes.
 category:
   - backup
+dependencies:
+  - rook-ceph-cluster
 type: platform
 scope:
   - workspace
--- a/services/grafana-logging/metadata.yaml
+++ b/services/grafana-logging/metadata.yaml
@@ -2,6 +2,8 @@ displayName: Grafana Logging
 description: Logging dashboard used to view logs aggregated to Grafana Loki. Requires Grafana Loki.
 category:
   - logging
+dependencies:
+  - grafana-loki
 type: platform
 scope:
   - workspace
--- a/services/project-grafana-logging/metadata.yaml
+++ b/services/project-grafana-logging/metadata.yaml
@@ -2,6 +2,8 @@ displayName: Grafana Logging
 description: Logging dashboard used to view logs aggregated to Grafana Loki. Requires Grafana Loki.
 category:
   - logging
+dependencies:
+  - project-grafana-loki
 type: platform
 scope:
   - project
--- a/services/project-logging/metadata.yaml
+++ b/services/project-logging/metadata.yaml
@@ -2,6 +2,8 @@ displayName: Project Logging
 description: Application by D2iQ that defines resources for the Logging Operator, which are used for directing project logs to its Loki application.
 category:
   - logging
+dependencies:
+  - logging-operator
 type: platform
 scope:
   - project
--- a/services/rook-ceph-cluster/metadata.yaml
+++ b/services/rook-ceph-cluster/metadata.yaml
@@ -3,6 +3,8 @@ description: Managed by Rook Ceph Operator, Ceph Cluster provides storage servic
 category:
   - logging
   - backup
+dependencies:
+  - rook-ceph
 type: platform
 scope:
   - workspace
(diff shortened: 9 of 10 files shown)


services/grafana-loki/metadata.yaml
```diff
@@ -2,6 +2,8 @@ displayName: Grafana Loki
 description: Horizontally-scalable, highly-available, multi-tenant log aggregation system inspired by Prometheus. Requires Logging Operator.
 category:
   - logging
+requiredDependencies:
+  - rook-ceph-cluster
```
cbuto, line 6: its possible to deploy Loki without rook-ceph, but it requires custom configuration to point Loki at another storage backend. will `requiredDependencies` allow that?

services/velero/metadata.yaml
```diff
@@ -2,6 +2,8 @@ displayName: Velero
 description: Open source tool to safely backup and restore, perform disaster recovery, and migrate Kubernetes cluster resources and persistent volumes.
 category:
   - backup
+dependencies:
+  - rook-ceph-cluster
```
cbuto, line 6: same as Loki, it's possible to deploy Velero without rook-ceph but it requires the user to configure Velero correctly. 

does `dependencies` mean we just show a warning?

clintonmedbery: Yep, the user will see a modal asking them to proceed.

services/grafana-loki/metadata.yaml
clintonmedbery: No it will not. I can update. But we might want to update docs as well. cc @gvasquezvargas

mstruebing (approved):

mstruebing:
Do you want to update these in our local UI environment as well?

clintonmedbery:
@mstruebing That's the plan!

services/grafana-loki/metadata.yaml
```diff
@@ -2,6 +2,8 @@ displayName: Grafana Loki
 description: Horizontally-scalable, highly-available, multi-tenant log aggregation system inspired by Prometheus. Requires Logging Operator.
 category:
   - logging
+dependencies:
+  - rook-ceph-cluster
```
gracedo, line 6: for the same reason as here, I think we also need to add the same dependencies block to `velero` metadata.yaml (soft dependency on `rook-ceph-cluster`)

clintonmedbery: Added. cc @gvasquezvargas

nataliepina (approved):
✔️

services/project-grafana-loki/metadata.yaml
```diff
@@ -2,6 +2,10 @@ displayName: Grafana Loki
 description: Horizontally-scalable, highly-available, multi-tenant log aggregation system inspired by Prometheus. Requires Logging Operator.
 category:
   - logging
+dependencies:
+  - project-logging
+  - grafana-loki
```
cbuto, line 7: I'm not sure if this requires `grafana-loki` 🤔 

* `grafana-loki` is the workspace deployment of Loki (which shouldn't affect this app). or maybe there was another reason this was here?
* `project-logging` configures the log flows/outputs for fluentd/fluentbit so logs are pushed to Loki... which also requires the `logging-operator`. so both of those look good!
* This app also uses `rook-ceph` and `rook-ceph-cluster`, but it could be configured to not use them.

services/grafana-loki/metadata.yaml
```diff
@@ -2,6 +2,8 @@ displayName: Grafana Loki
 description: Horizontally-scalable, highly-available, multi-tenant log aggregation system inspired by Prometheus. Requires Logging Operator.
 category:
   - logging
+dependencies:
```
cbuto, line 5: To collect and query logs out of the box, users need to install these:

* `logging-operator` - configures fluentd/fluentbit to push logs to Loki
* `grafana-logging` - grafana UI for querying Loki
*  `rook-ceph`/`rook-ceph-cluster` - for backend storage, but you can configure Loki to use another backend like s3

services/kube-oidc-proxy/metadata.yaml
```diff
@@ -2,6 +2,9 @@ displayName: kube-oidc-proxy
 description: Proxy to consistently authenticate to managed Kubernetes clusters, on multi-cloud, using OIDC. Requires cert-manager and Traefik.
 category:
   - sso
+dependencies:
```
cbuto, line 5: btw, we don't show this app in the UI but it won't hurt to have these here!

services/kubernetes-dashboard/metadata.yaml
```diff
@@ -2,6 +2,8 @@ displayName: Kubernetes Dashboard
 description: General purpose, web-based UI for Kubernetes clusters. It allows users to manage applications running in a cluster and troubleshoot them, as well as manage the cluster itself.
 category:
   - monitoring
+dependencies:
+  - traefik
```
cbuto, line 6: In the UI, we already require all "foundational" apps to be deployed. Traefik is one of the foundational apps. 

would the foundational app logic already handle telling users about this dependency?

services/project-grafana-loki/metadata.yaml
gracedo: We "require" users to deploy the ws level logging stack as a prereq to deploying the project logging stack, which is probably why those are defined here https://docs.d2iq.com/dkp/2.4/multi-tenant-logging-overview

services/grafana-loki/metadata.yaml
gracedo: How do we want to "chain" dependencies? e.g. `rook-ceph-cluster` is defined here but not `rook-ceph` altho both are technically required. But `rook-ceph` is defined as a required dep on `rook-ceph-cluster` -- so does that mean we want to define just the "first" level of deps (so, just `rook-ceph-cluster), and not the entire chain (e.g. `rook-ceph` and `rook-ceph-cluster`)? Not sure which is more confusing to users.

services/project-grafana-loki/metadata.yaml
cbuto: ah gotcha, in that case we can keep `grafana-loki` here

cbuto (approved):

services/grafana-loki/metadata.yaml
gracedo: ☝🏻 just this last remaining question. We have a few places where both `rook-ceph` and `rook-ceph-cluster` are defined and others where just `rook-ceph-cluster` is defined, so just want to be consistent

clintonmedbery: I removed the chaining after discussing it with the UI team. It kind of confuses the process and if we did want to make the user experience better we can still do that programmatically through the UI. Sorry for the back and forth.

gracedo (approved):

PR #1122, opened 2023-04-03
feat: add kubetunnel-proxy ingress config

Azhovan:

 https://d2iq.atlassian.net/browse/D2IQ-93666

```release-note

```

**Checklist**

### DIFF
--- a/services/kubetunnel/0.0.15/defaults/cm.yaml
+++ b/services/kubetunnel/0.0.15/defaults/cm.yaml
@@ -16,3 +16,17 @@ data:
       kubectlImage:
         repository: bitnami/kubectl
         tag: 1.24.6
+    deploymentAnnotations:
+      secret.reloader.stakater.com/reload: "kubetunnel-reverseproxy-tls"
+    ingress:
+      enabled: true
+      annotations:
+        kubernetes.io/ingress.class: kommander-traefik
+        traefik.ingress.kubernetes.io/router.tls: "true"
+      path: /dkp/kubetunnel-proxy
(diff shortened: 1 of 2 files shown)


services/kubetunnel/0.0.15/defaults/cm.yaml
```diff
@@ -16,3 +16,13 @@ data:
...
         repository: bitnami/kubectl
         tag: 1.24.6
+    deploymentAnnotations:
+      secret.reloader.stakater.com/reload: "kubetunnel-reverseproxy-tls"
+    ingress:
+      enabled: true
+      annotations:
+        kubernetes.io/ingress.class: kommander-traefik
+        traefik.ingress.kubernetes.io/router.tls: "true"
+      path: /dkp/kubetunnel-proxy
```
Azhovan, line 26: this looks not correct to me

services/kubetunnel/0.0.15/defaults/cm.yaml
```diff
@@ -16,3 +16,17 @@ data:
       kubectlImage:
         repository: bitnami/kubectl
         tag: 1.24.6
+    deploymentAnnotations:
+      secret.reloader.stakater.com/reload: "kubetunnel-reverseproxy-tls"
+    ingress:
```
palexster, line 21: Can we enforce the mTLS setup in this configuration to prevent other components from using the reverse-proxy without passing through the authenticated ingress?

palexster (approved):

mhrabovcin (approved):

PR #1123, opened 2023-04-03
feat: Introduce kommander-ui service

mikolajb:
This change creates a separate service for kommander-ui.

https://d2iq.atlassian.net/browse/D2IQ-92244

```release-note
Kommander-ui is a separate service now and it is deployed thourhg appdeployment like other services.
```

**Checklist**

### DIFF
--- a/services/kommander-ui/metadata.yaml
+++ b/services/kommander-ui/metadata.yaml
@@ -1,0 +1,5 @@
+type: internal
+scope:
+  - workspace
+dependencies:
+  - kommander
--- a/services/kommander-ui/metadata.yaml
+++ b/services/kommander-ui/metadata.yaml
@@ -1,5 +1,5 @@
 type: internal
 scope:
   - workspace
-dependencies:
+requiredDependencies:
   - kommander
--- a/services/kommander/0.6.0/defaults/cm.yaml
+++ b/services/kommander/0.6.0/defaults/cm.yaml
@@ -76,14 +76,7 @@ data:
         prometheus-thanos-traefik: "0.0.1"
         thanos: "12.4.3"
     kommander-ui:
-      ingress:
-        enabled: true
-        extraAnnotations:
-          kubernetes.io/ingress.class: kommander-traefik
-          traefik.ingress.kubernetes.io/router.tls: "true"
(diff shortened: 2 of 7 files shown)


takirala (approved):

jimmidyson:
Will this introduce any downtime to kommander ui on upgrade?

services/kommander/0.6.0/defaults/cm.yaml
```diff
@@ -74,14 +74,7 @@ data:
...
     kommander-ui:
-      ingress:
-        enabled: true
-        extraAnnotations:
-          kubernetes.io/ingress.class: kommander-traefik
-          traefik.ingress.kubernetes.io/router.tls: "true"
-          traefik.ingress.kubernetes.io/router.middlewares: "${workspaceNamespace}-stripprefixes@kubernetescrd,${workspaceNamespace}-forwardauth@kubernetescrd"
-        path: /dkp/kommander/dashboard
-        graphqlPath: /dkp/kommander/dashboard/graphql
+      enabled: false
```
gracedo, line 77: maybe one more pr in kommander to remove kommander-ui as a subchart to go along with this and https://github.com/mesosphere/kommander-cli/pull/1035 ?

services/kommander-ui/metadata.yaml
```diff
@@ -0,0 +1,5 @@
+type: internal
+scope:
+  - workspace
+dependencies:
```
gracedo, line 4:
```suggestion
requiredDependencies:
```

I think it needs to be `requiredDependencies` bc there is a HR dependency here which will always block deployment

services/kommander/0.6.0/defaults/cm.yaml
mikolajb: https://github.com/mesosphere/kommander/pull/3315

mikolajb:
> Will this introduce any downtime to kommander ui on upgrade?

@jimmidyson From what I tested so far, we can expect a downtime of up to 2 minutes. We could probably reduce it by installing/upgrading kommander and kommander-ui apps at the same time. How does it sound?

jimmidyson:
@mikolajb As long as product know and are OK with it, then all good :)

gracedo (approved):

PR #1130, opened 2023-04-06
feat: Add dkp priority class manifests to common/base

gracedo:
Add new DKP priority class manifests directly to the gitrepo under `/common/base` since these manifests should always be deployed to both the mgmt cluster and the attached clusters. These will automatically get applied during upgrade first (with gitrepo update), before other apps are upgraded, which is required since these upgraded apps will be updated to use the priority classes.

This works for attached cluster upgrades as well, because the kommander flux controller will be upgraded already with kommander upgrade, and once the upgraded flux controller is running, it will automatically apply the new "federated global manifests" (which include the kustomization pointing to `./common/priority-claasses/` -- added in https://github.com/mesosphere/kommander/pull/3281) before workspace upgrade is performed.

https://d2iq.atlassian.net/browse/D2IQ-96266

```release-note

```

**Checklist**

### DIFF
--- a/common/priority-classes/dkp-priority-classes.yaml
+++ b/common/priority-classes/dkp-priority-classes.yaml
@@ -1,0 +1,23 @@
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: dkp-critical-priority
+value: 100002000
+globalDefault: false
+description: "This is the highest priority class that is used for critical priority DKP workloads."
+---
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: dkp-high-priority
+value: 100001000
+globalDefault: false
+description: "This is the priority class that is used for high priority DKP workloads."
+---
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: dkp-default-priority
+value: 100000000
+globalDefault: true
+description: "This is the global default priority class that is automatically applied to pods without a priority class."
--- a/common/priority-classes/dkp-priority-classes.yaml
+++ b/common/priority-classes/dkp-priority-classes.yaml
@@ -13,11 +13,3 @@ metadata:
 value: 100001000
 globalDefault: false
 description: "This is the priority class that is used for high priority DKP workloads."
----
-apiVersion: scheduling.k8s.io/v1
-kind: PriorityClass
-metadata:
-  name: dkp-default-priority
-value: 100000000
-globalDefault: true
-description: "This is the global default priority class that is automatically applied to pods without a priority class."
--- a/common/priority-classes/kustomization.yaml
+++ b/common/priority-classes/kustomization.yaml
@@ -1,6 +1,4 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - ../helm-repositories/
-  - ../priority-classes/
-  - ../values/
+  - dkp-priority-classes.yaml


common/priority-classes/dkp-priority-classes.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: dkp-high-priority
+value: 100001000
+globalDefault: false
+description: "This is the priority class that is used for high priority DKP workloads."
+---
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: dkp-default-priority
```
sonofgibs, line 20: Would we want to name this default-priority since it'll likely be applied to client specific pods?

sonofgibs (approved):

common/priority-classes/dkp-priority-classes.yaml
```diff
@@ -0,0 +1,23 @@
...
+value: 100000000
+globalDefault: true
```
cbuto, line 22: I wonder if there will ever a need to disable this default priority class 🤔

gracedo: Hmm, yeah, I was going to ask if we _needed_ this priority class which is set as the global default. Since it is applied to ALL pods without a priority class, just wondering why we would want to force this vs just letting it default to 0 (without priority class basically.. I feel like a default is only relevant if a user has created other priority classes against the number that is set as the default - so it seems like it would be more user-defined). And yeah I would be concerned about a user who might have set their own globalDefault class, or wanted to add their own, or adjust this one. It does seem like this type of default class increases the chances that a user would want to customize it.

So I guess, two questions:
1. Do we want to create a third "default" dkp priority class?
2. If yes, should it be set as the global default, knowing that users can't adjust this? (without writing to git, which, if we are going to be suggesting that, then I would rather not set it as the global default to begin with) If we don't set it as the global default, we can still create a third class but maybe name it something other than "default"?

cc @sonofgibs

common/priority-classes/dkp-priority-classes.yaml
gracedo: I'd prefer to label the priority classes that we're creating in such a way that it's obvious they're DKP created classes, but I had a question about whether we want to create this global default class to begin with: https://github.com/mesosphere/kommander-applications/pull/1130#discussion_r1160814770

common/priority-classes/dkp-priority-classes.yaml
sonofgibs: I would lean more towards having a default class because users may have a specific default value they use that would be the same as dkp-high or dkp-critical. I agree though that users will likely want the ability to customize the default class.

karishmagadiya: If we are not defining this default class does it mean that all customers Pods would be categorized under "system-cluster-critical" priority class which would be lower than dkp-critical-priority and  dkp-high-priority unless customers have customized default class ?

cbuto: based on the standup discussion and what @gracedo mentioned, it seems like we definitely don't want to create a global default. if we don't set the third priority class as a default, then maybe we don't need it anymore 🤔

gracedo: Without a priority class that is set as the global default, all pods without a priority class (e.g. customer pods) would have a priority `0`. The two priority classes that are shipped by default with k8s (`system-cluster-critical` and `system-node-critical`) are set higher than our dkp priority classes. Pods without a priority class won't be using these as the default.

From https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass:
> PriorityClass also has two optional fields: globalDefault and description. The globalDefault field indicates that the value of this PriorityClass should be used for Pods without a priorityClassName. Only one PriorityClass with globalDefault set to true can exist in the system. If there is no PriorityClass with globalDefault set, the priority of Pods with no priorityClassName is zero.

In general, I don't think we should create a priority class that is set as the global default. But since we've created `dkp-critical-priority` and `dkp-high-priority`, maybe we can provide an additional `dkp-low-priority` (that we won't be using at this moment but maybe in the future we might?). Users can use any of these priority classes to schedule their own pods if they'd like, or to override the priorities on dkp workloads that we set by default. Having this third "low" class seems like it might make the story a bit more complete and add some more meaning to the different levels provided (critical, high, low ... do we also want a medium 😅 )

takirala: (nit) if we are not using something then probably its worth not shipping it.. techincially not setting a priority class on a pod means it uses a value of 0 so we are already shipping a global default (the `0` is what we are "shipping"). its pretty easy to create a priority class anyway so i am leaning towards shipping only what we need. its up to the user if they want to create another priority class (and name it `dkp-low-priority` or whatever they want).

On a different tangent - If we want to ship a global default we can use cli to dynamically inject a patch to gauge if there is a global default priority class and if there is already a global default then it can add a patch to use `globalDefault: false` into the git repository during install/upgrade time. But if we dont want to ship a global default then i think we just ship what we use and let the user create what they want.

No strong taste though.

common/priority-classes/dkp-priority-classes.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: dkp-critical-priority
+value: 100002000
+globalDefault: false
```
takirala, line 6:
```suggestion
```

common/priority-classes/dkp-priority-classes.yaml
```diff
@@ -0,0 +1,23 @@
...
+description: "This is the highest priority class that is used for critical priority DKP workloads."
+---
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: dkp-high-priority
+value: 100001000
+globalDefault: false
```
takirala, line 14:
```suggestion
```

takirala (commented):
> If you delete a PriorityClass, existing Pods that use the name of the deleted PriorityClass remain unchanged, but you cannot create more Pods that use the name of the deleted PriorityClass.

https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption

`but you cannot create more Pods that use the name of the deleted PriorityClass.` sounds fair but something we need to pay attention to (like we cant change these names again, we have to make sure any future restructures to `common/base` do not cause deletion/recreation etc.,). nothing to be alarmed about but just making sure we are aware of this.

common/priority-classes/dkp-priority-classes.yaml
gracedo: Hard to predict how users will be using priority classes, but it just feels like better UX (admittedly, probably overkill at the moment tho) to have low, medium, high and critical priority classes, it feels more like a complete story. That way users can look at our docs to see what each app's priority is set to (currently, all highs and criticals) and then schedule their pods accordingly using our predefined classes without having to look up the actual priority class number and then creating their own classes with a number above or below. It seems easier to just use our preset classes and they just need to know which classes are higher or lower vs keeping track of numbers.

But, no idea if this is an actual problem that will happen in practice, so maybe it's just overengineering to provide two extra classes that we aren't actually using right now. We can perhaps add them later on a need basis if we end up adding apps that we think should be lower than high, or if we receive user feedback that it would be nice to provide some preset classes for them to use.

I'm going to remove this third class for now, we can continue the discussion and always add more classes back. For now will just leave the two classes we absolutely need and will be using in our apps

takirala:
https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#non-preempting-priority-class

> preemptionPolicy defaults to PreemptLowerPriority, which will allow pods of that PriorityClass to preempt lower-priority pods (as is existing default behavior). If preemptionPolicy is set to Never, pods in that PriorityClass will be non-preempting.

We can mention this in our documentation if we want to talk about creating a "global default"

cwyl02 (approved):

PR #1136, opened 2023-04-12
feat: Enable etcd metrics in Prometheus

alembiewski:
Deploys `etcd-metrics-proxy` component based on [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) as a part of `kube-promethues-stack` to enable etcd metrics scraping by Prometheus.
More details are in the [doc](https://docs.google.com/document/d/1W0AZ6fGMMtFWsjEFO4KibolCM9CfXAl27_qyAGNIvTQ/edit?usp=sharing).

https://d2iq.atlassian.net/browse/D2IQ-96642

- The changes to the `additionalServiceMonitors` will be copied to https://github.com/mesosphere/kommander-cli/blob/main/pkg/installer/config/manifests/kube-prometheus-stack/overrides.yaml after this PR is approved.
- The proposed solution won't work in EKS/AKS clusters, where control plane nodes are managed by AWS directly and hidden from the end users. The etcd-metrics-proxy DaemonSet will be a no-op and won't break Prometheus in such a scenario:
```
[11/04/23 5:12:24] ➜  kube-rbac-proxy git:(alex/check-etcd-access) ✗ k get daemonset etcd-metrics-proxy
  NAME                 DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR                            AGE
  etcd-metrics-proxy   0         0         0       0            0           node-role.kubernetes.io/control-plane=   26m
```
 
```release-note
Enable etcd metrics in Prometheus
```

**Checklist**


services/kube-prometheus-stack/44.2.1/defaults/cm.yaml
```diff
@@ -450,15 +468,13 @@ data:
...
-    #   serviceMonitor:
-    #     scheme: "https"
-    #     caFile: "/etc/prometheus/secrets/etcd-certs/ca.crt"
-    #     certFile: "/etc/prometheus/secrets/etcd-certs/server.crt"
-    #     keyFile: "/etc/prometheus/secrets/etcd-certs/server.key"
+      # enable grafana dashboard for etcd
+      enabled: true
+      # kubeEtcd Service and ServiceMonitor are disabled, etcd metrics are handled by 'etcd-metrics-proxy' component
+      service:
+        enabled: false
```
alembiewski, line 475: There was [a discussion recently](https://d2iq.slack.com/archives/CR8GEL5TL/p1681286317515369) about this service - does it make sense to have to enable that in `kubeEtcd` given that is used by customers? After giving it some thought, I think we probably shouldn't do that for security reasons.

gracedo (approved):
For the kommander-cli changes, I can take care of that PR when this merges. There's a bit more work to do to make sure this config change only applies to v2.6+

alembiewski:
Thanks for taking care of kommander-cli changes, @gracedo! I [moved](https://github.com/mesosphere/kommander-applications/pull/1136/commits/728c486c696949499fa3d46aad9e58618afaefe7) the ServiceMonitor to the `additionalServiceMonitors` because there was a race-condition issue related to the missing `ServiceMonitor` CRD (I guess this is the reason why we have other ServiceMonitors defined there, so this is not new).

gracedo:
✅ Tests are passing on https://github.com/mesosphere/kommander/pull/3307

lukeogg (approved):

PR #1165, opened 2023-04-18
fix: Remove unnecessary traefik dependency from apps

gracedo:
There were a few apps that had traefik set as a HR dependency, when they are technically not dependent on it (they successfully deploy if traefik is not deployed).

```release-note

```

**Checklist**

### DIFF
--- a/services/centralized-kubecost/0.33.1/release/release.yaml
+++ b/services/centralized-kubecost/0.33.1/release/release.yaml
@@ -4,9 +4,6 @@ metadata:
   name: centralized-kubecost
   namespace: ${releaseNamespace}
 spec:
-  dependsOn:
-    - namespace: ${releaseNamespace}
-      name: traefik
   chart:
     spec:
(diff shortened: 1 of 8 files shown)


services/centralized-kubecost/0.33.1/release/release.yaml
```diff
@@ -4,9 +4,6 @@ metadata:
   name: centralized-kubecost
   namespace: ${releaseNamespace}
 spec:
-  dependsOn:
```
gracedo, line 7: @gvasquezvargas we should double check that the docs are aligned with these changes (traefik shouldn't be listed as a dep for centralized-kubecost, dkp-insights-management, kube-prometheus-stack, and kubecost)

lukeogg (approved):
LGTM!

cwyl02 (approved):

takirala (approved):

PR #1172, opened 2023-04-19
feat: switch to bitnami/thanos

cwyl02:
switch to bitnami/thanos

https://d2iq.atlassian.net/browse/D2IQ-96780

```release-note

```

**Checklist**


services/thanos/12.4.3/defaults/cm.yaml
```diff
@@ -0,0 +1,64 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: thanos-12.4.3-d2iq-defaults
+  namespace: ${releaseNamespace}
+data:
+  values.yaml: |
+    ---
+    image:
```
cwyl02, line 11: The default here is [using docker hub](https://github.com/bitnami/charts/blob/main/bitnami/thanos/values.yaml#L54-L57).
But since I am overriding the image does this mean I also need to change the chart bump tool? @gracedo @bcmendoza ?

gracedo: hmm. can we use the default image instead of overriding?

cwyl02: we can. just 1 more image competes for dockerhub limit. but guess thats ok since customer can always override themselves if they are really bothered by this.

gracedo (approved):

lukeogg (approved):

PR #1183, opened 2023-04-27
ci: Update e2e upgrade from version to v2.5.0-dev

gracedo:
we missed some bumps after release

```release-note

```

**Checklist**


gracedo:
FINE, https://github.com/mesosphere/kommander-applications/pull/1182 wins 😂

PR #1212, opened 2023-05-01
feat: dkp critical priorityclass - velero&kommander

cwyl02:
dkp critical for velero, kommander

https://d2iq.atlassian.net/browse/D2IQ-96267

```release-note

```

**Checklist**

### DIFF
--- a/services/kommander-appmanagement/0.6.0/defaults/cm.yaml
+++ b/services/kommander-appmanagement/0.6.0/defaults/cm.yaml
@@ -6,6 +6,7 @@ metadata:
   namespace: ${releaseNamespace}
 data:
   values.yaml: |
+    priorityClassName: "dkp-critical-priority"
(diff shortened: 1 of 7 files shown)


sonofgibs (approved):

gracedo (commented):
I think we need to block this until https://d2iq.atlassian.net/browse/D2IQ-96868 is done, or else flux will fail to deploy due to missing priority class. BUT, we actually have a separate task for bumping flux https://d2iq.atlassian.net/browse/D2IQ-96282 because it's a little more complicated bc I think it need to be coordinated with the flux manifests in kommander repo? so maybe, can we remove flux from this pr and handle that one in a separate PR and block it until https://d2iq.atlassian.net/browse/D2IQ-96868 is done?

gracedo (approved):

PR #1224, opened 2023-05-04
feat: dkp critical for gitea&chartmuseum

cwyl02:
apply dkp critical priority class for gitea and chart museum

https://d2iq.atlassian.net/browse/D2IQ-96267

```release-note

```

**Checklist**

### DIFF
--- a/services/gitea/8.2.0/defaults/cm.yaml
+++ b/services/gitea/8.2.0/defaults/cm.yaml
@@ -6,6 +6,7 @@ metadata:
 data:
   values.yaml: |-
     ---
+    priorityClassName: "dkp-critical-priority"
     image:
       rootless: true
     ingress:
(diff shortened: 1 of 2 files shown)


services/gitea/8.2.0/defaults/cm.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
 data:
   values.yaml: |-
     ---
+    priorityClassName: "dkp-critical-priority"
```
cbuto, line 9: for gitea, we'll need to apply the priority class to postgres/memcached as well

```
priorityClassName: <priority class>

memcached:
  priorityClassName: <priority class>

postgresql:
  primary:
    priorityClassName: <priority class>
```

cwyl02: great catch!! done

cbuto (approved):

sonofgibs (approved):

PR #1226, opened 2023-05-04
feat: dkp high for Kubecost, Centralized Kubecost, Kubernetes Dashboard

sonofgibs:
Update app configs to use DKP High PriorityClass for monitoring apps

```release-note

```

**Checklist**

### DIFF
--- a/hack/validate-default-apps-versions.sh
+++ b/hack/validate-default-apps-versions.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-latestKommanderVersion=$(gfind "services/kommander" -maxdepth 1 -type d -regextype sed -regex '.*/[0-9]\+.[0-9]\+.[0-9]\+' -printf "%f\n" | sort --version-sort | tail -1)
+latestKommanderVersion=$(find "services/kommander" -maxdepth 1 -type d -regextype sed -regex '.*/[0-9]\+.[0-9]\+.[0-9]\+' -printf "%f\n" | sort --version-sort | tail -1)
 
 validate_apps_in_yaml_path() {
   readarray apps < <(yq -e '.data["values.yaml"]' "services/kommander/$latestKommanderVersion/defaults/cm.yaml" | yq -e "$1")
--- a/hack/validate-default-apps-versions.sh
+++ b/hack/validate-default-apps-versions.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-latestKommanderVersion=$(find "services/kommander" -maxdepth 1 -type d -regextype sed -regex '.*/[0-9]\+.[0-9]\+.[0-9]\+' -printf "%f\n" | sort --version-sort | tail -1)
+latestKommanderVersion=$(gfind "services/kommander" -maxdepth 1 -type d -regextype sed -regex '.*/[0-9]\+.[0-9]\+.[0-9]\+' -printf "%f\n" | sort --version-sort | tail -1)
 
 validate_apps_in_yaml_path() {
   readarray apps < <(yq -e '.data["values.yaml"]' "services/kommander/$latestKommanderVersion/defaults/cm.yaml" | yq -e "$1")
--- a/services/centralized-kubecost/0.34.0/defaults/cm.yaml
+++ b/services/centralized-kubecost/0.34.0/defaults/cm.yaml
@@ -6,7 +6,9 @@ metadata:
 data:
   values.yaml: |
     ---
-    priorityClassName: "dkp-high-priority"
+    priority:
+      enabled: true
+      name: dkp-high-priority
     cost-analyzer:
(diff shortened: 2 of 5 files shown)


services/centralized-kubecost/0.34.0/defaults/cm.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
 data:
   values.yaml: |
     ---
+    priorityClassName: "dkp-high-priority"
```
gracedo, line 9: (applies to the below kubecost values as well) we use an umbrella chart that brings in upstream kubecost ("cost-analyzer" helm chart) as a subchart -- so all the upstream values are nested under `cost-analyzer` (line 10 in this diff here).

The upstream helm chart values are here: https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/values.yaml there are currently 2 places where priorityClassName can be set, but I am seeing now that these are both components that are not enabled by default. I'm going to actually look at contributing a PR upstream to add support to set priority class on all the deployments. Let's hold off on the 2 kubecost apps here until I get that in!

services/nvidia-gpu-operator/22.9.1/defaults/cm.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   namespace: ${releaseNamespace}
 data:
   values.yaml: |-
+    priorityClassName: "dkp-high-priority"
```
gracedo, line 8: here's the helm chart values: https://github.com/NVIDIA/gpu-operator/blob/master/deployments/gpu-operator/values.yaml I see that they are already defaulting to using `system-node-critical`, which I think is probably the right thing to do here (I don't think we should override them) so let's actually not make any changes here! I'll make a note in the doc about this

```suggestion
```

gracedo (commented):
Let's remove the override for nvidia, and also the kubecost apps for now (until I get an upstream PR in)

services/centralized-kubecost/0.34.0/defaults/cm.yaml
gracedo: https://github.com/kubecost/cost-analyzer-helm-chart/blob/1723242ea6744b2d20cdda6dea9a4a59ffe22faa/cost-analyzer/values.yaml#L487-L490

Took a closer look at kubecost as I went in to make the changes -- they do actually support creating a priority class and using it in some (but not all) deployments. We can provide our own priority class name and it'll get used:

```
priority:
  enabled: true
  name: dkp-high-priority
```

I will still look into seeing whether I should open an upstream pr to allow setting it for all deployments not just the main kubecost ones but we can set this field now without waiting! This should be set in both kubecost and centralized-kubecost

services/kubecost/0.34.0/defaults/cm.yaml
```diff
@@ -17,6 +17,7 @@ data:
           enabled: true
         grafana:
           enabled: true
+        priorityClassName: "dkp-high-priority"
```
gracedo, line 20:
```suggestion
```

kubecost pulls in prometheus and grafana subcharts, we'll want to set priority classes for those deployments as well. For prometheus, [server](https://github.com/kubecost/cost-analyzer-helm-chart/blob/1723242ea6744b2d20cdda6dea9a4a59ffe22faa/cost-analyzer/charts/prometheus/values.yaml#L551) and [alertmanager](https://github.com/kubecost/cost-analyzer-helm-chart/blob/1723242ea6744b2d20cdda6dea9a4a59ffe22faa/cost-analyzer/charts/prometheus/values.yaml#L51). Doesn't look like they support setting priority classes for grafana yet, I'll look at adding it there too.

services/centralized-kubecost/0.34.0/defaults/cm.yaml
sonofgibs: Good catch! Updates have been made

hack/validate-default-apps-versions.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-latestKommanderVersion=$(find "services/kommander" -maxdepth 1 -type d -regextype sed -regex '.*/[0-9]\+.[0-9]\+.[0-9]\+' -printf "%f\n" | sort --version-sort | tail -1)
+latestKommanderVersion=$(gfind "services/kommander" -maxdepth 1 -type d -regextype sed -regex '.*/[0-9]\+.[0-9]\+.[0-9]\+' -printf "%f\n" | sort --version-sort | tail -1)
```
cbuto, line 4: the image used in CI won't have `gfind` unfortunately, which is why pre-commit is failing. 

you can get pre-commit working locally with:
1. `brew install findutils`
2. then add this in your `.zshrc` file:
```
PATH=$(brew --prefix)/opt/findutils/libexec/gnubin:$PATH
```

then source it in your shell and it should be good to go!

gracedo:
closing, these are being taken care of in different PRs!

PR #1228, opened 2023-05-05
feat: dkp high for dka reloader

sonofgibs:
Update app configs to use DKP High PriorityClass for foundational apps

```release-note

```

**Checklist**

### DIFF
--- a/services/reloader/1.0.24/defaults/cm.yaml
+++ b/services/reloader/1.0.24/defaults/cm.yaml
@@ -7,6 +7,7 @@ data:
   values.yaml: |
     ---
     reloader:
+      priorityClassName: "dkp-high-priority"
       reloadOnCreate: true
       logFormat: json
       deployment:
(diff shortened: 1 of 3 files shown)


services/reloader/1.0.24/defaults/cm.yaml
```diff
@@ -7,6 +7,7 @@ data:
   values.yaml: |
     ---
     reloader:
+      priorityClassName: "dkp-high-priority"
```
cbuto, line 10: looking at the upstream chart, it looks like this belongs under the `deployment` key

https://github.com/stakater/Reloader/blob/master/deployments/kubernetes/chart/reloader/templates/deployment.yaml#L71-L73

gracedo: @sonofgibs on this ticket https://d2iq.atlassian.net/browse/D2IQ-96018 I linked most of the apps with their upstream helm chart values, it might be helpful to reference for each app you are overriding here to check where to be overriding the `priorityClassName` field! There may also be more than one place

cbuto (approved):
🚀

cwyl02 (approved):

PR #1241, opened 2023-05-11
feat: add high priority for kiali

sonofgibs:
Update app configs to use DKP High PriorityClass for kiali

```release-note

```

**Checklist**


services/jaeger/2.43.0/defaults/cm.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
 data:
   values.yaml: |
     jaeger:
+      priorityClassName: dkp-high-priority
```
cbuto, line 9: I couldn't find any info on this, does the Jaeger custom resource support this?

it looks like there's also `priorityClassName` that we can set on the operator:

https://github.com/jaegertracing/helm-charts/blob/main/charts/jaeger-operator/values.yaml#L96

services/kiali/1.67.0/defaults/cm.yaml
```diff
@@ -33,6 +33,7 @@ data:
             in_cluster_url: http://jaeger-jaeger-operator-jaeger-query.istio-system.svc.cluster.local:16685
             use_grpc: true
         deployment:
+          priorityClassName: dkp-high-priority
```
cbuto, line 36: based on the kiali CR (https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.deployment.priority_class_name), it looks like this should be:
```suggestion
          priority_class_name: dkp-high-priority
```


we'll also want to set this on the operator itself - https://github.com/kiali/helm-charts/blob/master/kiali-operator/templates/deployment.yaml#LL35C8-L35C8

gracedo (commented):
I think we're just missing setting `priorityClassName` at the top-level! https://github.com/kiali/helm-charts/blob/5f69912b558276cffb8f8b36471b3a0182330399/kiali-operator/values.yaml#L23

gracedo:
After pushing the last commit to set priority class on kiali operator deployment, and testing this on the daily:
```
k get deployments -A -o=jsonpath='{range .items[*]}{@.metadata.name}{"\t\t"}{@.spec.template.spec.priorityClassName}{" \n"}'  | grep kiali
kiali-kiali-operator		dkp-high-priority
kiali		dkp-high-priority
```

PR #1247, opened 2023-05-12
fix: Use array instead of map for apps values

gracedo:
Right now, since `kommander-licensing.defaultEnterpriseApps ` and `attached.prerequisites.defaultApps` take in a map, overrides do not actually override the entire map, just the overlapping keys. So, a user who overrides the values hoping to "remove" apps does not actually result in those apps being removed from the list of default apps.

To fix this, we change these helm chart fields to take in a list of app names instead of a map of app name to app version. kommander controllers are updated to use the latest version of the app to deploy.

https://d2iq.atlassian.net/browse/D2IQ-97199

https://github.com/mesosphere/kommander/pull/3411 and this PR unfortunately depend on each other. I will plan to merge https://github.com/mesosphere/kommander/pull/3411 first after CI is green there, without reverting the k-apps ref to avoid breaking kommander CI in the interim. Then, I will rerun tests here off the new version of kommander charts, along with another kommander PR pointing to this branch. Once everything is green, I will merge this PR and revert the k-apps ref back to `main` on kommander.

```release-note

```

**Checklist**


cbuto (approved):

karishmagadiya (approved):

gracedo:
`kommander-applications/install-e2e-tests` is failing with `e2e_suite_test.go:39: failed to load runtime config: failed get application versions for "github.com/mesosphere/kommander-applications.git?ref=685f07dc2ff03b31f9e9303f358cea4969e5bf17": *helm.Parser: template: kommander/templates/flux-operator/deployment_flux-operator.yaml:58:15: executing "kommander/templates/flux-operator/deployment_flux-operator.yaml" at <has "gatekeeper" .Values.attached.prerequisites.defaultApps>: error calling has: Cannot find has on type map` but I'm not sure why because this PR updates the fields from using a map to a key, and the defaults have also been updated on the kommander helm chart. I'm going to admin merge this now to avoid breaking the daily cluster tomorrow, but this install e2e test will likely continue to fail on k-apps PRs until we figure out how to fix it.

PR #1254, opened 2023-05-15
refactor: Use system-cluster-critical priority for rook-ceph

gracedo:
We decided in https://d2iq.slack.com/archives/C04SS4T0EN9/p1684189296640239 to set rook-ceph operator priority to `system-cluster-critical` to match the priority class of rook-ceph-cluster, which depends on the operator.

```release-note

```

**Checklist**


cbuto (approved):
💯

takirala (approved):

PR #1246, opened 2023-05-12
fix: wait for webhook to be deleted

cbuto:
**What problem does this PR solve?**:
Let's explicitly wait for the webhook config to be deleted. It turns out the operator manages the webhook config and not the helm chart.

**Which issue(s) does this PR fix?**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Checklist**

- [ ] If the PR adds a version bump, ensure there is no breaking change in Licensing model (or NA).
- [ ] If a chart is changed or app configuration is significantly changed, the chart version is correctly incremented (so that apps are not automatically upgraded from a previous version of DKP).

PR #1257, opened 2023-05-16
fix: ensure pre-upgrade runs prior to upgrade

cbuto:
backports https://github.com/mesosphere/kommander-applications/pull/1246 minus the waiting for the webook since it will already be disabled in 2.5

https://d2iq.atlassian.net/browse/D2IQ-97225

```release-note

```

**Checklist**

### DIFF
--- a/services/rook-ceph-cluster/1.11.4/pre-install.yaml
+++ b/services/rook-ceph-cluster/1.11.4/pre-install.yaml
@@ -1,7 +1,7 @@
 apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
 kind: Kustomization
 metadata:
-  name: rook-ceph-cluster-pre-install-v1.11.4
+  name: rook-ceph-cluster-prereq-jobs-v1.11.4
   namespace: ${releaseNamespace}
 spec:
   force: true
--- a/services/rook-ceph-cluster/1.11.4/pre-install.yaml
+++ b/services/rook-ceph-cluster/1.11.4/pre-install.yaml
@@ -1,7 +1,7 @@
 apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
 kind: Kustomization
 metadata:
-  name: rook-ceph-cluster-pre-install
+  name: rook-ceph-cluster-pre-install-v1.11.4
   namespace: ${releaseNamespace}
 spec:
   force: true
--- a/services/rook-ceph-cluster/1.11.4/pre-install/ceph-crd-check.yaml
+++ b/services/rook-ceph-cluster/1.11.4/pre-install/ceph-crd-check.yaml
@@ -35,12 +35,12 @@ subjects:
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: dkp-ceph-pre-install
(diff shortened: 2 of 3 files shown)


gracedo (approved):
we have a few places where we have pre-install (and maybe pre-upgrade) kustomizations defined.. maybe we can consider making this change everywhere if we rely on them running. I suspect it's not required in other apps since we haven't seen anything break yet, but worth an audit and cleanup. I think there are some pre-upgrade jobs that right now only run for the applicable affected version that we can maybe also just remove.

cbuto:
> we have a few places where we have pre-install (and maybe pre-upgrade) kustomizations defined.. maybe we can consider making this change everywhere if we rely on them running. I suspect it's not required in other apps since we haven't seen anything break yet, but worth an audit and cleanup. I think there are some pre-upgrade jobs that right now only run for the applicable affected version that we can maybe also just remove.

I can take a look at pre-upgrade jobs, most could probably be removed. I think the strange flux behavior we saw would only affect jobs that are meant to run as a part of pre-upgrade (pre-install jobs would be okay because the kustomization doesn't already exist). I'm also still not entirely sure why Flux was behaving that way 🤔

gracedo:
@cbuto I guess I find these pre-install and pre-upgrade kustomizations/jobs a little confusing because they are basically all being used interchangeably since we can't really distinguish between install and upgrade (vs actual helm hooks that only run pre-install before installs, pre-upgrade before upgrades). I'm not sure if it's more confusing to name them all the same (pick pre-install or pre-upgrade) or just another name altogether to make it obvious this thing is going to/needs to run before the helm chart gets installed and/or upgraded.

Just to reiterate.. we would need to version the kustomizations if they are required to run before the chart is upgraded. Otherwise they will still always run pre-install, but we can't rely on when they would run during upgrade.

cbuto:
> @cbuto I guess I find these pre-install and pre-upgrade kustomizations/jobs a little confusing because they are basically all being used interchangeably since we can't really distinguish between install and upgrade (vs actual helm hooks that only run pre-install before installs, pre-upgrade before upgrades). I'm not sure if it's more confusing to name them all the same (pick pre-install or pre-upgrade) or just another name altogether to make it obvious this thing is going to/needs to run before the helm chart gets installed and/or upgraded.
> 
> Just to reiterate.. we would need to version the kustomizations if they are required to run before the chart is upgraded. Otherwise they will still always run pre-install, but we can't rely on when they would run during upgrade.

totally agree. after looking at the jobs we have now, rook-ceph-cluster is the only service that still needs a "pre-upgrade" job (to wait for the operator to be upgraded). the others can be removed. maybe i should just rename it to `rook-ceph-cluster-prereq-jobs-v1.11.4` 🤔 

the other jobs we have for velero and loki are just needed prior to installing, so those are actual "pre-install" jobs. 

>Just to reiterate.. we would need to version the kustomizations if they are required to run before the chart is upgraded. Otherwise they will still always run pre-install, but we can't rely on when they would run during upgrade.

yep that's what we saw in the 2.4 upgrade at least

sonofgibs (approved):

PR #1258, opened 2023-05-16
chore: clean up old pre-upgrade jobs

cbuto:
these pre-upgrade jobs are left over from many versions ago, no need to keep them around now!

```release-note

```

**Checklist**


cbuto:
in addition to the CI checks, i ran through an upgrade manually to verify istio/jaeger ✅

gracedo (approved):

sonofgibs (approved):

PR #1290, opened 2023-05-24
feat: Set priority on kustomization jobs

gracedo:
Set priority class on pre/post-install jobs backed by kustomizations. These are one-off jobs that run at installation to properly set up the application, so I think that prioritizing them would be safe. Users won't be able to configure this, but as they run then exit, I don't think we would see that use case.

https://d2iq.atlassian.net/browse/D2IQ-97328

```release-note

```

**Checklist**

### DIFF
--- a/services/grafana-loki/0.69.14/grafana-loki-pre-install-jobs/pre-install.yaml
+++ b/services/grafana-loki/0.69.14/grafana-loki-pre-install-jobs/pre-install.yaml
@@ -43,6 +43,7 @@ spec:
     spec:
       serviceAccountName: grafana-loki-pre-install
       restartPolicy: OnFailure
+      priorityClassName: dkp-critical-priority
       containers:
         - name: pre-install
           image: bitnami/kubectl:1.24.6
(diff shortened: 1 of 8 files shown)


cbuto (approved):

services/grafana-loki/0.69.14/grafana-loki-pre-install-jobs/pre-install.yaml
```diff
@@ -95,4 +96,6 @@ spec:
                 echo "The CephCluster '.status.ceph.capacity.lastUpdated' field is not populated yet, retrying after 10s..."
                 sleep 10
               done
+
+              echo "CephCluster capacity updated at: $ceph_cluster_capacity"
```
cbuto, line 100: ceph is sloooooooow :P

bcmendoza (approved):

gracedo:
😢 it looks like setting `force` on helmrelease upgrades doesn't actually do a delete and recreate but a "replace" which I guess is still just a patch https://github.com/fluxcd/flux2/discussions/2854 I would like to be able to prioritize these jobs since they're so essential for setting up ceph, so I am thinking about adding `ttlSecondsAfterFinished` to the job. I will open a PR in m/charts to add this field, and remove the OBC changes from this PR to merge the other changes first

PR #1293, opened 2023-05-25
feat: Trigger docs automation for published releases

bcmendoza:
Triggers the Confluence docs automation in dkp-infra-tools when there is a new release published. It can be tested via workflow dispatch.

https://d2iq.atlassian.net/browse/D2IQ-97039

Workflow dispatch test: https://github.com/mesosphere/kommander-applications/actions/runs/5081036474/jobs/9130932074
Dispatched job: https://github.com/mesosphere/dkp-infra-tools/actions/runs/5081913964/jobs/9130935727

```release-note

```

**Checklist**

### DIFF
--- a/.github/workflows/publish-appversions.yml
+++ b/.github/workflows/publish-appversions.yml
@@ -9,24 +9,21 @@ jobs:
   publish-test:
     name: Publish docs test
     if: ${{ github.event_name == 'workflow_dispatch' }}
-    runs-on:
-    - self-hosted
-    - small
+    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-appversions.yml
```diff
@@ -9,29 +9,27 @@ jobs:
   publish-test:
     name: Publish docs test
     if: ${{ github.event_name == 'workflow_dispatch' }}
-    runs-on:
```
cbuto, line 12: we should stick with the self-hosted runners

cbuto (approved):

PR #1322, opened 2023-06-09
feat: Release DKP-Insights Backend v0.5.0-beta.1

MessKon:
Release DKP-Insights Backend v0.5.0-beta.1


licenses.d2iq.yaml
```diff
@@ -143,6 +147,14 @@ resources:
...
       - url: https://github.com/mesosphere/dkp-insights
         ref: main
+  - container_image: docker.io/mesosphere/insights:v0.2.3
+    sources:
+      - url: https://github.com/mesosphere/dkp-insights
+        ref: v0.2.3
+  - container_image: docker.io/mesosphere/insights:v0.4.1
+    sources:
+      - url: https://github.com/mesosphere/dkp-insights
+        ref: v0.4.1
```
cbuto, line 157: i think we can use `${image_tag}` here if the image tag matches the tag on the repo (which looks like it does)
```suggestion
        ref: ${image_tag}
```

services/dkp-insights/0.4.1/defaults/cm.yaml
```diff
@@ -0,0 +1,1486 @@
...
+          - name: seaweedfs-filer-config
+            configMap:
+              name: {{ include "chart.insightsSeaweedFSCMName" . }}
+              items:
+                - key: filer.toml
+                  path: filer.toml
+        grpcPort: 18888
+        initContainers: |
+          - name: init-db
+            image: docker.io/mesosphere/insights:v0.2.3
```
cbuto, line 1295: just out of curiosity, should this image tag match what's used here: https://github.com/mesosphere/kommander-applications/pull/1322/files#diff-4f759953a5a7257599836148fbf89038b59090ce1f26f7171ddd24a08f72259aR173? 

that would let us ship one less image in the airgapped bundle, but not sure if this was done intentionally or not

services/dkp-insights/metadata.yaml
```diff
@@ -0,0 +1,15 @@
+displayName: DKP Insights
+description: DKP Insights is D2iQ’s analytics tool used to ensure clusters are secure and in debugging, monitoring, and anticipating future issues with your cluster.
+category:
+  - observability
+  - monitoring
+certifications:
+  - airgapped
+type: catalog
```
cbuto, line 8: we might want this to be:
```suggestion
type: platform
```

licenses.d2iq.yaml
```diff
@@ -6,10 +6,14 @@ ignore:
   - docker.io/bitnami/kubectl:1.26.4
   - docker.io/bitnami/kubectl:1.24.1
   - docker.io/bitnami/memcached:1.6.15-debian-11-r8
+  - docker.io/bitnami/postgresql:11.14.0-debian-10-r28
   - docker.io/bitnami/postgresql:11.16.0-debian-11-r9
   - docker.io/library/busybox:1
+  - docker.io/chrislusf/seaweedfs:3.20
```
cbuto, line 12: we might be able to actually pull the licenses for seaweedfs and the fairwinds images below, I'm not totally sure how to figure that out though. @mhrabovcin might know best

cbuto (commented):
left a few comments!

kaiwalyajoshi:
@MessKon perhaps instead of using `v0.4.1` as the basis, we can use the newer `v0.5.0-beta.1` which was cut here: https://github.com/mesosphere/insights-catalog-applications/tree/beta/services/dkp-insights/0.5.0-beta.1

That will remove the need for `seaweedfs` and many of the other components raised in the review which really won't ship with the official `v0.5.0` releases.

licenses.d2iq.yaml
```diff
@@ -143,6 +147,10 @@ resources:
     sources:
       - url: https://github.com/mesosphere/dkp-insights
         ref: main
+  - container_image: docker.io/mesosphere/insights:v0.5.0-beta.1
+    sources:
+      - url: https://github.com/mesosphere/dkp-insights
+        ref: main
```
mhrabovcin, line 153: Here we should point to a specific tag. These is a replacement variable we can use to map container image tag to specific git tag by using.

```
        ref: ${image_tag}
```

We should update also the image mapping above `docker.io/mesosphere/insights-management:v0.5.0-beta.1` which points to the `main` as well.

MessKon: great. i'll update both, thank you

common/helm-repositories/mesosphere-repos.yaml
```diff
@@ -40,3 +40,15 @@ spec:
...
   timeout: 1m
   url: "${helmMirrorURL:=https://mesosphere.github.io/dkp-insights-charts-management}"
+---
+apiVersion: source.toolkit.fluxcd.io/v1beta1
+kind: HelmRepository
+metadata:
+  name: mesosphere.github.io-dkp-insights-charts-attached
+  namespace: kommander-flux
+  labels:
+    kommander.d2iq.io/dkp-airgapped: supported
```
cbuto, line 50: it looks like this is [added dynamically by one of our controllers](https://github.com/mesosphere/kommander/blob/ae6e771b09cfb2cb83f975260e0415942f87c269/federation/pkg/controllers/helpers/flux.go#L360-L372) for catalog apps, we can drop this (it probably wouldn't break anything but just to avoid confusion)

services/dkp-insights/0.5.0-beta.1/helmrelease.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
```
cbuto, line 1: my only other suggestion might be to move this HelmRelease behind a Flux Kustomization like the insights management app does: 

https://github.com/mesosphere/kommander-applications/blob/main/services/dkp-insights-management/0.5.0/helmrelease.yaml

this can make things more flexible if we need to add jobs to automate upgrade/install issues in the future (you can check out the [Loki app](https://github.com/mesosphere/kommander-applications/tree/main/services/grafana-loki/0.69.16) as an example). If we don't do this now, moving the HelmRelease later on would likely cause things to be uninstalled/reinstalled. 

I'm also good keeping this as-is since insights is probably less likely to need some kind of automation like that!

cbuto (approved):
had a few more suggestions but overall looks good to me!

mhrabovcin (approved):

PR #1340, opened 2023-06-16
fix: duplicate key in fluentbit config defaults

takirala:

The key for priority class is duplicated at https://github.com/mesosphere/kommander-applications/blob/04fd26194dee601b01a3bd482618daa2b393dd68/services/fluent-bit/0.30.3/defaults/cm.yaml#L13 and https://github.com/mesosphere/kommander-applications/blob/04fd26194dee601b01a3bd482618daa2b393dd68/services/fluent-bit/0.30.3/defaults/cm.yaml#L21

Found this when testing out the new config service endpoint.

```release-note

```

**Checklist**

### DIFF
--- a/services/fluent-bit/0.30.3/defaults/cm.yaml
+++ b/services/fluent-bit/0.30.3/defaults/cm.yaml
@@ -18,8 +18,6 @@ data:
         cpu: 350m
         memory: 250Mi
 
-    priorityClassName: system-node-critical
-
     securityContext:
       privileged: true
 


gracedo (approved):
Innnteresting. Does that mean we've been deploying it with system-node-critical in previous releases o_O I guess it's fine to lower that priority now?

cbuto (approved):

takirala:
> Innnteresting. Does that mean we've been deploying it with system-node-critical in previous releases o_O I guess it's fine to lower that priority now?

yeah i think its fine to lower the priority.

PR #1347, opened 2023-06-21
fix(istio): expose metrics to allow prometheus scraping

cbuto:
Fixes the metrics scraping issue found via the daily cluster alerts.

The istio metrics host defaults to `127.0.0.1`, which doesn't allow prometheus to scrape the istio operator metrics. 

I'm not sure if this fix should go into the [chart defaults](url) or here, open to suggestions!

https://d2iq.atlassian.net/browse/D2IQ-97182

```release-note

```

**Checklist**

### DIFF
--- a/services/istio/1.17.2/defaults/cm.yaml
+++ b/services/istio/1.17.2/defaults/cm.yaml
@@ -32,3 +32,8 @@ data:
       image: bitnami/kubectl
       tag: 1.26.4
       priorityClassName: "dkp-critical-priority"
+    operator:
+      # expose metrics for prometheus scraping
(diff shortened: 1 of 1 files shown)


gracedo (approved):
let's just keep it here, since istio is a forked chart so easier not to adjust the default values/templates there!

cbuto:
@gracedo sounds good!

also, i don't think we'll need to backport this since it was only introduced in https://github.com/mesosphere/charts/pull/1405. previous releases didn't scrape the operator metrics at all it seems.

takirala (approved):

PR #1350, opened 2023-06-23
feat: Bump dex to add priority to certs jobs

gracedo:
bump dex to latest, which includes fix to set priority on the certs jobs. since priorityClassName is an immutable field, this also requires a pre-upgrade hook to delete the jobs first, which will result in a re-run of the jobs. In my testing, I did not see any negative impact from regenerating the dex certs.

depends on https://github.com/mesosphere/charts/pull/1439

https://d2iq.atlassian.net/browse/D2IQ-97813

```release-note

```

**Checklist**


services/dex/2.12.3/dex.yaml
```diff
@@ -9,9 +9,9 @@ spec:
       chart: dex
       sourceRef:
         kind: HelmRepository
-        name: mesosphere.github.io-charts-stable
+        name: gracedo.github.io-charts-stable
```
gracedo, line 12: TODO

cbuto (approved):

cwyl02 (approved):

PR #1354, opened 2023-06-28
fix: bump knative

adityajariwala:
Fixes a bug - upstream Knative YAML files have a weird indentation error. Fixed in m/charts

```release-note

```

**Checklist**


adityajariwala:
@d2iq-mergebot test all

gracedo (commented):
Let's maybe test this installation manually on a cluster since our tests missed this last time (knative is probably not actually enabled in the tests)

PR #1359, opened 2023-06-30
fix: Don't change original permissions of etcd certificates

alembiewski:
This PR changes how etcd certificates are handled in `etcd-metrics-proxy` by copying original certs and modifying the copy. The existing approach modifies the original files, which causes the CIS benchmark to fail: 
```
Failed a CIS Benchmark test 1.1.19: Ensure that the Kubernetes PKI directory and file ownership is set to root:root (Automated)
```
Also, the PR bumps the `kube-rbac-proxy` image to the latest version.

https://d2iq.atlassian.net/browse/D2IQ-97242

```release-note

```

**Checklist**

### DIFF
--- a/services/kube-prometheus-stack/46.8.0/etcd-metrics-proxy/kube-rbac-proxy.yaml
+++ b/services/kube-prometheus-stack/46.8.0/etcd-metrics-proxy/kube-rbac-proxy.yaml
@@ -91,7 +91,7 @@ spec:
               mountPath: /etc/kube-rbac-proxy/etcd
       containers:
         - name: kube-rbac-proxy
-          image: quay.io/brancz/kube-rbac-proxy:v0.14.1
+          image: quay.io/brancz/kube-rbac-proxy:v0.14.2
           args:
             - "--secure-listen-address=0.0.0.0:8443"
             - "--tls-cert-file=/etc/kube-rbac-proxy/tls/tls.crt"
--- a/services/kube-prometheus-stack/46.8.0/etcd-metrics-proxy/kube-rbac-proxy.yaml
+++ b/services/kube-prometheus-stack/46.8.0/etcd-metrics-proxy/kube-rbac-proxy.yaml
@@ -75,25 +75,32 @@ spec:
       initContainers:
         - name: init
           image: docker.io/library/busybox:1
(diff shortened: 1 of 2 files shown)


gracedo:
```
Commit 560e1a5d67:
1: CT1 Title does not follow ConventionalCommits.org format 'type(optional-scope): description': "Update kube-rbac-proxy image to v0.14.2 in licenses.d2iq.yaml"
Commit caeb90908e:
1: CT1 Title does not follow ConventionalCommits.org format 'type(optional-scope): description': "Bump kube-rbac-proxy to v0.14.2"
```
`pre-commit` is failing due to un-conventional commits

services/kube-prometheus-stack/46.8.0/etcd-metrics-proxy/kube-rbac-proxy.yaml
```diff
@@ -75,25 +75,32 @@ spec:
       initContainers:
         - name: init
           image: docker.io/library/busybox:1
-          command: ["/bin/chown", "-R", "65532:65532", "/etc/certs/etcd"]
+          command: ["/bin/sh", "-c"]
+          args:
+            - >
+              cp /etc/certs/etcd/ca.crt /etc/certs/etcd/healthcheck* /etc/kube-rbac-proxy/etcd/ &&
```
cneth, line 81: I believe this copy would need to get 'refreshed' every time the etcd certs expire, right?

We use kubeadm to install kubernetes, which creates (by default) certs with a 1 year expiry.
However, in 2.4 we added a feature that allows the customer to set a custom renewal frequency.

https://docs.d2iq.com/dkp/2.5/aws-certificate-renewal

I am not sure if we have a mechanism to restart pods when the cluster PKI gets updated - we do have the 'reloader', but I believe it can only watch secrets/certs...

alembiewski: Turns out the control-plane nodes are restarted during certificate rotation, so the pods will be recreated automatically ([source](https://d2iq.slack.com/archives/C01MPKVBB5E/p1688737943790949?thread_ts=1688726068.985899&cid=C01MPKVBB5E)).

cneth (approved):

gracedo (approved):

cbuto (approved):

PR #1390, opened 2023-07-07
fix: explicitly declare this quota

cwyl02:
so that it works on GKE 1.26+

workaround is from: https://github.com/openservicemesh/osm/issues/4408#issuecomment-1302008888

```release-note

```

**Checklist**


cwyl02:
accidently opened an unwanted PR

PR #1392, opened 2023-07-08
fix: Remove resourcequota from gk/fb, add to certm

gracedo:
GKE clusters limit priority class consumption of system-cluster-critical and system-node-critical by default, to get around it we must create ResourceQuota in each ns that will deploy pods using that priority.

Prior to upcoming 2.7 release, we were only using the system-*-critical classes in apps that were setting it by default, which were gatekeeper, fluent-bit, and ceph. Gatekeeper deploys the resourcequota by default to ensure that it deploys on GKE clusters (https://github.com/open-policy-agent/gatekeeper/pull/1282), and we added resourcequota manifest to fluent-bit to also resolve this issue on GKE clusters. Now that we are setting priority class on all our apps, many of which are using the k8s critical classes, we should be deploying these resourcequotas at the DKP level, not by specific apps.

This PR removes the resourcequotas from gatekeeper and fluent-bit. There will be a PR in kommander to create these resourcequotas for every workspace and project ns on the attached clusters instead.

One exception is cert-manager, which is using system-cluster-critical and is deployed to the `cert-manager` namespace instead of the workspace namespace -- we should look into a more automated way to deploy resource quotas into all targeted namespaces, not just the workspace namespace, but in interest of time (upcoming release) we can make an exception for cert-manager while we think about how to solve this long-term for a future release. To note, if a user decided to override priority for some apps that target other namespaces (e.g. centralized-kubecost) to use `system-cluster-critical` then it would fail to deploy since that app deploys into the `kubecost` namespace which would not have the resourcequota. I think this is unlikely to happen. If in the rare even it does, we can ask the user to create the resourcequota manually.

https://d2iq.atlassian.net/browse/D2IQ-98169

```release-note

```

**Checklist**


cwyl02 (approved):

sonofgibs (approved):

services/cert-manager/1.12.2/priorityclass-resource-quota/resourcequota.yaml
```diff
@@ -1,14 +1,15 @@
 apiVersion: v1
 kind: ResourceQuota
 metadata:
-  name: fluent-bit-critical-pods
-  namespace: ${releaseNamespace}
+  name: cert-manager-critical-pods
+  namespace: cert-manager
 spec:
   hard:
-    pods: "1G"
```
cbuto, line 8: should we keep `1G` (a large number) since this isn't configurable? 🤔

i doubt we'd ever need to change it since i don't think cert-manager scales horizontally or anything

gracedo: yes! now that we know what 1G is (probably)

cbuto (approved):

PR #1398, opened 2023-07-13
feat: set mtls-proxy image in a defaults CM

cbuto:
Sets the mtls-proxy image tag/repo in a defaults configmap that is deployed alongside the HR. Previously, this was defined directly in the prometheus proxy controller.

This also bumps the ghostunnel image based on the recently built image on our fork https://github.com/mesosphere/ghostunnel.

https://d2iq.atlassian.net/browse/D2IQ-98180

```release-note

```

**Checklist**

### DIFF
--- a/services/kommander/0.6.0/dynamic-helmreleases/mtls-proxy/defaults/cm.yaml
+++ b/services/kommander/0.6.0/dynamic-helmreleases/mtls-proxy/defaults/cm.yaml
@@ -1,10 +1,9 @@
 ---
-# this is deployed by the kommander kustomization
 apiVersion: v1
 kind: ConfigMap
 metadata:
   name: mtls-proxy-0.1.8-d2iq-defaults
-  namespace: ${releaseNamespace}
+  namespace: ${namespace}
 data:
   values.yaml: |
     # built from https://github.com/mesosphere/ghostunnel
--- a/services/kommander/0.6.0/dynamic-helmreleases/mtls-proxy/defaults/kustomization.yaml
+++ b/services/kommander/0.6.0/dynamic-helmreleases/mtls-proxy/defaults/kustomization.yaml
@@ -1,4 +1,5 @@
+---
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - kommander.yaml
+- cm.yaml


services/kommander/0.6.0/dynamic-helmreleases/mtls-proxy/defaults/cm.yaml
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: mtls-proxy-0.1.8-d2iq-defaults
```
mhrabovcin, line 5: Why there is `0.1.8` in the name?

cbuto: that's from the mtls-proxy chart version

https://github.com/mesosphere/kommander-applications/blob/9cebeb79ac009fe39b0382b66cc60a23f42039d2/services/kommander/0.6.0/dynamic-helmreleases/mtls-proxy/mtls-proxy.yaml#L19

sonofgibs (approved):

takirala (approved):

services/kommander/0.6.0/dynamic-helmreleases/mtls-proxy/defaults/cm.yaml
cbuto, line 5: this might not work since we create multiple kustomizations per cluster that would deploy this same CM. so I either need to use a unique name here or move this back to the controller 🤔

jimmidyson: Can't we deploy this defaults cm as part of the kommander kustomization? Then it will be available to use when the controller creates the dynamic mtls proxy helmreleases.

jimmidyson: There should only be a single defaults configmap rather than one per deployment I think.

cbuto: yeah agreed, I’ll move the creation to the kommander kustomization and rename it so it’s only created once

jimmidyson (approved):

services/kommander/0.6.0/kustomization.yaml
```diff
@@ -1,4 +1,5 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
+  - dynamic-helmreleases/mtls-proxy/defaults/cm.yaml
```
gracedo, line 4: (discussed in standup but commenting for posterity) -- this means the default cm only gets deployed once into the kommander ns -- but the mtls-proxy HR is deployed into workspace namespaces, so I believe the cm's would need to exist in the same namespace

cbuto: yup thanks, i'll need to see if its possible to deploy once per WS with this current structure 🤔

cbuto: i think i'll have to update the controller to create a Kustomization pointing to just the defaults CM directory

jimmidyson: Ah man that is annoying, thanks for your diligence @gracedo

sonofgibs (approved):

adityajariwala (approved):

cbuto:
closing for now, see https://github.com/mesosphere/kommander/pull/3612#issuecomment-1642423972

PR #1412, opened 2023-07-19
feat: Update jaeger allinone img to match operator version

gracedo:
noticed that the jaeger `allinone` image was using a pretty outdated version -- updating it to the same version that the operator is running.
ALSO: noticed most of our dashboard cm's had outdated versions, updated them here as well

```release-note

```

**Checklist**


cbuto (approved):
good catch!

takirala (approved):

PR #1417, opened 2023-07-20
feat: Add pre-upgrade job to delete jaeger HR

gracedo:
For some reason, the newly added `priorityClassName` field to the jaeger spec in the default cm is not getting applied to the existing jaeger CR in the cluster on upgrade. There may be a race when the jaeger CRD is updated (which adds that new spec field) and when the config is applied? I also tried setting `force: true` on the upgrade spec on the helmrelease (to force the recreation of resources on upgrade vs updating), but I believe this only applies to the new version onwards, not to the previous version being upgraded (as the previous HR spec does not have force set to true on upgrades). I also tried a job to delete the `jaeger` CR prior to upgrading, but the new CR applied on upgrade still did not have the priority field set. Deleting the HR altogether and letting it come back up to the latest version works (as expected -- fresh installs of the jaeger app did not run into this issue).

I added a check to the HR version to ONLY delete it if it is the previous version (running on dkp 2.5) to avoid needlessly recreating the HR on upgrades within 2.6. We can remove the job in 2.7. 

I tested this on the daily by deploying v2.40.0, then upgrading. The job deleted the jaeger HR, and when the new one came up, the priority class was set properly on the jaeger spec and resulting operator pod.

job logs to delete hr:
```
checking jaeger helmrelease and deleting it pre-upgrade if v2.40.0
version 2.40.0
deleting jaeger helmrelease
helmrelease.helm.toolkit.fluxcd.io "jaeger" deleted
```

after upgrade, `jaeger` spec:
```
spec:
  agent:
    config: {}
    options: {}
    resources: {}
  allInOne:
    config: {}
    image: jaegertracing/all-in-one:1.46.0
    metricsStorage: {}
    options:
      query:
        base-path: /dkp/jaeger
    priorityClassName: dkp-high-priority
    resources: {}
```

and jaeger pods
```
jaeger-jaeger-operator-77645f4c8b-2nft4                                    dkp-high-priority
jaeger-jaeger-operator-jaeger-d49d8dbc6-985xd                              dkp-high-priority
delete-jaeger-helmrelease-bm7nr                                            dkp-high-priority
```

https://d2iq.atlassian.net/browse/D2IQ-98315

```release-note

```

**Checklist**

### DIFF
--- a/services/jaeger/2.46.2/pre-upgrade/delete-jaeger-hr.yaml
+++ b/services/jaeger/2.46.2/pre-upgrade/delete-jaeger-hr.yaml
@@ -1,0 +1,73 @@
+# For upgrades from DKP <2.6 to 2.6.x.
+# **REMOVE THIS AFTER 2.6 IS RELEASED**
+# Delete the jaeger helmrelease prior to upgrading.
+# This must be done because upgrading does not apply the newly added priorityClassName field on the
+# Jaeger spec, perhaps because the CRD has not been upgraded yet. Recreating the HelmRelease applies the
+# Jaeger spec properly. This job will only delete the v2.40.0 helmrelease (from 2.5), so rerunning this job
+# during upgrades within 2.6 will not result in the recreation of the helmrelease.
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: jaeger-pre-upgrade
+  namespace: ${releaseNamespace}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: jaeger-pre-upgrade
+  namespace: ${releaseNamespace}
+rules:
+  - apiGroups: ["helm.toolkit.fluxcd.io"]
+    resources: ["helmreleases"]
+    verbs: ["get", "watch", "list", "delete"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: jaeger-pre-upgrade
+  namespace: ${releaseNamespace}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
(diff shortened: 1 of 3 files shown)


services/jaeger/2.46.2/pre-upgrade/delete-jaeger-hr.yaml
```diff
@@ -0,0 +1,73 @@
...
+              /bin/bash <<'EOF'
+              set -o nounset
+              set -o errexit
+              set -o pipefail
+
+              echo "checking jaeger helmrelease and deleting it pre-upgrade if v2.40.0"
+
+              VERSION=$(set -o errexit; kubectl get helmrelease -n ${releaseNamespace} jaeger --ignore-not-found -o=jsonpath='{.spec.chart.spec.version}')
+              echo "version $VERSION"
+              if [[ $VERSION == "2.40.0" ]]; then
```
cbuto, line 68: we likely won't bump this jaeger version in a 2.5 patch, but if we do then the upgrade from that patch version to 2.6.0 won't have the priority class set 🤔

gracedo: yeah, I don't think we really ever bump versions in patch releases unless there is a major bug; I think we should be safe here, jaeger is generally one of our more stable apps that we haven't really had any bug reports for. but to be safe, I guess I can make it trigger for anything less than 2.46.2

cwyl02 (approved):

cbuto (approved):

services/jaeger/2.46.2/pre-upgrade/delete-jaeger-hr.yaml
cbuto: looks good! thanks

PR #1435, opened 2023-07-28
ci: Use prev k8s version for upgrade e2e test

gracedo:
Made the change to use the same k8s version for upgrade tests since the previous version is compatible on 1.26

Held off on updating to 1.27 because it looks like there were issues with getting that build working in CI

depends on https://github.com/mesosphere/kommander-applications/pull/1455

```release-note

```

**Checklist**

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -9,12 +9,12 @@ E2E_TIMEOUT       ?= 120m
 # flexible with our testing as well as testing against the same patch version as we deliver
 # by default with DKP.
 # See https://github.com/mesosphere/kind-docker-image-automation/ for the build repo.
-E2E_KINDEST_IMAGE ?= "mesosphere/kind-node-ci:v1.26.3"
+E2E_KINDEST_IMAGE ?= "ghcr.io/mesosphere/kind-node-ci:v1.26.6"
 
 # Kommander applications are upgraded on the previous k8s version first, before the cluster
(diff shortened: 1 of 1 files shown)


jimmidyson:
This is a tricky one because we do actually still install kommander on previous kubernetes version before kube version upgrade. That is the supported upgrade process afaik. We should probably test upgrading from both previous k8s version, and current version. Wdyt @gracedo?

cbuto (approved):

gracedo:
> This is a tricky one because we do actually still install kommander on previous kubernetes version before kube version upgrade. That is the supported upgrade process afaik. We should probably test upgrading from both previous k8s version, and current version. Wdyt @gracedo?

Hmm, you're right that it would be more accurate to use the previous k8s version for the upgrade test, because k-apps are upgraded first before upgrading k8s (at least for now). Although the main focus of the test here is to test the application upgrades (not too much focus on k8s version) we might as well mimic the actual upgrade scenario since it is easy to here. Install tests will still happen on current k8s version, upgrade tests will happen on previous k8s version.

takirala (approved):

gracedo:
I believe it's already been seen that current version gatekeeper is not compatible with 1.27 so the tests may fail -- will revert the last commit (trying out 1.27) if I see the same. Will let https://d2iq.atlassian.net/browse/D2IQ-97976 take care of the bump in that case once we get the apps in shape (have added a line to that ticket to bump the test version here)

PR #1439, opened 2023-07-31
feat: Add pre-upgrade jobs to run helm-mapkubeapis

cbuto:
Add pre-upgrade jobs to kubecost, logging-operator, and gatekeeper to run helm-mapkubeapis tool that takes care of editing the helmrelease secret to remove the podsecuritypolicy manifest which is unsupported starting in k8s 1.25.

This affects clusters that were upgraded from 2.4, in which the PSP api was still supported and these 3 charts had deployed PSPs. Since kommander applications are upgraded prior to k8s, we saw this:
- upgrade to kommander 2.5 --> these apps upgraded successfully, with PSP resources still deployed
- upgrade to k8s 1.25 --> PSP api was removed, so all existing PSPs are deleted
- upgrade to kommander 2.6 --> upgrade errors in these 3 apps, explanation here: https://helm.sh/docs/topics/kubernetes_apis/#helm-users

> Helm fails in this scenario because it attempts to create a diff patch between the current deployed release (which contains the Kubernetes APIs that are removed in this Kubernetes version) against the chart you are passing with the updated/supported API versions. The underlying reason for failure is that when Kubernetes removes an API version, the Kubernetes Go client library can no longer parse the deprecated objects and Helm therefore fails when calling the library. Helm unfortunately is unable to recover from this situation and is no longer able to manage such a release. 

The following section https://helm.sh/docs/topics/kubernetes_apis/#updating-api-versions-of-a-release-manifest tells us how to recover from this error -- including using this tool https://github.com/helm/helm-mapkubeapis to help us do it easily via a single command. We have included this plugin into the kubetools image: https://github.com/mesosphere/kommander/pull/3657 and are using this image in these newly added pre-upgrade jobs.

```release-note

```

**Checklist**

### DIFF
--- a/services/gatekeeper/3.12.0/pre-upgrade/pre-upgrade.yaml
+++ b/services/gatekeeper/3.12.0/pre-upgrade/pre-upgrade.yaml
@@ -13,6 +13,9 @@ rules:
   - apiGroups: [""]
     resources: ["secrets"]
     verbs: ["get", "watch", "list", "update", "patch", "create"]
+  - apiGroups: ["helm.toolkit.fluxcd.io"]
+    resources: ["helmreleases"]
+    verbs: ["get", "watch", "list"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
@@ -51,8 +54,13 @@ spec:
             - |-
               /bin/bash <<'EOF'
               set -o nounset
-              set -o errexit
               set -o pipefail
+
+              kubectl get helmreleases.helm.toolkit.fluxcd.io -n ${releaseNamespace} gatekeeper
+              if [[ $? -ne 0 ]]; then
+                echo "Since the gatekeeper HelmRelease does not exist, this might not be an upgrade scenario. Exiting..."
+                exit 0
+              fi
 
               helm mapkubeapis kommander-gatekeeper --namespace ${releaseNamespace}
               EOF
--- a/services/gatekeeper/3.12.0/pre-upgrade/pre-upgrade.yaml
+++ b/services/gatekeeper/3.12.0/pre-upgrade/pre-upgrade.yaml
@@ -56,5 +56,5 @@ spec:
 
               # TODO: wait for HelmRelease to not be in-progress?
 
-              helm mapkubeapis gatekeeper --namespace ${releaseNamespace}
+              helm mapkubeapis kommander-gatekeeper --namespace ${releaseNamespace}
               EOF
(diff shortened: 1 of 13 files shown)


services/kubecost/0.35.1/pre-upgrade/pre-upgrade.yaml
```diff
@@ -0,0 +1,61 @@
...
+  template:
+    metadata:
+      name: kubecost-pre-upgrade
+    spec:
+      serviceAccountName: kubecost-pre-upgrade
+      restartPolicy: OnFailure
+      priorityClassName: dkp-high-priority
+      containers:
+        - name: kubetools
+          image: mesosphere/kommander2-kubetools:${kommanderAppManagementKubetoolsImageRepository}
```
cbuto, line 48: will this kubetools image tag replace properly?

services/kubecost/0.35.1/pre-upgrade/pre-upgrade.yaml
```diff
@@ -0,0 +1,61 @@
...
+          command:
+            - sh
+            - "-c"
+            - |-
+              /bin/bash <<'EOF'
+              set -o nounset
+              set -o errexit
+              set -o pipefail
+              
+              # TODO: wait for HelmRelease to not be in-progress?
```
cbuto, line 58: maybe we need something like https://github.com/mesosphere/charts/blob/master/stable/object-bucket-claim/templates/object-bucket-claim.yaml#L119-L127?

not sure if that matters or not

services/kubecost/0.35.1/pre-upgrade/pre-upgrade.yaml
gracedo: updated to use kommander chart version `${kommanderChartVersion:=v2.6.0-dev}`

cbuto: we'll just need to update the release script to handle replacing that in these files 👀

mikolajb (approved):

services/kubecost/0.35.1/pre-upgrade/pre-upgrade.yaml
cbuto: done in https://github.com/mesosphere/kommander-applications/pull/1439/commits/85c4a3dcb8520e4274fa7889c7968a37bb982416

gracedo (approved):

services/gatekeeper/3.12.0/pre-upgrade/pre-upgrade.yaml
```diff
@@ -0,0 +1,58 @@
...
+          command:
+            - sh
+            - "-c"
+            - |-
+              /bin/bash <<'EOF'
+              set -o nounset
+              set -o errexit
+              set -o pipefail
+
+              helm mapkubeapis kommander-gatekeeper --namespace ${releaseNamespace}
```
cbuto, line 57: we need to add a check here if the HR exists, to avoid initial installs failing

takirala (approved):

PR #1437, opened 2023-07-31
fix: disable PSPs explictly to avoid upgrade failures

palexster:
**What problem does this PR solve?**:

https://d2iq.atlassian.net/browse/D2IQ-98509

**Which issue(s) does this PR fix?**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Checklist**

- [ ] If the PR adds a version bump, ensure there is no breaking change in Licensing model (or NA).
- [ ] If a chart is changed or app configuration is significantly changed, the chart version is correctly incremented (so that apps are not automatically upgraded from a previous version of DKP).

PR #1440, opened 2023-07-31
fix: disable PSPs explictly to avoid upgrade failures (2.6)

gracedo:
* fix: disable PSPs

* fix(gatekeeper): Disable psps by default

PSP was removed in k8s 1.25, we should explicitly disable PSPs in our default values.

---------

backports https://github.com/mesosphere/kommander-applications/pull/1437

```release-note

```

**Checklist**


lukeogg (approved):
LGTM!

PR #1454, opened 2023-08-01
feat: Add pre-upgrade jobs to run helm-mapkubeapis (#1439) (2.5)

cbuto:
Add pre-upgrade jobs to kubecost, logging-operator, and gatekeeper to run helm-mapkubeapis tool that takes care of editing the helmrelease secret to remove the podsecuritypolicy manifest which is unsupported starting in k8s 1.25.

This affects clusters that were upgraded from 2.4, in which the PSP api was still supported and these 3 charts had deployed PSPs. Since kommander applications are upgraded prior to k8s, we saw this:
- upgrade to kommander 2.5 --> these apps upgraded successfully, with PSP resources still deployed
- upgrade to k8s 1.25 --> PSP api was removed, so all existing PSPs are deleted
- upgrade to kommander 2.6 --> upgrade errors in these 3 apps, explanation here: https://helm.sh/docs/topics/kubernetes_apis/#helm-users

> Helm fails in this scenario because it attempts to create a diff patch between the current deployed release (which contains the Kubernetes APIs that are removed in this Kubernetes version) against the chart you are passing with the updated/supported API versions. The underlying reason for failure is that when Kubernetes removes an API version, the Kubernetes Go client library can no longer parse the deprecated objects and Helm therefore fails when calling the library. Helm unfortunately is unable to recover from this situation and is no longer able to manage such a release. 

The following section https://helm.sh/docs/topics/kubernetes_apis/#updating-api-versions-of-a-release-manifest tells us how to recover from this error -- including using this tool https://github.com/helm/helm-mapkubeapis to help us do it easily via a single command. We have included this plugin into the kubetools image: https://github.com/mesosphere/kommander/pull/3657 and are using this image in these newly added pre-upgrade jobs.

```release-note

```

**Checklist**


cbuto:
I've tested:

* upgrade from 2.4 to 2.5.2 (fixed by disabling the PSPs explicitly on upgrade) ✅ 
* upgrade from 2.4 to 2.5.1 to 2.5.2 ✅ 
* installs for 2.5.2 ✅

services/gatekeeper/3.11.1/pre-upgrade/pre-upgrade.yaml
```diff
@@ -0,0 +1,65 @@
...
+              /bin/bash <<'EOF'
+              set -o nounset
+              set -o pipefail
+
+              kubectl get helmreleases.helm.toolkit.fluxcd.io -n ${releaseNamespace} gatekeeper
+              if [[ $? -ne 0 ]]; then
+                echo "Since the gatekeeper HelmRelease does not exist, this might not be an upgrade scenario. Exiting..."
+                exit 0
+              fi
+
```
cwyl02, line 63: nit: add a comment about `kommander-gatekeeper` value is from `.spec.releaseName` of the gatekeeper HR -- this seems a bit different than the other 2 "normal" cases.

cwyl02 (approved):

services/kubecost/0.33.1/kustomization.yaml
gracedo, line 1: can you bump kubecost version? I forgot to in https://github.com/mesosphere/kommander-applications/commit/861da23d6c91666bf31c1f3acf76248d876ef9b6 😅

cbuto: yup will do.... i think we missed centralized-kubecost in all of this as well 🤔

gracedo (approved):

PR #1462, opened 2023-08-04
fix: not skipping authz for kube dashboard

cwyl02:
TFA-full ingresses are skipping the k8s API authorization thus a user can just access the URL(just seeing the empty page) without any RBAC permission

see TFA Code about this config option:

https://github.com/mesosphere/traefik-forward-auth/blob/master/internal/handlers/server.go#L497-L504

https://d2iq.atlassian.net/browse/D2IQ-97861#

```release-note

```

**Checklist**

### DIFF
--- a/services/traefik-forward-auth-mgmt/0.3.9/defaults/cm.yaml
+++ b/services/traefik-forward-auth-mgmt/0.3.9/defaults/cm.yaml
@@ -34,6 +34,9 @@ data:
       whitelist: []
       enableRBAC: true
       enableImpersonation: true
+      rbacPassThroughPaths:
+        - "/dkp/kiali/"
+        - "/dkp/kiali/*"
     ingress:
       enabled: true
       annotations:
--- a/services/traefik-forward-auth-mgmt/0.3.9/defaults/cm.yaml
+++ b/services/traefik-forward-auth-mgmt/0.3.9/defaults/cm.yaml
@@ -34,11 +34,6 @@ data:
       whitelist: []
       enableRBAC: true
       enableImpersonation: true
-      rbacPassThroughPaths:
-        - "/dkp/kubernetes/"
-        - "/dkp/kubernetes/*"
-        - "/dkp/kiali/"
-        - "/dkp/kiali/*"
     ingress:
       enabled: true
       annotations:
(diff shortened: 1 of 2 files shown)


cwyl02:
Should hold merging the 2.6 backport until 2.6 GA

services/traefik-forward-auth-mgmt/0.3.9/defaults/cm.yaml
```diff
@@ -34,11 +34,6 @@ data:
       whitelist: []
       enableRBAC: true
       enableImpersonation: true
-      rbacPassThroughPaths:
```
gracedo, line 37: was just looking back in history to try to understand why this was added in the first place and if we want to be removing this since it was explicitly added in the past.. most I can find is https://github.com/mesosphere/kubernetes-base-addons/pull/815/commits/346aede1c7cd9270d11f6e53171e616ff976966f where kiali was added because "skip tfa rbac as kiali handles it" (commit message). Just want to make sure we actually want to undo this (not sure what was the impetus in the first place), let's check with product (and deepak authored this commit in the past so check with him again, maybe he can recall why 😄 )

cwyl02: true. Maybe keep Kiali there but only removing kube dashboard?

services/traefik-forward-auth-mgmt/0.3.9/defaults/cm.yaml
```diff
@@ -34,11 +34,6 @@ data:
...
-        - "/dkp/kubernetes/"
```
cwyl02, line 38: For kube dashboard tho, I think if we either do this removal here or we remove the Kube dashboard Roles. Otherwise it's a slightly confusing UI.

services/traefik-forward-auth-mgmt/0.3.9/defaults/cm.yaml
cwyl02: talked to Deepak, kiali needs to stay

sonofgibs (approved):

gracedo (approved):

PR #1465, opened 2023-08-07
ci: migrate tc remaining builds to gha

shaeed-noordeen:
Migration from TC -> GHA

```release-note

```

**Checklist**


.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+    if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}
+    runs-on:
+      - self-hosted
+      - kind-large
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
```
cbuto, line 48: do we need to login to Dockerhub/nexus in both the install and upgrade tests?

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+      - uses: webfactory/ssh-agent@v0.7.0
+        with:
+          ssh-private-key: ${{ secrets.PRIVATE_SSH_KEY }}
+
+      - name: Setup go
```
cbuto, line 53: it looks like the upgrade tests below don't have this step and they passed, maybe we can drop this?

.github/workflows/create-dev-tag.yml
```diff
@@ -0,0 +1,86 @@
...
+
+env:
+  GOWORK: off
+  GOPRIVATE: github.com/mesosphere
+  DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+  DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+  GH_ACCESS_TOKEN: ${{ secrets.MESOSPHERECI_USER_TOKEN }}
+
+jobs:
+  create-dev-tag:
```
cbuto, line 42: so we need to run this job on a cron for every release branch, is that what this job currently does? 🤔

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+
+      - name: Tar files
+        run: sudo tar -cvzf report-e2e.tar .tmp/kommander-e2e/.e2e/
+        if: always()
+
+      - name: Upload report-e2e.json
+        uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: report-e2e
```
cbuto, line 77: lets rename this
```suggestion
          name: install-e2e-tests
```

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,114 @@
...
```
cbuto, line 113:
```suggestion
          name: upgrade-e2e-tests
```

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - uses: webfactory/ssh-agent@v0.7.0
+        with:
+          ssh-private-key: ${{ secrets.PRIVATE_SSH_KEY }}
+
+      - name: Setup asdf
+        uses: asdf-vm/actions/setup@v1
```
cbuto, line 98: >The following actions uses node12 which is deprecated and will be forced to run on node16: asdf-vm/actions/setup@v1. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

```suggestion
        uses: asdf-vm/actions/setup@v2
```

.github/workflows/manifest-validation.yml
```diff
@@ -0,0 +1,31 @@
...
+      - small
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - uses: webfactory/ssh-agent@v0.7.0
+        with:
+          ssh-private-key: ${{ secrets.PRIVATE_SSH_KEY }}
+
+      - name: Setup asdf
+        uses: asdf-vm/actions/setup@v1
```
cbuto, line 26:
```suggestion
        uses: asdf-vm/actions/setup@v2
```

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,109 @@
+name: E2E Tests
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled]
```
gracedo, line 4: we shouldn't rerun e2e tests every time the PR is labeled, it'll be pretty costly (e.g. just now I added the `ready-for-review` label and it triggered the tests). this came up recently in our team as well and Tarun mentioned looking into https://github.com/orgs/community/discussions/27031#discussioncomment-3254376

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,109 @@
+name: E2E Tests
```
gracedo, line 1: @shaeed-noordeen Let's remove the e2e tests from being run in GHA and leave them in TC for now -- we plan on removing these tests altogether in favor of application specific testing which is underway this release.

gracedo (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+      - 'v*'
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  release:
+    name: Release
+    if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}
```
cbuto, line 15: we can remove this `if` statement, this will only run when tags are pushed

.github/workflows/create-dev-tag.yml
```diff
@@ -0,0 +1,71 @@
...
+    runs-on:
+    - self-hosted
+    - small
+    strategy:
+      matrix:
+        branch:
+          - main
+          - release-2.1
```
cbuto, line 25: we can remove `release-2.1` and `release-2.2` from this list

.github/workflows/create-dev-tag.yml
```diff
@@ -0,0 +1,71 @@
+name: Create Development Tag
+
+on:
+  workflow_dispatch:
+
+  schedule:
+    - cron: '5 20 * * *'
```
cbuto, line 7: the teamcity trigger is set to 20:05 PST every day. but this cron is using UTC. 

I think we want to use this:

```suggestion
    - cron: '5 3 * * *'
```

.github/workflows/create-dev-tag.yml
```diff
@@ -0,0 +1,71 @@
...
+    strategy:
+      matrix:
+        branch:
+          - main
+          - release-2.1
+          - release-2.2
+          - release-2.3
+          - release-2.4
+          - release-2.5
+          - release-2.6
```
mikolajb, line 30: We need to automate adding to this list or find a different solution

shaeed-noordeen: cant we edit the yaml file when a new branch is needed?

cbuto:
@shaeed-noordeen looks good but will we handle moving the [pre-commit job](https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_Kommander2_KommanderApplications_PreCommitChecks/4046811?) in another PR?

.github/workflows/create-dev-tag.yml
```diff
@@ -0,0 +1,69 @@
...
+  GH_ACCESS_TOKEN: ${{ secrets.MESOSPHERECI_USER_TOKEN }}
+
+jobs:
+  create-dev-tag:
+    runs-on:
+    - self-hosted
+    - small
+    strategy:
+      matrix:
+        branch:
```
cbuto, line 23: looking at this again, this will run on every branch when we cut release branches? I think we only want it to run on `main` if we are listing the release branches like this.

I think we can add an `if` statement to specify that this job should only run on main?

shaeed-noordeen:
> Member



> @shaeed-noordeen looks good but will we handle moving the [pre-commit job](https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_Kommander2_KommanderApplications_PreCommitChecks/4046811?) in another PR?

Doesn't the lint workflow already cover the pre-commit?
https://github.com/mesosphere/kommander-applications/blob/main/.github/workflows/lint.yaml

we should still move it to self-hosted runners though

cbuto:
> > Member
> 
> > @shaeed-noordeen looks good but will we handle moving the [pre-commit job](https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_Kommander2_KommanderApplications_PreCommitChecks/4046811?) in another PR?
> 
> Doesn't the lint workflow already cover the pre-commit? https://github.com/mesosphere/kommander-applications/blob/main/.github/workflows/lint.yaml
> 
> we should still move it to self-hosted runners though

oh yeah you are right. for some reason we still had the TC job running.

cbuto (approved):

.github/workflows/create-dev-tag.yml
mikolajb: I mean, we can but we don't want to 🙂

PR #1466, opened 2023-08-07
fix: remove postRenderer from Loki HelmRelease (2.5)

cbuto:
The fix for the missing annotations was included in `0.69.4` so we can remove these post render patches now

```release-note

```

**Checklist**


gracedo (approved):

cbuto:
https://github.com/mesosphere/kommander/pull/3704 is related to this change since i had to bump the loki versions

PR #1510, opened 2023-08-21
feat: Bump gitea to 9.2.0

gracedo:
Upgrades the following apps to use version 9.2.0 of chart gitea:
* gitea from version 8.2.0 to 9.2.0


services/gitea/9.2.0/defaults/cm.yaml
```diff
@@ -62,12 +73,11 @@ data:
...
-      image:
-        tag: 1.6.15-debian-11-r8
+    postgresql-ha:
+      enabled: false
     postgresql:
+      enabled: true
       primary:
         priorityClassName: "dkp-critical-priority"
-      image:
-        tag: 11.16.0-debian-11-r9
```
cbuto, line 73: in the last major gitea bump, we explicitly wanted to pin this postgres version to avoid having to deal with upgrading to postgres 15, which is a non-trivial task to automate.

gracedo: ahh good call, I couldn't figure out the reasoning behind why it was pinned :)

PR #1511, opened 2023-08-22
fix(grafana): Use KPS for grafana-logging service discovery

gracedo:
project-grafana-logging was failing due to permission errors trying to fetch crds (likely to check for the servicemonitor crd). we limit projects from being able to access cluster resources or resources from other namespaces. instead, let's disable the service monitor for grafana-logging apps and use labels for service discovery

```
Helm install failed: unable to build kubernetes objects from release manifest: error validating "": error validating data: failed to check CRD: failed to list CRDs: customresourcedefinitions.apiextensions.k8s.io is forbidden: User "system:serviceaccount:kuttl-prj-2:kuttl-prj-2" cannot list resource "customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope
```

```release-note

```

**Checklist**


gracedo:
the failing `kommander2/test/multi-cluster/logging-per-project` kuttl test is passing with this fix: https://github.com/mesosphere/kommander/pull/3787

cbuto (approved):

PR #1514, opened 2023-08-30
feat: add ai-navigator-app

mstruebing:
* Closes D2IQ-98636

This PR adds the ai-navitgator-app to our `kommander-applications` catalog.

https://d2iq.atlassian.net/browse/D2IQ-98636

~~This PR is a WIP.~~
~~We need to change the description and the icon.~~

~~I've applied these resources to the daily and get:~~
```
│ 2023/08/30 09:29:11 licenses.kommander.mesosphere.io is forbidden: User "system:serviceaccount:kommander:default" cannot list resource "licenses" in API group "kommander.mesosphere.io" in the namespace "kommander"                                                                      │
```

~~What do I need to do so that the service would be able to read licenses and the corresponding secret?~~

~~Also, is it okay to hardcode the namespace to `kommander`? (this application should always be deployed in this namespace)~~

~~Should I split these resources into multiple files or is it okay to have them in one?~~

I think this needs to be merged into the 2.6 branch to be available in 2.6.1.
I'm not sure how commit messages come into play here, but should I change it to  `fix`?

~~Is there a way to apply the catalog from this branch on a cluster? Right now I simply applied the `Deployment` and `Service` definition, but for a real test I would like to use it like it would be used in reality.~~

```release-note

```

**Checklist**


services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
```diff
@@ -0,0 +1,52 @@
+# TODO: Should these be separate files?
+---
+apiVersion: v1
```
cbuto, line 3: in the long term, we should build a helm chart to wrap all these files. there’s currently no way for customers to configure apps that don’t use HelmReleases. 

I know this is necessarily customer facing but I’m thinking things like increasing cpu/memory, etc.


it might also make it easier to test without needing to clone this repo or install kommander to deploy it.

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
```diff
@@ -0,0 +1,52 @@
...
+kind: ConfigMap
+metadata:
+  name: ai-navigator-app-config
+  namespace: kommander
+data:
+  CHATBOT_URL: "https://chatbot.api.d2iq.com/v1/chat"
+  AUTHORIZER_URL: "https://auth.api.d2iq.com"
+  LICENSE_NAMESPACE: "kommander" # TODO: use kommander namspace, is this hardcoded ok?
```
cbuto, line 11: for the namespace, you can use substitution variables like `${releaseNamespace}`

i.e.
https://github.com/mesosphere/kommander-applications/blob/2853c7f030ad633ca23e4f6f02aefdda3cce1555/services/kommander/0.7.0/kommander.yaml#L6

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
```diff
@@ -0,0 +1,52 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: ai-navigator-app
+  template:
+    metadata:
+      labels:
+        app: ai-navigator-app
+    spec:
+      containers:
```
cbuto, line 30: > What do I need to do so that the service would be able to read licenses and the corresponding secret?

you will need a `Role`, `RoleBinding`, and a `ServiceAccount`. then you apply that service account to this deployment. 

This is a simple example: https://github.com/mesosphere/kommander-applications/blob/main/services/grafana-loki/0.69.16/grafana-loki-pre-install-jobs/pre-install.yaml

services/ai-navigator-app/metadata.yaml
```diff
@@ -0,0 +1,12 @@
+displayName: DKP AI Navigator
+description: DKP AI Navigator is D2iQ’s cool chatbot tool - TODO CHANGE
+category:
+  - tools
+type: platform
```
cbuto, line 5: should we make this `type: internal`? Or do we want it to show up in the UI?

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
cbuto, line 3: > Is there a way to apply the catalog from this branch on a cluster?

You could apply this app to the daily if you clone the git repo and just commit these files in the same location they are in this repo. happy to help with this if you want. 

or there’s a CLI flag you can use to point to another repo `—kommander-applications-repository` during install.

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
```diff
@@ -0,0 +1,52 @@
...
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: ai-navigator-app
+  template:
+    metadata:
+      labels:
+        app: ai-navigator-app
+    spec:
```
cbuto, line 29: there’s a few things we should set on this deployment like cpu/memory limits, securityContext (run as non-root, etc). but we can handle those later on.

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
mstruebing: > in the long term, we should build a helm chart to wrap all these files. there’s currently no way for customers to configure apps that don’t use HelmReleases.

@mikolajb argued that we should try to avoid a `HelmRelease` if possible.

services/ai-navigator-app/metadata.yaml
mstruebing: It should show up in the UI, yes, but it should always be deployed into the `kommander` namespace.
We will build a hack to let it only show up on the management cluster workspace in the first place and later build a proper solution as we might need this for other apps as well. This is a different namespace but is much more accurate than having it available in every workspace.

Does your suggestion regarding the `$(releaseNamespace}` is still valid in this case?

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
mstruebing: I assume by that you mean our cluster internal gitea instance? If so, I think I can figure out how to access this :)

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
mstruebing: You are right - is there a good way to find good limit/request values other than just guessing?

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
cbuto: yep exactly! once you commit these files you should see the `ClusterApp` show up

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
cbuto: the ideal way is to observe the utilization in a running cluster while the app is under some sort of load and estimate based on those values. we have some grafana dashboards builtin that you can use

services/ai-navigator-app/metadata.yaml
cbuto: yep the `$(releaseNamespace}` var will still work. Hardcoding the namespace will work as well but i'd recommend using the variable since there wouldn't be any cons to doing so

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
cbuto: I see, we can discuss further but using a helm chart has many pros like: long-term maintainability, testing, and configuration. it definitely require more work to get it setup with releases, etc but imo its worth it to make our lives easier in the future.

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
mstruebing: Done ✅

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
mstruebing: Done and tested ✅

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
mstruebing: I found this in another application definition:

```
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop:
            - ALL
          readOnlyRootFilesystem: true
          runAsNonRoot: true
          seccompProfile:
            type: RuntimeDefault
```

Are these the same properties we would need to set for `securityContext` for this application?

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
mikolajb: Yes, we can defiantly reevaluate this assumption once we have it running. It will probably be slightly faster to get MVP if we don't have the chart 🤔

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
```diff
@@ -0,0 +1,97 @@
...
+      serviceAccountName: ai-navigator-app
+      containers:
+        - name: ai-navigator-app
+          image: mesosphere/ai-navigator-app:development # TODO: Version
```
mikolajb, line 64: Do you want to land it once there is a version or earlier?

We could progress with even before the `ai-navigator-app` first release - in that case, let's just make it an empty customization (no deployments, etc).

services/ai-navigator-app/metadata.yaml
mstruebing: Thanks it works :)

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
mstruebing: I'm not sure what advantage we have when landing this without the resources?
Also, we still need to get the correct icon and description.

I would argue that we should create dev releases regularly so it's easy to test this service. 
What do you think?

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
cbuto: sounds good, we can sync up later on 👍

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
mstruebing: Security context added! :)

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
mstruebing: We do not want regular releases.

mstruebing:
@mikolajb @cbuto is this alright for now?
Limits and requests are currently still missing but I think it would be good to set these after we can have a look at some charts.

services/ai-navigator-app/metadata.yaml
```diff
@@ -1,8 +1,25 @@
 displayName: DKP AI Navigator
-description: DKP AI Navigator is D2iQ’s cool chatbot tool
+description: The AI Navigator is a state-of-the-art chatbot powered by advanced artificial intelligence, purpose-built for seamless Kubernetes cluster management and DKP environment optimization
```
nataliepina, line 2: ☝🏻 just need a period at the end of this sentence.

services/ai-navigator-app/0.1.0/ai-navigator-app.yaml
```diff
@@ -0,0 +1,106 @@
...
```
mikolajb, line 64:
```suggestion
          image: mesosphere/ai-navigator-app:v0.2.0
```

mikolajb (approved):

nataliepina (approved):

cbuto (approved):
looks good! it would be nice to have tickets for the follow up work just so we don't forget about them.

cbuto:
@mstruebing can we backport this to 2.6 as-is?

mstruebing:
@cbuto yes we can! :)

cbuto:
@mstruebing looks like we'll have to manually create the backport by cherry-picking the commit(s) 😞

mstruebing:
@cbuto it's here: https://github.com/mesosphere/kommander-applications/pull/1570

PR #1538, opened 2023-09-01
fix: disables telemetry transmission for Grafana

tillt:
Prevents Grafana from phoning home with telemetry data.

https://d2iq.atlassian.net/browse/D2IQ-99007

```release-note

```

**Checklist**

### DIFF
--- a/services/centralized-grafana/48.3.1/defaults/cm.yaml
+++ b/services/centralized-grafana/48.3.1/defaults/cm.yaml
@@ -74,6 +74,7 @@ data:
           default_home_dashboard_path: "/tmp/dashboards/global-overview.json"
         analytics:
           reporting_enabled: false
+          check_for_updates: false
 
       datasources:
         datasources.yaml:
--- a/services/centralized-grafana/48.3.1/defaults/cm.yaml
+++ b/services/centralized-grafana/48.3.1/defaults/cm.yaml
@@ -72,6 +72,8 @@ data:
           allow_loading_unsigned_plugins: "grafana-piechart-panel"
         dashboards:
           default_home_dashboard_path: "/tmp/dashboards/global-overview.json"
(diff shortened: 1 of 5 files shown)


palexster (approved):

gracedo (approved):

tillt:
@d2iq-mergebot backport release-2.6

services/centralized-grafana/48.3.1/defaults/cm.yaml
```diff
@@ -72,6 +72,8 @@ data:
           allow_loading_unsigned_plugins: "grafana-piechart-panel"
         dashboards:
           default_home_dashboard_path: "/tmp/dashboards/global-overview.json"
+        analytics:
+          reporting_enabled: false
```
gracedo, line 76: > check_for_updates
Set to false, disables checking for new versions of Grafana from Grafana’s GitHub repository. When enabled, the check for a new version runs every 10 minutes. It will notify, via the UI, when a new version is available. The check itself will not prompt any auto-updates of the Grafana software, nor will it send any sensitive information.

https://github.com/grafana/helm-charts/blob/bd618deb347d6e756d7964ec2a838b4f1110fd48/charts/grafana/values.yaml#L723-L730 grafana chart also defaults `check_for_updates: true` so we'll also need to explicitly disable it

tillt: I was considering doing that in another PR -- your call

gracedo: let's just include it in this one so we don't have to have 2 prs per backport!

tillt: Now in here as well @gracedo

PR #1544, opened 2023-09-04
chore: bump gitea to v9.3.0

Azhovan:
Apply required changes to install successfully Gitea on Kommander namespace. 

https://d2iq.atlassian.net/browse/D2IQ-98618

```release-note

```

**Checklist**


services/gitea/9.3.0/defaults/cm.yaml
```diff
@@ -46,9 +56,11 @@ data:
         periodSeconds: 10
         successThreshold: 1
         failureThreshold: 10
-    statefulset:
+    deployment:
       annotations:
         secret.reloader.stakater.com/reload: ${tlsCertificateSecret}
+      labels:
+        "deployment.kubernetes.io/pod-name": "gitea-0"
```
jimmidyson, line 63: These labels will now be added to every pod replica and the `-0` suffix isn't the clearest here (it was previously used to reference the `0` th index pod from the statefulset). 

Are there not labels that already exist on the pod from the gitea deployment spec (https://gitea.com/gitea/helm-chart/src/branch/main/templates/gitea/deployment.yaml#L40) that we could use?

Azhovan: Apparently  I can re-use one if the labels:  `app.kubernetes.io/name=gitea`  
```shell
➜  k get po -l app.kubernetes.io/name=gitea -n kommander --kubeconfig test1xxx.conf
NAME                     READY   STATUS    RESTARTS   AGE
gitea-646459b978-tcgjl   1/1     Running   0          4h40m

```

services/gitea/9.3.0/defaults/cm.yaml
```diff
@@ -46,9 +56,11 @@ data:
         periodSeconds: 10
         successThreshold: 1
         failureThreshold: 10
-    statefulset:
+    deployment:
```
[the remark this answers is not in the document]
Azhovan: In the previous version (v8.x), PVC was automatically managed by statefulset, while in the new version (v9.x), it is automatically handled by helmchart. It may make sense to add those properties for upgrading, but not for installation. This is because, during installation, no PVC is created automatically if `persistence.create=false`.

Another point that isn’t entirely clear is that, [according to the documentation](  https://docs.gitea.com/installation/upgrade-from-gitea#backup-for-downgrade), database compatibility is not assured during major upgrades.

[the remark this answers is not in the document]
cbuto: hmmm yeah sounds like we'll need to somehow handle setting those config values during upgrade but not during install 🤔.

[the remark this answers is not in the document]
Azhovan: I think we can manually create a PVC object, filling in its name value with the `data-gitea-0` that is the value in gitea `v8.x` and older versions.

```yaml
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name:  data-gitea-0
  namespace: kommander
  annotations:
    helm.sh/resource-policy: keep
spec:
  accessModes:
    - ReadWriteOnce
  volumeMode: Filesystem
  resources:
    requests:
      storage: 10Gi
```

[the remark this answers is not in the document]
cbuto: that would work. It would be nice to have new installations use a `RWX` PV but not required necessarily (we need to tackle it at some point (https://d2iq.atlassian.net/browse/D2IQ-88808)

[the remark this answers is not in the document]
Azhovan: I see, I would say lets tackle it in it's ticket! just in case to be safe :smile:

services/gitea/9.3.0/defaults/pvc.yaml
```diff
@@ -0,0 +1,20 @@
+---
+# In the previous versions (v8.x and prior), PVC was automatically managed by the Statefulset, while in the
+# new versions (v9.x and later), it is automatically handled by the Helm chart.
+# In order to remain backward compatible with previous installations, we manually create a PVC object,
+# filling in its name value with the data-gitea-0 that is the value in gitea v8.x and older versions.
+kind: PersistentVolumeClaim
```
cbuto, line 6: will flux be able to deploy this resource 🤔 ? does it complain that there is already a resource deployed with this name?

jimmidyson: Would need to set create pvc to false too I think https://gitea.com/gitea/helm-chart/src/branch/main/templates/gitea/pvc.yaml#L1

services/gitea/9.3.0/defaults/kustomization.yaml
```diff
@@ -1,4 +1,5 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
+  - pvc.yaml
```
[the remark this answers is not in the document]
jimmidyson: Can't we just retain the pvc when the statefulset is deleted and then just point the gitea deployment at that?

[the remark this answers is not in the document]
cbuto: i think the problem is we still need to create a PVC for fresh installs

services/gitea/9.3.0/defaults/pvc.yaml
Azhovan: PVC is only created for new installations. For upgrade processes, we keep the existing PVC.

```yaml
    persistence:
      enabled: true
      create: false
      mount: true
      claimName: "data-gitea-0"
```

Azhovan:
@cbuto @jimmidyson @gracedo 
is there any thing that should be done here?

services/gitea/9.3.0/defaults/cm.yaml
```diff
@@ -37,6 +37,17 @@ data:
         service:
           REQUIRE_SIGNIN_VIEW: false
           DISABLE_REGISTRATION: true
+        session:
```
cbuto, line 40: i was looking at a bundle after the upgrade and it looks like the default redis deployment is a statefulset with 6 replicas. I'd be a bit concerned about the resource usage increase (mainly for pre-prov), we should sync about this 🤔

services/gitea/9.3.0/pvc.yaml
```diff
@@ -0,0 +1,22 @@
...
+# To avoid breaking the compatibility with previous installations, we do not let Helm chart create PVC (see `persistence.create` in cm.yaml file)
+# and for new installations we create a PVC object using the name `data-gitea-0` that is the same name used in
+# gitea v8.x and older versions. This way, we can keep the existing PVC for upgrade process.
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: data-gitea-0
+  namespace: ${releaseNamespace}
+  annotations:
+    helm.sh/resource-policy: keep
```
cbuto, line 14: now that we maintain this PVC behind a Flux Kustomization that has `prune: true`, I was thinking it might be worth adding this annotation to prevent pruning (it might help us in the future)

`kustomize.toolkit.fluxcd.io/prune: disabled`

https://fluxcd.io/flux/components/kustomize/kustomizations/#kustomizetoolkitfluxcdioprune

Azhovan: very good idea!

services/gitea/9.3.0/defaults/cm.yaml
Azhovan: for now, I've changed it to 3, I'll also check it with the team

Azhovan: apparently it is not possible to install it with 3 nodes. this is logs from the gitea-redis-cluster pods when I tried to install it with 3 nodes.   
```shell
*** ERROR: Invalid configuration for cluster creation.
*** Redis Cluster requires at least 3 master nodes.
*** This is not possible with 3 nodes and 1 replicas per node.
*** At least 6 nodes are required.
```

Azhovan: using nodes=4 or 5 also resulted in redis deployment failure!

services/gitea/9.3.0/defaults/cm.yaml
```diff
@@ -17,13 +17,13 @@ data:
         traefik.ingress.kubernetes.io/router.tls: "true"
       hosts:
         - paths:
-          - path: /dkp/kommander/git
-            pathType: ImplementationSpecific
+            - path: /dkp/kommander/git
+              pathType: ImplementationSpecific
```
tillt, line 21: This looks like you intended too much

services/gitea/9.3.0/defaults/cm.yaml
cbuto: oh i see because they are master/replica sets, sounds good then.

cbuto (approved):
looks good + one more step towards HA 🎉

PR #1549, opened 2023-09-06
fix: create dev tag pipeline

palexster:
https://d2iq.atlassian.net/browse/D2IQ-99064
Successful run https://github.com/mesosphere/kommander-applications/actions/runs/6096139790

### DIFF
--- a/.github/workflows/create-dev-tag.yml
+++ b/.github/workflows/create-dev-tag.yml
@@ -9,17 +9,17 @@ on:
 env:
   GOWORK: off
   GOPRIVATE: github.com/mesosphere
-  DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
-  DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
-  GH_ACCESS_TOKEN: ${{ secrets.MESOSPHERECI_USER_TOKEN }}
+
+permissions:
+  contents: write
 
 jobs:
   create-dev-tag:
-    if: github.ref == 'refs/heads/main'
     runs-on:
     - self-hosted
     - small
     strategy:
+      fail-fast: false
       matrix:
         branch:
           - main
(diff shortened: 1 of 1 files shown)


cbuto (approved):

cwyl02 (approved):

.github/workflows/create-dev-tag.yml
```diff
@@ -39,8 +39,8 @@ jobs:
       - name: Configure git for private repo access
         run: git config --global url.https://$GH_ACCESS_TOKEN@github.com/.insteadOf https://github.com/
 
-      - name: Setup Go
-        uses: actions/setup-go@v4
+      - name: Setup asdf
+        uses: asdf-vm/actions/setup@v1
```
jimmidyson, line 43: Prefer `v2`

```suggestion
        uses: asdf-vm/actions/setup@v2
```

.github/workflows/create-dev-tag.yml
```diff
@@ -52,15 +52,25 @@ jobs:
...
+          git config user.name "ci-mergebot"
+          git tag -s -f -m "${{ env.TAG }}" ${{ env.TAG }}
 
       - name: Push tag
-        run: git push --force --tags origin ${{ env.TAG }}
+        run: |
+          ## Committer email does not match GPG private key email
+          ## So we need to set it expliclty here
+          git config --global user.email "ci-mergebot@d2iq.com"
+          git config --global user.name "ci-mergebot"
```
jimmidyson, line 72: Can't we use `git_config_global: true` in import gpg step?

jimmidyson: Or alternatively use `git_committer_name` and `git_committer_email` in the import gpg step?

.github/workflows/create-dev-tag.yml
```diff
@@ -52,15 +52,25 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.MESOSPHERECI_USER_TOKEN }}
 
       - name: Create tag
-        run: git tag -s -f -m "${{ env.TAG }}" ${{ env.TAG }}
+        run: |
+          ## Committer email does not match GPG private key email
+          ## So we need to set it expliclty here
+          git config user.email "ci-mergebot@d2iq.com"
+          git config user.name "ci-mergebot"
```
jimmidyson, line 64: Not sure why this needs to be done once here and again for global below? As mentioned though, I think we can just use the gpg import action to do this.

.github/workflows/create-dev-tag.yml
palexster: I did try this https://github.com/mesosphere/kommander-applications/actions/runs/6096111493/job/16540998959 but it was not signed for the same user/mail.

jimmidyson:
@palexster Just had a thought... this is just pushing to kommander-applications repo, right? In that case you don't need to set up GPG key, or use different token, you can just use workflow permissions. We do this in some of our projects that use release-please and just need to specify permissions such as in https://github.com/mesosphere/mindthegap/blob/main/.github/workflows/release-please-main.yaml#L11-L12:

```
permissions:
  contents: write
```

palexster:
> @palexster Just had a thought... this is just pushing to kommander-applications repo, right? In that case you don't need to set up GPG key, or use different token, you can just use workflow permissions. We do this in some of our projects that use release-please and just need to specify permissions such as in https://github.com/mesosphere/mindthegap/blob/main/.github/workflows/release-please-main.yaml#L11-L12:
> 
> ```
> permissions:
>   contents: write
> ```

Thanks, actually I think the git commit identity is working now (I think I had switched email/name the first time 🤦🏻 ), but I'll switch to that which seems simpler. 

However, I discovered that the `make repo.dev` is not working on branches != main 
https://github.com/mesosphere/kommander-applications/actions/runs/6109028667/job/16579197876
The `tail -1` piped command was making everything green, but it was actually not generating any version tag.

palexster:
Succesful run: https://github.com/mesosphere/kommander-applications/actions/runs/6110666848

jimmidyson:
@palexster Like I said before, you don't need the GPG step (do we really need to sign tags :sweat_smile: ?) or any git setup at all as you're just pushing a tag to the repo so using the `${{github.token}}` should work fine for both cloning and pushing the tag. The permissions is probably not necessary as the default permissions for the github token in this repo have read and write permissions already so you can definitely remove that regardless.

Up to you which way you go, but personally I would prefer to use the short-lived repo scoped token rather than the mesosphere ci token - but I really don't mind so do what you think is best.

palexster:
> @palexster Like I said before, you don't need the GPG step (do we really need to sign tags 😅 ?) or any git setup at all as you're just pushing a tag to the repo so using the `${{github.token}}` should work fine for both cloning and pushing the tag. The permissions is probably not necessary as the default permissions for the github token in this repo have read and write permissions already so you can definitely remove that regardless.
> 
> Up to you which way you go, but personally I would prefer to use the short-lived repo scoped token rather than the mesosphere ci token - but I really don't mind so do what you think is best.

Interesting perspective, thanks. I did not consider the difference between token scopes and durations. Moved to this approach in 81a62e88bfd81a710a8bf3fc1d2195775a8601f1 .

PR #1557, opened 2023-09-07
fix: fixes linter to stay in the release-branch 2.6

tillt:

```release-note

```

**Checklist**


jimmidyson (approved):

Azhovan (approved):

make/pre-commit.mk
```diff
@@ -11,4 +11,4 @@ ifeq ($(wildcard $(PRE_COMMIT_CONFIG_FILE)),)
 	$(error Cannot find pre-commit config file $(PRE_COMMIT_CONFIG_FILE). Specify the config file via PRE_COMMIT_CONFIG_FILE variable)
 endif
 	env SKIP=$(SKIP) pre-commit run -a --show-diff-on-failure --config $(PRE_COMMIT_CONFIG_FILE)
-	git fetch origin release-2.6 && gitlint --ignore-stdin --commits origin/release-2.6..HEAD
```
jimmidyson: Why is this not necessary? If we don't fetch the branch then it won't be available if we only do a shallow clone for example, as is the default in GHA

PR #1568, opened 2023-09-08
feat(koperator): Revert to using arbitrary chart version

gracedo:
After additional discussions, it looks like we can simplify things by avoiding the syncing of koperator chart version with dkp/kommander version (involves pre/postrelease jobs, and postrelease jobs right now are risky since those bumps are not reviewed prior to releasing). Instead, we can follow the "kcli way" of getting its version which is set at build time via goreleaser ldflags https://github.com/mesosphere/kommander-cli/blob/97eb1585529a8b3183eb306d1bcdf85186a6ad86/.goreleaser.yml#L22-L27. This is how the kcli will be triggering ugprade, by using this version to update `kommandercore` (and `workspace`) with, so we should do the same in the kommander operator (and other controllers). Instead of using the chart version and passing the version down through flags, use the version as set at build time.

https://d2iq.atlassian.net/browse/D2IQ-97955

```release-note

```

**Checklist**

### DIFF
--- a/common/kommander-operator/helmrelease.yaml
+++ b/common/kommander-operator/helmrelease.yaml
@@ -12,7 +12,7 @@ spec:
         kind: GitRepository
         name: management
         namespace: kommander-flux
-      version: 2.7.0-dev
+      version: 0.2.0
   interval: 15s
   install:
     crds: CreateReplace
--- a/common/kommander-operator/helmrelease.yaml
+++ b/common/kommander-operator/helmrelease.yaml
@@ -12,7 +12,7 @@ spec:
         kind: GitRepository
         name: management
         namespace: kommander-flux
-      version: 0.2.0
+      version: 0.3.0
   interval: 15s
   install:
     crds: CreateReplace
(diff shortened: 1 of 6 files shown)


common/kommander-operator/helmrelease.yaml
```diff
@@ -12,7 +12,7 @@ spec:
         kind: GitRepository
         name: management
         namespace: kommander-flux
-      version: 2.7.0-dev
+      version: 0.2.0
```
cbuto, line 15: should this be bumped?
```suggestion
      version: 0.3.0
```

gracedo: oops yes thank you!

gracedo: fixed in https://github.com/mesosphere/kommander-applications/pull/1568/commits/36f693de7d25ec2852de88ffa3d84db712cc74bc

cbuto (approved):

PR #1571, opened 2023-09-11
fix: disables loki telemetry

tillt:
Disables loki telemetry.

https://d2iq.atlassian.net/browse/D2IQ-99148

```release-note
disables loki telemetry
```

**Checklist**

### DIFF
--- a/services/grafana-loki/0.69.16/defaults/cm.yaml
+++ b/services/grafana-loki/0.69.16/defaults/cm.yaml
@@ -14,6 +14,8 @@ data:
         common:
           compactor_address: http://{{ include "loki.compactorFullname" . }}:3100
         auth_enabled: false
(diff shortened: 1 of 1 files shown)


msdolbey (approved):
👍 
Fix it Fix it

services/grafana-loki/0.69.16/defaults/cm.yaml
```diff
@@ -14,6 +14,8 @@ data:
         common:
           compactor_address: http://{{ include "loki.compactorFullname" . }}:3100
         auth_enabled: false
+        analytics:
```
cbuto, line 17: we will also need to set this in `project-grafana-loki`

cbuto (approved):
looks good

PR #1579, opened 2023-09-12
ci(test): Bump k8s to 1.27.6

gracedo:
update e2e tests with k8s 1.27

https://d2iq.atlassian.net/browse/D2IQ-98552

```release-note

```

**Checklist**

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -9,13 +9,13 @@ E2E_TIMEOUT       ?= 120m
 # flexible with our testing as well as testing against the same patch version as we deliver
 # by default with DKP.
 # See https://github.com/mesosphere/kind-docker-image-automation/ for the build repo.
-E2E_KINDEST_IMAGE ?= "ghcr.io/mesosphere/kind-node-ci:v1.26.6"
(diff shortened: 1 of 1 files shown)


gracedo:
I propose to close this PR and punt on upgrading k8s for this e2e test. We can wait for app specific testing, and till then, rely on our kommander and nightly test suites.

https://d2iq.atlassian.net/browse/D2IQ-98552?focusedCommentId=566427
> This needs to be revisited. The two “hacked together” e2e tests are copied directly from the e2e repo and are run in TC. We decided not to migrate them to GHA due to this being a public repo and the logs of these runs would be available to GH users.
> 
> Bumping k8s would cause these 2 e2e tests running in TC to fail. But I think since these are currently the only tests in k-apps, the k8s bump only affects these two tests anyways, so we might want to just leave it like this for now (running previous k8s version) and wait for k-apps specific testing EDI to bring in proper tests. This is probably acceptable for now because we will have kommander and nightly tests running on l8s that should catch any incompatible apps.

PR #1608, opened 2023-09-19
feat: patch kcore status to kommander ver pre upgrade

cwyl02:
pre 2.7 kommander core status.version is empty. This prevent 2.7 koperator from triggering the declarative upgrade logic.

https://d2iq.atlassian.net/browse/D2IQ-98960

```release-note

```

**Checklist**

### DIFF
--- a/charts/kommander-operator/templates/pre_upgrade_kommandercore_hook.yaml
+++ b/charts/kommander-operator/templates/pre_upgrade_kommandercore_hook.yaml
@@ -1,0 +1,86 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Chart.Name }}-pre-upgrade
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": before-hook-creation
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ .Chart.Name }}-pre-upgrade
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": before-hook-creation
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+  - kind: ServiceAccount
+    name: {{ .Chart.Name }}-pre-upgrade
+    namespace: {{ .Release.Namespace }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Chart.Name }}-pre-upgrade
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+data:
+  pre_process_kcore.sh: |-
+    #!/bin/bash
+    set -eo pipefail
+
+    # get the version of kommander before upgrade
+    KOMMANDER_VERSION_PREUPGRADE=$(kubectl get helmrelease -n kommander kommander -o jsonpath='{.status.lastAppliedRevision}')
+    cat <<EOF > kcore_patch.yaml
+    apiVersion: dkp.d2iq.io/v1alpha1
+    kind: KommanderCore
+    metadata:
+      name: kommander-core
+    status:
+      version: $KOMMANDER_VERSION_PREUPGRADE
+    EOF
+
+    kubectl patch kommandercore kommander-core --subresource='status' --type='merge' --patch-file kcore_patch.yaml
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Chart.Name }}-pre-upgrade
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "0"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+spec:
+  template:
+    spec:
+      serviceAccountName: {{ .Chart.Name }}-pre-upgrade
+      {{- if .Values.priorityClassName }}
+      priorityClassName: "{{ .Values.priorityClassName }}"
+      {{- end }}
+      containers:
+        - name: {{ .Chart.Name }}-pre-upgrade
+          image: "{{ .Values.kubetools.image.repository | default "mesosphere/kommander2-kubetools" }}:{{ .Values.kubetools.image.tag }}"
+          command: ["/bin/bash","-c"]
+          args: ["/bin/scripts/pre_process_kcore.sh"]
+          volumeMounts:
+            - name: script
+              mountPath: /bin/scripts
+      volumes:
+        - name: script
+          configMap:
+            defaultMode: 0770
+            name: {{ .Chart.Name }}-pre-upgrade
+      restartPolicy: OnFailure
+---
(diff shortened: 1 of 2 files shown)


cwyl02:
tested on 2.6.0 cluster by directly updating management git repo with this change and `kommandercore` 's status.version is updated to `v2.6.0` as expected.

charts/kommander-operator/templates/pre_upgrade_kommandercore_hook.yaml
```diff
@@ -0,0 +1,86 @@
...
+    cat <<EOF > kcore_patch.yaml
+    apiVersion: dkp.d2iq.io/v1alpha1
+    kind: KommanderCore
+    metadata:
+      name: kommander-core
+    status:
+      version: $KOMMANDER_VERSION_PREUPGRADE
+    EOF
+
+    kubectl patch kommandercore kommander-core --subresource='status' --type='merge' --patch-file kcore_patch.yaml
```
gracedo, line 54: should we check first that the status version is not already populated? I guess I'm thinking about if this script gets rerun after kommandercore has been upgraded, if it might overwrite it to the previous version again. Like with config updates for example if that triggers a HR "upgrade" action

charts/kommander-operator/templates/pre_upgrade_kommandercore_hook.yaml
```diff
@@ -0,0 +1,86 @@
+---
```
gracedo, line 1: let's add a comment and create a ticket to remove this job in 2.8.0

mikolajb:
Just a question: why empty `version` would trigger an upgrade?

charts/kommander-operator/templates/pre_upgrade_kommandercore_hook.yaml
cwyl02: That's a good point! But even with that I think a helm upgrade after kommander upgrade will not make it patch to the previous version cuz at that point kommander HR is already at the upgraded version, thus this patch will result in no changes.

charts/kommander-operator/templates/pre_upgrade_kommandercore_hook.yaml
cwyl02: done

cwyl02:
> Just a question: why empty `version` would trigger an upgrade?

this is what will happen in short:

1. pre-upgrade job patching `.status.version` to 2.6 to `KommanderCore`
2. post-install job patching `.spec.version` to 2.7 to `KommanderCore`
3. (k-operator helm upgrade finished) k-operator sees `.status.version` != `.spec.version` -> trigger upgrade

gracedo:
The flow is documented in the KEP also if more context is needed: https://github.com/mesosphere/dkp-platform/blob/master/keps/sig-api/20230111-installer-controller.md#kommandercore-upgrade it was decided to do it this way vs bumping api version and using conversion webhooks (nothing is finalized, but we wanted to avoid bumping the api version)

charts/kommander-operator/templates/pre_upgrade_kommandercore_hook.yaml
```diff
@@ -0,0 +1,87 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: {{ .Chart.Name }}-pre-upgrade
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": before-hook-creation
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
```
cbuto, line 23: can we create a `Role` for this instead of user cluster admin? I think it should work since we are only patching an object in the same namespace where this runs

charts/kommander-operator/templates/pre_upgrade_kommandercore_hook.yaml
```diff
@@ -0,0 +1,87 @@
...
+metadata:
+  name: {{ .Chart.Name }}-pre-upgrade
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": before-hook-creation
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
```
cbuto, line 14: once we move to a role, this can be a `RoleBinding`

charts/kommander-operator/templates/pre_upgrade_kommandercore_hook.yaml
gracedo: I see what you mean, that's true.. I guess I just don't feel so comfortable with this hook potentially writing that status after the first initial state, because it should really be solely controlled by the operator. There is a possibility that before upgrade succeeds in the operator, this script is rerun and prematurely updates the status to the current ("new") version, in which case it is possible that the operator mid-upgrade process think that upgrade is done and won't continue on? It's probably not very likely, but I think it should be easy to avoid this by adding a check here just to limit the updating of these resources to the operator besides this one-time script which is meant to just run once for existing clusters

charts/kommander-operator/templates/pre_upgrade_kommandercore_hook.yaml
cwyl02: I was following the pattern of post_install job. And I think we can also just add `hook-succeeded` to the`hook-delete-policy` to all these RBAC resources? Probably want to do that too for post install job

cbuto: oh actually KommanderCore is a cluster scoped resource so we do need a ClusterRole but ideally we could limit the role to only be able to patch kommandercore

charts/kommander-operator/templates/pre_upgrade_kommandercore_hook.yaml
cwyl02: done

charts/kommander-operator/templates/pre_upgrade_kommandercore_hook.yaml
cwyl02: Guess gonna address that in a separate PR, wdyt?

mikolajb:
> > Just a question: why empty `version` would trigger an upgrade?
> 
> this is what will happen in short:
> 
>     1. pre-upgrade job patching `.status.version` to 2.6 to `KommanderCore`
> 
>     2. post-install job patching `.spec.version` to 2.7 to `KommanderCore`
> 
>     3. (k-operator helm upgrade finished) k-operator sees `.status.version` != `.spec.version` -> trigger upgrade

Thanks @cwyl02! Let me verify if I understand it:
- `status.version` should be set by the operator, right? Why do we decide to set it here?
- `spec.version` should be set by a "client" - same question 🙂

gracedo:
> status.version should be set by the operator, right? Why do we decide to set it here?

This is only a one time job to run pre-upgrade for *existing* clusters undergoing an upgrade to dkp 2.7 (not for fresh 2.7 installs) because we must distinguish existing clusters from new clusters in order for the operator to know whether to trigger an upgrade or not. Without the initial patching of status, the operator would see these clusters as the same -- operator sees empty status = new cluster --> do not run upgrade.

>spec.version should be set by a "client" - same question 🙂

I'm not sure if we've decided yet whether to set `.spec.version` in the post-install hook, but this hook _creates_ the kommandercore resource for the first time, so creating it with the current version might make sense. We just want to make sure we don't accidentally overwrite the version again after the initial creation (so the field would only be updated by the client in the future). If we want the client to instead handle the creation of kommandercore as well, that would require some refactoring to remove the post-install hook from the koperator chart and to move it into an additional step in the kcli install workflow.

cwyl02:
@mikolajb sorry for the confusion, for some reason I mentioned that post-install job but that's actually not relevant here.

As Grace said, this pre-upgrade job is a one-time only thing to save an API version bump.

As for `.spec.version`, we haven't decided where exactly to set it, but setting in from k-cli makes sense to me.

mikolajb (approved):
Thanks for all the answers! We could probably consider setting it by a kommander operator but it LGTM! 🙂

cbuto (approved):

PR #1610, opened 2023-09-20
ci: trigger release workflow after creating dev tag

cbuto:
Releases aren't being triggered when we create tags. this option allows us to use the built in github token and just use a workflow dispatch to do the release. alternatively, using a PAT should also work. 

```release-note

```

**Checklist**

### DIFF
--- a/.github/workflows/create-dev-tag.yml
+++ b/.github/workflows/create-dev-tag.yml
@@ -12,6 +12,7 @@ env:
 
 permissions:
   contents: write
+  actions: write
 
 jobs:
   create-dev-tag:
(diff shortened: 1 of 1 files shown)


cwyl02 (approved):

jimmidyson:
I think you can get rid of most of the existing git config stuff in there btw

cbuto:
> I think you can get rid of most of the existing git config stuff in there btw

I tried getting rid of the git user config and it complained: https://github.com/mesosphere/kommander-applications/actions/runs/6253295944/job/16978375666

i was able to get rid of the private repo config step though

.github/workflows/release.yml
```diff
@@ -1,6 +1,7 @@
 name: Release
 
 on:
+  workflow_dispatch:
```
gracedo, line 4: can users outside of our org manually trigger jobs?

cbuto: from looking around, it looks like you need write access to the repo at least https://github.com/orgs/community/discussions/26053.

that's a good point though, should we only allow tags to trigger the release or allow manually triggering it?

jimmidyson: Manual triggers are great to retry stuff if there are failures.

palexster (approved):

cbuto:
requires https://github.com/mesosphere/kommander-applications/pull/1613 https://github.com/mesosphere/kommander-applications/pull/1614 https://github.com/mesosphere/kommander-applications/pull/1615

.github/workflows/release.yml
cbuto: wdyt @gracedo?

gracedo: Sg if limited to users with write access

gracedo (approved):

PR #1653, opened 2023-10-05
feat: provision kind clusters

Azhovan:
This PR provides the minimal required functionality for creating kind clusters, which are needed for application-specific testing.

```release-note

```

**Checklist**


apptests/kind/kind.go
```diff
@@ -0,0 +1,78 @@
...
+type Cluster struct {
+	provider           *cluster.Provider
+	kubeconfigFilePath string
+	name               string
+}
+
+const defaultClusterName = "kind"
+
+// CreateCluster creates a new kind cluster with the given name.
+func CreateCluster(ctx context.Context, name string) (*Cluster, error) {
```
palexster, line 21: Nit: Can we split the instance creation of the cluster and the actual cluster creation as we did on K-E2E. This has the advantage of validating the inputs before creating the cluster. It also has a more consistent set of methods (the object has the delete but not the create).

palexster (approved):

PR #1654, opened 2023-10-05
feat: add kubernetes client

Azhovan:
This PR adds the rest client, which is required to work with the test kind clusters

```release-note

```

**Checklist**


sonofgibs (approved):

apptests/client/client.go
```diff
@@ -0,0 +1,53 @@
...
+	clientset *kubernetes.Clientset
+}
+
+// NewClient creates a new kubernetes client.
+func NewClient(ctx context.Context, kubeconfigPath string) (*Client, error) {
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
```
mhrabovcin, line 25: Why is the context necessary here? I would remove it from the "constructor" function.

mhrabovcin (approved):

apptests/client/client.go
Azhovan: removed

mhrabovcin (approved):

PR #1655, opened 2023-10-05
fix: Exclude ai-navigator-app from the bundle

mikolajb:
Let's skip ai-navigator-app from the bundle.

https://d2iq.atlassian.net/browse/D2IQ-99372

```release-note

```

**Checklist**


palexster (approved):

cwyl02 (approved):

gracedo (approved):
We don't think non-airgapped users are using these bundles right? Juuust checking

PR #1663, opened 2023-10-10
feat: add kustomization

Azhovan:
This PR adds the Kustomization package, which is used to build the Kubernetes kustomizations in application specific testing.

```release-note

```

**Checklist**


apptests/kustomize/kustomize.go
```diff
@@ -0,0 +1,86 @@
...
+		resources:   resmap.New(),
+	}
+}
+
+// Build runs the kustomization process on the directory and substitutes the
+// environment variables in the resources.
+func (k *Kustomize) Build() error {
+	for k, v := range k.substitutes {
+		os.Setenv(k, v)
+	}
```
mhrabovcin, line 33: This function has a side effect of modifying the process environment where it runs, potentially being insecure as it doesn't revert the changes. I think we should add a `defer` call that will revert any possible changes.

Another thing is that this function cannot be called in parallel and so I think it should be lock protected.

jimmidyson: Use `envsubst.Eval` instead - do not update the environment like this.

apptests/kustomize/kustomize.go
```diff
@@ -0,0 +1,82 @@
...
+		if err != nil {
+			return err
+		}
+
+		// Convert the output string to a resource and append it to the resources map
+		res, err := newResourceFromString(out)
+		if err != nil {
+			return err
+		}
+		k.resources.Append(res)
```
mhrabovcin, line 58: Should this be reset when `Build` is called? If the `Build` function is called twice there will be the same resources twice.

mhrabovcin (approved):

jimmidyson (approved):

PR #1670, opened 2023-10-14
feat(koperator): Remove pre-upgrade job to set kcore version

gracedo:
I noticed that the koperator pre-upgrade job was failing due to insufficient perms `Error from server (Forbidden): kommandercores.dkp.d2iq.io "kommander-core" is forbidden: User "system:serviceaccount:kommander:kommander-operator-pre-upgrade" cannot get resource "kommandercores/status" in API group "dkp.d2iq.io" at the cluster scope`

However, I realized that since we will not be shipping the declarative upgrade feature in full in the next release, we don't need this pre-upgrade job at all to update existing kcore statuses with the current version. We can instead just let the operator treat it as it would a "fresh install" which is to update the status with the current version and exit without running the upgrade workflow steps that we have added so far.

https://d2iq.atlassian.net/browse/D2IQ-99516

```release-note

```

**Checklist**


cwyl02 (approved):

takirala (approved):
:tada:

PR #1683, opened 2023-10-18
chore: Dedupe images to reduce airgapped bundle size

bcm820:
[D2IQ-99209](https://d2iq.atlassian.net/browse/D2IQ-99209)

[D2IQ-99209](https://d2iq.atlassian.net/browse/D2IQ-99209)

```release-note

```

**Checklist**

[D2IQ-99209]: https://d2iq.atlassian.net/browse/D2IQ-99209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[D2IQ-99209]: https://d2iq.atlassian.net/browse/D2IQ-99209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


services/gitea/9.3.0/defaults/cm.yaml
```diff
@@ -84,7 +84,7 @@ data:
       primary:
         priorityClassName: "dkp-critical-priority"
       image:
-        tag: 11.16.0-debian-11-r9
+        tag: 15.2.0-debian-11-r21
```
takirala, line 87: What other entity is downloading this specific image? Also, can we just remove the override and use whatever is the upstream default (which seems to be `15.4.0-debian-11-r10` looking at the chart url)

bcm820: It's being installed by dkp-insights. I'll remove the override and see if I can update dkp-insights to use the same.

gracedo: since we downgraded gitea, I decided to just revert the postgres image overrides. we'll just let insights use their default image in case this downgraded postgres image isn't compatible

gracedo (approved):

takirala (approved):

PR #1690, opened 2023-10-23
chore: Adjust livenessprobe in gitea

mikolajb:

It seems we can shorted the time gitea becomes ready by tuning livenessProbe's initial delay.

https://d2iq.slack.com/archives/C03P4F14M7E/p1697803003706169?thread_ts=1697798961.547819&cid=C03P4F14M7E

```release-note

```

**Checklist**

### DIFF
--- a/services/gitea/9.3.0/defaults/cm.yaml
+++ b/services/gitea/9.3.0/defaults/cm.yaml
@@ -58,6 +58,10 @@ data:
         periodSeconds: 10
         successThreshold: 1
         failureThreshold: 10
+      # We use startupProbe unlike the default configuration,
+      # to shorten readiness time, set initialDelaySeconds to 5 seconds.
+      livenessProbe:
+        initialDelaySeconds: 5
     deployment:
       annotations:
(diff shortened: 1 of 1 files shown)


services/gitea/9.3.0/defaults/cm.yaml
```diff
@@ -58,6 +58,10 @@ data:
         periodSeconds: 10
         successThreshold: 1
         failureThreshold: 10
+      # We use startupProbe unlike the default configuration,
+      # to shorten readiness time, set initialDelaySeconds to 5 seconds.
+      livenessProbe:
+        initialDelaySeconds: 5
```
palexster, line 64: how long is the default value (curiosity)?

palexster (approved):

services/gitea/9.3.0/defaults/cm.yaml
tillt: gitea.livenessProbe.initialDelaySeconds | Initial delay before liveness probe is initiated | 200

tillt (approved):

PR #1692, opened 2023-10-23
fix: Set default timeouts for the high load

mikolajb:
We have low values for probe timeouts. We can increase these default values to avoid issues with larger setups.

https://d2iq.atlassian.net/browse/D2IQ-99405

```release-note

```

**Checklist**


tillt (approved):
lgtm

palexster (approved):

PR #1702, opened 2023-10-27
fix: make dkp-insights use "uninstall" upgrade remediation startegy

asekretenko:
The default `rollback` remediation strategy can (and often does) only make things worse in case of Insights
because after database migration has happened the older version of Insights will no longer work (we do not support database rollbacks).

However, we do test Insights for reinstall, and high availability is not expected from Insights at this point.
Hence, switching to `uninstall` here leads to more sane behavior on upgrade failures.

- This is part of the fix for https://d2iq.atlassian.net/browse/D2IQ-99624 (note that this does not fix the root cause, only makes it better visible).

**Checklist**


kaiwalyajoshi (approved):

takirala (approved):

sonofgibs (approved):

asekretenko:
Re-pushed with the commit signed

PR #1721, opened 2023-11-02
revert: Gitea Bump

ahmedElqutb:
revert `services/gitea`

The problems we currently see in our CI might strictly be connected to the Gitea bump we recently did. That bump may not bring us any advantages but only the issues we saw. Reverting might be a quick and risk-minimised solution.

https://d2iq.atlassian.net/browse/D2IQ-99546


tillt (approved):
lgtm

Azhovan (approved):

Azhovan:
@tillt this is not a bug, is it? we just don't know what is wrong in our setup

tillt:
Something does not work as intended, hence a problem that light be a bug

PR #1735, opened 2023-11-09
fix: xxclude ai-nav info agent from the bundle

msdolbey:
The ai-navigator-cluster-info-agent would be in the airgapped bundle, and it shouldn't be

No JIRA - I just thought of it now, but I basically just tried to lift Mikołaj's change here and add the agent to this exclude:
https://github.com/mesosphere/kommander-applications/pull/1655

To be honest - I am mostly just hopeful that this will work the way I want it to.

No change, this will just exclude it from the airgapped bundle (new feature, so no change to users)

**Checklist**


mikolajb (approved):
🤜🏻🤛🏻

PR #1760, opened 2023-12-06
fix: Pin grafana-logging to 8.5

gracedo:
Pin grafana-logging to 8.5.26 which is the latest functioning version of grafana that loads the kubernetes audit dashboard successfully. Using 9+ causes the dashboard to crash upon loading, and after fixing the mem to avoid the crash, only the audit logs panel shows data. In interest of time and release timeline, let's just use grafana 8.5 and continue to investigate why newer grafana versions are causing these issues.

https://d2iq.atlassian.net/browse/D2IQ-99725

image

```release-note

```

**Checklist**

### DIFF
--- a/licenses.d2iq.yaml
+++ b/licenses.d2iq.yaml
@@ -95,6 +95,12 @@ resources:
       - license_path: LICENSE
         ref: v${image_tag%-rootless}
         url: https://github.com/go-gitea/gitea
+  - container_image: docker.io/grafana/grafana:8.5.26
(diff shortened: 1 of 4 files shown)


msdolbey (approved):
👍 
We'd like to resolve this in a peaceful way

msdolbey (approved):

palexster (approved):

mikolajb (approved):

PR #1776, opened 2023-12-19
fix: bump chisel container image to distroless 1.9.1

mhrabovcin:

```release-note

```

**Checklist**


mhrabovcin:
We cannot run chisel in distroless container as our controllers assume presence of `sh` in the container image.
https://github.com/mesosphere/kubetunnel/blob/031451e039d8332a6d97c0afdf355836da781e21/pkg/controllers/tunnel_server_deployment_controller.go#L262

PR #1778, opened 2023-12-19
build: drop ai-navigator images from listing for bundle

takirala:

ai-navigator images need not be populated in the airgapped bundle as it does not work (yet) in airgapped environments.

https://d2iq.atlassian.net/browse/D2IQ-99753

```release-note

```

**Checklist**

### DIFF
--- a/hack/list-images.sh
+++ b/hack/list-images.sh
@@ -132,11 +132,12 @@ gojq --yaml-input --raw-output 'select(.kind | test("^(?:Deployment|Job|CronJob|
                                 ./services/velero/*/{pre,post}-install/* \
                                 >>"${IMAGES_FILE}"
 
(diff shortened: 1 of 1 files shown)


takirala (commented):
Merging this change will require us to strip off the images from `licenses.d2iq.yaml` file as well. I am not sure about the correctness of this fix.

If this is not acceptable for license scanning and what not (cc @mhrabovcin ), one other option would be move this fix to https://github.com/mesosphere/kommander/blob/main/make/release.mk#L30-L72 as well so that the images are excluded just before building.

gracedo (approved):

cwyl02 (approved):

cwyl02:
looks like need to fix licenses.d2iq.yaml as well

PR #1785, opened 2023-12-21
fix: bump kubecost sidecar image

palexster:

Let's bump the kiwigrid/sidecar image to the latest one to avoid critical CVEs in auxiliary libraries.

```release-note

```

**Checklist**

### DIFF
--- a/services/kube-prometheus-stack/48.3.1/defaults/cm.yaml
+++ b/services/kube-prometheus-stack/48.3.1/defaults/cm.yaml
@@ -351,6 +351,9 @@ data:
         path: /dkp/grafana
         pathType: ImplementationSpecific
       sidecar:
+        image:
+          repository: kiwigrid/k8s-sidecar
+          tag: 1.25.3  
         dashboards:
           enabled: true
           label: grafana_dashboard
--- a/services/kube-prometheus-stack/48.3.1/defaults/cm.yaml
+++ b/services/kube-prometheus-stack/48.3.1/defaults/cm.yaml
@@ -353,7 +353,7 @@ data:
       sidecar:
         image:
           repository: kiwigrid/k8s-sidecar
-          tag: 1.25.3  
+          tag: 1.25.3
         dashboards:
           enabled: true
           label: grafana_dashboard
--- a/services/kubecost/0.37.1/defaults/cm.yaml
+++ b/services/kubecost/0.37.1/defaults/cm.yaml
@@ -75,6 +75,10 @@ data:
           analytics:
             reporting_enabled: false
             check_for_updates: false
+        sidecar:
+          image:
+            repository: kiwigrid/k8s-sidecar
+            tag: 1.25.3
 
       kubecostProductConfigs:
         grafanaURL: "/dkp/kubecost/grafana"
--- a/licenses.d2iq.yaml
+++ b/licenses.d2iq.yaml
@@ -148,7 +148,7 @@ resources:
       - license_path: LICENSE.txt
         ref: ${image_tag}
         url: https://github.com/jimmidyson/configmap-reload
-  - container_image: docker.io/kiwigrid/k8s-sidecar:1.25.0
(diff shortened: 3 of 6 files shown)


services/kubecost/0.37.1/defaults/cm.yaml
```diff
@@ -75,6 +75,10 @@ data:
           analytics:
             reporting_enabled: false
             check_for_updates: false
+        sidecar:
```
mhrabovcin, line 78: I think you need to create a new service version when changing the default CM.

services/centralized-grafana/48.3.1/centralized-grafana.yaml
```diff
@@ -11,7 +11,7 @@ spec:
         kind: HelmRepository
         name: mesosphere.github.io-charts-staging
         namespace: kommander-flux
-      version: 48.3.1
+      version: 48.3.2
```
gracedo, line 14: not sure if this bump was intentional but the latest version is actually 48.3.3 https://github.com/mesosphere/charts/blob/6135b0df4858e916d6bd5705ca63a43c885153bf/staging/kube-prometheus-stack/Chart.yaml#L24

services/centralized-kubecost/0.37.1/release/release.yaml
```diff
@@ -11,7 +11,7 @@ spec:
         kind: HelmRepository
         name: mesosphere.github.io-charts-stable
         namespace: kommander-flux
-      version: 0.37.1
+      version: 0.37.2
```
gracedo, line 14: this helm chart version doesn't exist! and I think it's fine not to bump the app version, we just need it to be bumped between dkp versions if there are any changes, and we've already bumped this app from 2.6, so we should be good to keep it at the current app version even if we make config changes

services/kube-prometheus-stack/48.3.2/defaults/cm.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: kube-prometheus-stack-48.3.1-d2iq-defaults
+  name: kube-prometheus-stack-48.3.2-d2iq-defaults
```
gracedo, line 4: same comment, we don't need to bump the app version

services/kubecost/0.37.2/defaults/cm.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: kubecost-0.37.1-d2iq-defaults
+  name: kubecost-0.37.2-d2iq-defaults
```
gracedo, line 4: and here as well

gracedo (changes requested):
We don't need to bump the app versions since they have been bumped from 2.6

services/kubecost/0.37.2/defaults/cm.yaml
palexster: I thought this was required (reporting the comment), so I did it everywhere. https://github.com/mesosphere/kommander-applications/pull/1785#discussion_r1434016668

Glad to rollback if not needed.

services/centralized-grafana/48.3.1/centralized-grafana.yaml
palexster: no it was a mistake, thanks for your review.

gracedo (approved):
thank you!

services/kube-prometheus-stack/48.3.1/defaults/cm.yaml
```diff
@@ -351,6 +351,9 @@ data:
         path: /dkp/grafana
         pathType: ImplementationSpecific
       sidecar:
+        image:
+          repository: kiwigrid/k8s-sidecar
+          tag: 1.25.3
```
gracedo, line 356:
```suggestion
          tag: 1.25.3
```

PR #1787, opened 2023-12-21
fix: drop airgapped support for Insights flags that use `bitnami-shell` 

asekretenko:
This prevents the `bitnami-shell` image with hard to remove critical CVEs from being pulled into the airgapped bundle by the `dkp-insights` app.

The side effect of this change is that without this image `postgres.tls` and `postgres.volumePermissions` become unusable in air-gapped environments. Note that we neither use these flags by default, nor can they add any significant amount of security in the Insights use case.

A backport of this PR into 2.7 will hopefully close https://github.com/mesosphere/kommander/issues/4129

**Checklist**

### DIFF
--- a/services/dkp-insights/1.0.1/helmrelease/list-images-values.yaml
+++ b/services/dkp-insights/1.0.1/helmrelease/list-images-values.yaml
@@ -1,8 +1,11 @@
 postgresql:
   metrics:
     enabled: true
-  volumePermissions:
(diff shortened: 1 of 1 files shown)


gracedo (approved):

kaiwalyajoshi (approved):

asekretenko:
re-pushed with a signed commit

PR #1804, opened 2024-01-02
fix: bump kubectl image to v1.27.9

palexster:

```release-note

```

**Checklist**

### DIFF
--- a/licenses.d2iq.yaml
+++ b/licenses.d2iq.yaml
@@ -45,6 +45,7 @@ ignore:
   - docker.io/bitnami/kubectl:1.22.0
   - docker.io/bitnami/kubectl:1.24.1
   - docker.io/bitnami/kubectl:1.26.4
+  - docker.io/bitnami/kubectl:1.27.9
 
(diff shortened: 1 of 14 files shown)


licenses.d2iq.yaml
```diff
@@ -45,6 +45,7 @@ ignore:
   - docker.io/bitnami/kubectl:1.22.0
   - docker.io/bitnami/kubectl:1.24.1
   - docker.io/bitnami/kubectl:1.26.4
+  - docker.io/bitnami/kubectl:1.28.5
```
gracedo, line 48: is this being bumped due to CVEs? 2.7 is on 1.27 so I wonder if we should use 1.27 kubectl images instead of 1.28 🤔

palexster: Changed in 2303943c5a55ed173035c31b3e5e96999058f5f8

gracedo (approved):
thank you!!

we should go back (later) and bump any kubectl images in the various charts we host in m/charts also (defaults or otherwise)

PR #1824, opened 2024-01-10
fix: template kubectl images in a release

palexster:
This PR lets all the kubectl images explicitly set in k-apps to be configurable using dedicated variables.


mhrabovcin (approved):
🚀

PR #1843, opened 2024-01-17
ci: Switch to devbox

mikolajb:
Let's switch to devbox (from asdf)

```release-note

```

**Checklist**


.github/workflows/create-dev-tag.yml
```diff
@@ -35,8 +35,14 @@ jobs:
...
-      - name: Setup asdf
-        uses: asdf-vm/actions/setup@v2
+      - name: Install NIX
+        uses: cachix/install-nix-action@v24
+
+      - name: Install devbox
+        uses: jetpack-io/devbox-install-action@v0.7.0
+        with:
+          enable-cache: true
+          skip-nix-installation: true
```
jimmidyson, line 45: Why do this?

mikolajb: Why to skip `nix` installation? I think there is an issue with installing it on our self hosted runners. The other step `cachix/install-nix-action` works fine.

make/go.mk
```diff
@@ -1,12 +1,11 @@
 .PHONY: go-test
 go-test: go-lint
-go-test: install-tool.golang
-	cd hack/release && go test -v -race -covermode=atomic -coverprofile=coverage.out ./...
+	devbox run -- "cd hack/release && go test -v -race -covermode=atomic -coverprofile=coverage.out ./..."
 
 .PHONY: mod-tidy
 mod-tidy: install-tool.golang
-	cd hack/release && go mod tidy
+	devbox run -- "cd hack/release && go mod tidy"
```
jimmidyson, line 7: No need for devbox run prefix in any makefiles

make/flux.mk
```diff
@@ -1,5 +1,5 @@
 .PHONY: flux-update
 flux-update: ## Updates flux manifests
-flux-update: install-tool.flux2 install-tool.kustomize install-tool.github-cli install-tool.yq
+flux-update:
```
jimmidyson, line 3: No need for these lines without dependencies, they're refundable y

.github/workflows/validate-licenses.yaml
```diff
@@ -19,12 +19,20 @@ jobs:
...
+        uses: jetpack-io/devbox-install-action@v0.7.0
+        with:
+          enable-cache: true
+          skip-nix-installation: true
+
       - name: Generate image list
         run: |
           export PATH=${PWD}/.local/bin:$PATH
-          make install-tool.helm install-tool.go.envsubst install-tool.go.gojq
+          make install-tool.go.envsubst install-tool.go.gojq
```
jimmidyson, line 35: These can also be installed via devbox btw, although `envsubst` is a different project it provides the same usage.

devbox.json
```diff
@@ -0,0 +1,26 @@
...
+    "github-cli": "latest",
+    "kustomize": "latest",
+    "pre-commit": "latest",
+    "golangci-lint": "latest",
+    "direnv": "latest",
+    "kubernetes-helm": "latest",
+    "yq-go": "latest",
+    "go": "1.20",
+    "docker": "latest",
+    "gitlint": "latest"
```
jimmidyson, line 14: nit: Perhaps we can keep these sorted?

jimmidyson, line 14: Let's add gojq and envsubst in here
```suggestion
    "gitlint": "latest",
    "gojq": "latest",
    "envsubst": "latest"
```

hack/flux/update-flux.sh
```diff
@@ -34,11 +39,11 @@ function update_flux() {
     flux install -n kommander-flux --export > templates/flux.yaml
     cp "$REPO_ROOT/hack/flux/flux-update-kustomization.yaml" templates/kustomization.yaml
     cp "$REPO_ROOT"/hack/flux/templates/* templates/
-    kustomize build --output templates templates
+    devbox run kustomize build --output templates templates
```
jimmidyson, line 42: We shouldn't need `devbox run` in here should we?

make/all.mk
```diff
@@ -4,7 +4,6 @@ include $(INCLUDE_DIR)make.mk
 include $(INCLUDE_DIR)shell.mk
 include $(INCLUDE_DIR)help.mk
 include $(INCLUDE_DIR)repo.mk
-include $(INCLUDE_DIR)ci.mk
 include $(INCLUDE_DIR)docker.mk
 include $(INCLUDE_DIR)flux.mk
 include $(INCLUDE_DIR)tools.mk
```
jimmidyson, line 9: I think we should be able to replace all tooling with devbox, we can provide custom flakes for any tools that don't have binary builds available from nix packages - see for example https://github.com/mesosphere/mindthegap/blob/main/hack/flakes/flake.nix#L15-L29

mikolajb: The only thing that we have now is `gh-dkp`, it is installed by `gh` itsef and always in the latest version. Wdyt about keeping it as it is?

mhrabovcin (approved):

mhrabovcin:
🚀

PR #1869, opened 2024-01-19
fix(kommander): Add ai-nav apps to list of mgmt apps

gracedo:
This ensures that the apps are handled during expansion to be removed.

ai navigator apps were sticking around post-expansion. in order to ensure that they are removed instead, we must add them to the list of management apps.

https://d2iq.atlassian.net/browse/D2IQ-99868

```release-note

```

**Checklist**


msdolbey (commented):
So far - this is 
preapproved

lukeogg (approved):
LGTM

msdolbey (approved):
👍 
This is the real deal

PR #1894, opened 2024-01-25
feat: Update name of HR

lukeogg:

The HR had a different name from the AP name. There is a requirement in Kommander that this name match for charts that will get the override. This fixes the inability to add config overrides in the UI, etc. 

https://d2iq.atlassian.net/browse/D2IQ-99867

- Tested in a local cluster against kapps branch.
- Check for correct mapping of config maps 

```release-note

```

**Checklist**

### DIFF
--- a/services/ai-navigator-app/0.2.0/helmrelease.yaml
+++ b/services/ai-navigator-app/0.2.0/helmrelease.yaml
@@ -1,7 +1,7 @@
 apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
 kind: Kustomization
 metadata:
-  name: ai-navigator-app
+  name: ai-navigator-app-helmrelease
   namespace: ${releaseNamespace}
 spec:
   force: false
--- a/services/ai-navigator-app/0.2.0/helmrelease.yaml
+++ b/services/ai-navigator-app/0.2.0/helmrelease.yaml
@@ -1,7 +1,7 @@
 apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
 kind: Kustomization
 metadata:
-  name: ai-navigator-cluster-info-api-helmrelease
+  name: ai-navigator-app
   namespace: ${releaseNamespace}
 spec:
   force: false
--- a/services/ai-navigator-app/0.2.0/helmrelease/helmrelease.yaml
+++ b/services/ai-navigator-app/0.2.0/helmrelease/helmrelease.yaml
@@ -1,7 +1,7 @@
 apiVersion: helm.toolkit.fluxcd.io/v2beta1
 kind: HelmRelease
 metadata:
-  name: ai-navigator-cluster-info-api
+  name: ai-navigator-app
   namespace: ${releaseNamespace}
 spec:
   chart:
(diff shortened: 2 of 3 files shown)


cwyl02:
Should we also change the name of the [default CM](https://github.com/mesosphere/kommander-applications/blob/main/services/ai-navigator-app/0.2.0/defaults/cm.yaml) for this app?

gracedo (approved):

gracedo (commented):
Can we rename it in a few other places too? like:

https://github.com/mesosphere/kommander-applications/blob/7e96d6bbfc19eb33511972a17dceaed6c66b624e/services/ai-navigator-app/0.2.0/defaults/cm.yaml#L4

https://github.com/mesosphere/kommander-applications/blob/7e96d6bbfc19eb33511972a17dceaed6c66b624e/services/ai-navigator-app/0.2.0/helmrelease.yaml#L4

services/ai-navigator-app/0.2.0/helmrelease.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
 kind: Kustomization
 metadata:
-  name: ai-navigator-cluster-info-api-helmrelease
+  name: ai-navigator-app
```
gracedo, line 4:
```suggestion
  name: ai-navigator-app-helmrelease
```

services/ai-navigator-app/0.2.0/helmrelease/helmrelease.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: helm.toolkit.fluxcd.io/v2beta1
 kind: HelmRelease
 metadata:
-  name: ai-navigator-cluster-info-api
+  name: ai-navigator-app
```
gracedo, line 4:
```suggestion
  name: ai-navigator-app-helmrelease
```

gracedo (approved):
just adding `-helmrelease` to the hr name to be super clear!

gracedo (changes requested):
```
2024-02-06 20:55:33 ERR K8s v1.29.0 | ai-navigator-app 0.2.0 | FluxKustomization: ai-navigator-app |     err="*helm.Parser: no node with this ID: {{"v1" "ConfigMap"} {"ai-navigator-cluster-info-api-0.2.0-d2iq-defaults" "kommander"}}"
```

We'll have to update the `valuesFrom` cm name on the HR spec as well

services/ai-navigator-app/0.2.0/helmrelease/helmrelease.yaml
lukeogg: Will this work? I thought this was the one that had to match the app name in order for the CM to apply correctly.

gracedo: oops sorry you're right! just the flux kust name not this one 🤦🏻‍♀️

gracedo (approved):

PR #1910, opened 2024-02-02
ci(gha): Fix labeler workflow after tooling bump

gracedo:
labeler workflow bump contained breaking changes in the labeler config file

```release-note

```

**Checklist**


msdolbey (approved):
👍 
oh my glob

PR #1941, opened 2024-02-16
fix: unpin KPS Grafana, update metrics path

alembiewski:
This PR fixes the metrics scraping issue with Grafana v10+. Also, it bumps Grafana in KPS to the latest version (v10.3.3)

Screenshot 2024-02-16 at 12 37 52

Screenshot 2024-02-16 at 12 37 17

https://d2iq.atlassian.net/browse/D2IQ-99819

```release-note

```

**Checklist**

### DIFF
--- a/services/kube-prometheus-stack/48.3.1/defaults/cm.yaml
+++ b/services/kube-prometheus-stack/48.3.1/defaults/cm.yaml
@@ -341,6 +341,7 @@ data:
       serviceMonitor:
         labels:
           prometheus.kommander.d2iq.io/select: "true"
+        path: "/dkp/grafana/metrics"
       ingress:
         enabled: true
         annotations:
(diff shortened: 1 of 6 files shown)


services/kube-prometheus-stack/48.3.1/defaults/cm.yaml
```diff
@@ -334,15 +334,14 @@ data:
             memory: 200Mi
     grafana:
       enabled: true
-      # pinning grafana to 9.5 since it appears that upgrading to v10 is causing issues with prometheus scraping
-      # grafana metrics
       image:
-        tag: 9.5.13
+        tag: 10.3.3
```
gracedo, line 338: can we just remove this config and let the default (with this version of KPS) be used?

gracedo (commented):
🙏🏻 thank you!! just one comment

services/kube-prometheus-stack/48.3.1/defaults/cm.yaml
alembiewski: done. also verified metrics scraping works in the default version (10.0.3)

PR #1950, opened 2024-02-19
fix: add a workaround to make the 5.x Velero chart adopt 4.x hook resources

asekretenko:
- Closes https://d2iq.atlassian.net/browse/D2IQ-99927
- Closes https://d2iq.atlassian.net/browse/D2IQ-99936

```release-note
NONE

```

**Checklist**

### DIFF
--- a/services/velero/5.2.2/helmrelease/helmrelease.yaml
+++ b/services/velero/5.2.2/helmrelease/helmrelease.yaml
@@ -21,8 +21,6 @@ spec:
       retries: 30
   upgrade:
     crds: CreateReplace
-    remediation:
-      retries: 30
   releaseName: velero
   valuesFrom:
     - kind: ConfigMap
--- a/services/velero/5.2.2/helmrelease/helmrelease.yaml
+++ b/services/velero/5.2.2/helmrelease/helmrelease.yaml
@@ -3,6 +3,8 @@ kind: HelmRelease
 metadata:
   name: velero
   namespace: ${releaseNamespace}
+  annotations:
+    velero.kommander.d2iq.io/no-hooks-chart: "true"
 spec:
   chart:
     spec:
--- a/services/velero/5.2.2/v4-hooks-adoption/v4-hooks-adoption.yaml
+++ b/services/velero/5.2.2/v4-hooks-adoption/v4-hooks-adoption.yaml
@@ -1,0 +1,89 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: velero-v4-hooks-adoption
+  namespace: ${releaseNamespace}
+---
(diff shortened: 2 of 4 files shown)


services/velero/5.2.2/helmrelease/helmrelease.yaml
```diff
@@ -3,6 +3,8 @@ kind: HelmRelease
 metadata:
   name: velero
   namespace: ${releaseNamespace}
+  annotations:
+    velero.kommander.d2iq.io/no-hooks-chart: "true"
```
gracedo, line 7: clever!

services/velero/5.2.2/v4-hooks-adoption/v4-hooks-adoption.yaml
```diff
@@ -0,0 +1,89 @@
...
+                sleep 30
+              done
+
+              # Hook resources will be identified by matching the labels values set by 4.x charts,
+              # with the presence of some `helm.sh/chart` value used as an additional guard against
+              # users setting these values of `app.kubernetes.io` labels on their own resources.
+              #
+              # In case a newer chart (which does not use hooks to install these) is already present, this is a no-op.
+
+              echo "Labeling and annotating resources installed by a 4.x Velero Helm chart hooks to allow adoption by the new chart"
```
gracedo, line 80: I assume it is probably safe to always run this again after it's been run / on a fresh install of 2.x chart? or do we want to add a safeguard to only run this once, like is there another defining label or annotation on the 4.x resource we can use to select against to avoid running it on a 5.x chart?

Either way, let's also create a follow up JIRA to handle cleaning up this job in the next release! The ticket can be created in https://d2iq.atlassian.net/browse/D2IQ-95564

gracedo (approved):

cwyl02 (approved):

services/velero/5.2.2/v4-hooks-adoption/v4-hooks-adoption.yaml
asekretenko: It is safe to run it against 5.x (and this is what can easily happen during a fresh install), as these annotations and label are identical to those set by Helm on chart-owned resources.

However, we indeed need to remove this job in the next release to decrease the maintenance burden. Filed https://d2iq.atlassian.net/browse/D2IQ-100067

services/velero/5.2.2/helmrelease/helmrelease.yaml
asekretenko: Initially I intended to watch for the proper chart version in the HelmRelease, but realized that this would be very prone to breakage in the face of patch releases. We would have either needed one more location where the version needs to be bumped, or in a somewhat non-trivial propagation of said version.

PR #1951, opened 2024-02-20
fix(kubectl): Bump all kubectl image defaults to 1.29.2

gracedo:
Bump kubectl image in values overrides to 1.29.2.

https://d2iq.atlassian.net/browse/D2IQ-99985

kubetools templating variables were added back recently, but I wanted to note that this variable is not actually getting substituted, so effectively we are still using the defaults and thus must still manually bump them here. we should revisit switching to using kubetools and getting it substituted when we get to implementing declarative installs (relevant jira ticket https://d2iq.atlassian.net/browse/D2IQ-87946)

```release-note

```

**Checklist**

### DIFF
--- a/licenses.d2iq.yaml
+++ b/licenses.d2iq.yaml
@@ -46,6 +46,7 @@ ignore:
   - docker.io/bitnami/kubectl:1.24.1
   - docker.io/bitnami/kubectl:1.26.4
   - docker.io/bitnami/kubectl:1.27.9
+  - docker.io/bitnami/kubectl:1.29.2
 
   # Fossa cannot scan C/CPP projects
   # See: https://docs.fossa.com/docs/c-cpp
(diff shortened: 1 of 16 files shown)


takirala (approved):

licenses.d2iq.yaml
```diff
@@ -46,6 +46,7 @@ ignore:
   - docker.io/bitnami/kubectl:1.24.1
   - docker.io/bitnami/kubectl:1.26.4
   - docker.io/bitnami/kubectl:1.27.9
+  - docker.io/bitnami/kubectl:1.29.2
```
takirala, line 49: for my own edification, do you know why we ignore these ?

msdolbey (approved):
👍 
Bump

licenses.d2iq.yaml
gracedo:
```  # Incorrect mapping was provided
  # - container_image: docker.io/bitnami/kubectl:1.19.7
  #   sources:
  #     - url: https://github.com/bitnami/bitnami-docker-kubectl
  #       ref: ${image_tag}-debian-10-r31
  #       license_path: LICENSE```

might explain it in this block? (but I'm not really sure either)

PR #1958, opened 2024-02-21
revert: "feat: Rename deprecated secretRef to certSecretRef (#1925)"

takirala:
Reverts mesosphere/kommander-applications#1925

`secretRef` to `certSecretRef` is a bigger change than we initially anticipated. when we reintroduce this feature, we should:

- have a patch to remove the `secretRef` from HelmRepository resources.
- update the `caFile` to `ca.crt` in the secret data.


gracedo (approved):

msdolbey (approved):
👍 
roll it on back

PR #1969, opened 2024-02-23
chore: migrate HelmRepository from `secretRef` to `certSecretRef` 

alembiewski:
This PR updates the existing patch to add certSecretRef to HelmRepository manifests and also introduced a `delete` patch to remove deprecated `secretRef` references.

https://d2iq.atlassian.net/browse/D2IQ-99228

```release-note

```

**Checklist**


mhrabovcin (approved):

common/airgapped/patch-add-cert-secret-ref.yaml
mikolajb, line 1: Do we need to clean it up from the existing management repository?

mikolajb (approved):

common/airgapped/patch-delete-secret-ref.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: source.toolkit.fluxcd.io/v1beta2
+kind: HelmRepository
+metadata:
+  name: not-used-in-a-patch
+spec:
+  secretRef:
+    $patch: delete
```
alembiewski, line 7: Looks like `$patch: delete` instruction breaks the `Check licenses.d2iq.yaml` check: https://github.com/mesosphere/kommander-applications/actions/runs/8018555753/job/21904680915?pr=1969

PR #2008, opened 2024-03-11
fix(ceph): Downgrade to 1.12.11

gracedo:

We are hitting https://github.com/rook/rook/issues/13614 in our preprov e2e tests -- let's downgrade back to 1.12 as we continue to investigate.

**Checklist**


msdolbey (approved):
👍 
Have to

PR #2009, opened 2024-03-11
chore: updating ui to most recent tag

msdolbey:
Just updating UI to the latest

n/a

Solely latest UI bump

Solely latest UI bump

**Checklist**


gracedo (approved):
if I find some time I can take a look at why the automation isn't working in kommander-ui

nataliepina:
I was going to say this should be automated like this one https://github.com/mesosphere/kommander-applications/pull/1766
I am not sure why it would not be working.

PR #2014, opened 2024-03-13
ci(gha): Update backport action repo

gracedo:
fix backport action

```release-note

```

**Checklist**

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -18,7 +18,7 @@ jobs:
       )
     steps:
       - name: Import GPG key
-        uses: crazy-max/ghaction-import-gpg@v4
+        uses: crazy-max/ghaction-import-gpg@v6
         with:
           gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
           git_tag_gpgsign: true
           git_config_global: true
       - name: Backport Action
-        uses: sqren/backport-github-action@v8.9.3
+        uses: sorenlouv/backport-github-action@v9.3.1
```
nataliepina, line 30: Looks like this can be [9.4.4](https://github.com/sorenlouv/backport/releases/tag/v9.4.4)

msdolbey (approved):
I'm good with either 9.4.4 or 9.3.1 - so take my 👍 
I would like to see some action

nataliepina (approved):

.github/workflows/backport.yaml
nataliepina: Update, don't use this version I guess? I saw some failures saying it can't find this version.

PR #2020, opened 2024-03-14
feat: updates nvidia values to use all latest and pins toolkit

faiq:
This updates the nvidia gpu operator chart to use newer values for the component versions.

```release-note

```

**Checklist**


mhrabovcin (approved):

mikolajb (approved):

msdolbey (approved):
👍 
Latest and greatest

PR #2044, opened 2024-03-28
feat: add an app for git-operator

asekretenko:
Adds an app for the git-operator to facilitate development of kommander-operator.

- https://d2iq.atlassian.net/browse/D2IQ-100236

```release-note
NONE
```

**Checklist**


mikolajb (approved):
Amazing! Let's create `release-2.8` before this commit so it doesn't end in 2.8 🙏🏻

PR #2052, opened 2024-04-01
fix(grafana-logging): Fix OOM issue after upgrade to v9+

johnliu25:
Fix the Grafana logging out of memory issue after upgrade to v9+
The Grafana get OOM killed status and the pod is in the CrachBackLoopoff State which causing the 502 and 404 issue. 
Screenshot 2024-04-01 at 2 25 21 PM

```
zihui.liu@C02F94FBML85 Desktop % k get pods -A | grep grafana-logging
kommander                           grafana-logging-68cf97499b-qvjvg                                    1/2     OOMKilled   8 (6m58s ago)   40m
```
After update the memory from 100Mi to 200Mi the issue has been solved. 
https://github.com/mesosphere/kommander-applications/blob/main/services/grafana-logging/6.60.2/defaults/cm.yaml#L75 
https://github.com/mesosphere/kommander-applications/blob/main/services/grafana-logging/6.60.2/defaults/cm.yaml#L78 
Screenshot 2024-04-01 at 2 25 49 PM

### DIFF
--- a/services/grafana-logging/6.60.2/defaults/cm.yaml
+++ b/services/grafana-logging/6.60.2/defaults/cm.yaml
@@ -72,10 +72,10 @@ data:
       # keep request = limit to keep this container in guaranteed class
       limits:
         cpu: 300m
-        memory: 100Mi
+        memory: 200Mi
       requests:
         cpu: 200m
-        memory: 100Mi
+        memory: 200Mi
 
     readinessProbe:
       httpGet:
--- a/services/project-grafana-logging/6.60.2/defaults/cm.yaml
+++ b/services/project-grafana-logging/6.60.2/defaults/cm.yaml
@@ -69,10 +69,10 @@ data:
       # keep request = limit to keep this container in guaranteed class
       limits:
         cpu: 300m
-        memory: 100Mi
+        memory: 200Mi
       requests:
         cpu: 200m
-        memory: 100Mi
+        memory: 200Mi
 
     readinessProbe:
       httpGet:


gracedo (commented):
Can you remove the version override https://github.com/mesosphere/kommander-applications/blob/main/services/grafana-logging/6.60.2/defaults/cm.yaml#L10-L12 to use to use the default version deployed now, with this fix?

And can you make the same changes to https://github.com/mesosphere/kommander-applications/blob/main/services/project-grafana-logging/6.60.2/defaults/cm.yaml#L10-L12 which is the project scoped version of the same app?

Thanks!

johnliu25:
@gracedo sure. will do.

johnliu25:
@gracedo Currently, I still need to figure out how to fix the API call issue. Otherwise if we remove the version override, the grafana will not working properly.
The grafana API changed from 
```
/dkp/grafana/api/datasources/proxy/*
```
to 
```
/dkp/grafana/api/ds/query
```
since the /dkp/grafana/api/datasources/proxy/* has been deprecated from Grafana v9. 
The response of /dkp/grafana/api/ds/query could not be consumed by grafana logging UI.

cwyl02 (approved):

gracedo:
> Fix the Grafana logging out of memory issue after upgrade to v9+
So this OOM issue only occurs in v9+, but there are also other issues upgrading to that version due to deprecated api? Do we need this memory bump then, if we aren't actually bumping to v9+ yet?

PR #2053, opened 2024-04-02
ci: GHA for app specific test

cwyl02:
1st iteration of GHA for app specific test that handles install.

```release-note

```

**Checklist**

### DIFF
--- a/.github/workflows/application-test-scenario.yaml
+++ b/.github/workflows/application-test-scenario.yaml
@@ -1,0 +1,47 @@
+name: Application Specific Test Scenario
+on:
+  workflow_dispatch:
+    inputs:
+      scenario:
+        description: 'The scenario to run, e.g., install or upgrade'
+        required: true
+        default: 'install'
+        type: string
+      app:
+        description: 'The kommander application to test, e.g. reloader,kube-prometheus-stack'
+        required: true
+        type: string
+  workflow_call:
+    inputs:
+      scenario:
+        description: 'The scenario to run, e.g., install or upgrade'
+        required: true
+        type: string
+      app:
+        description: 'The kommander application to test, e.g. reloader,kube-prometheus-stack'
+        required: true
+        type: string
+
+jobs:
+  run-test-scenario:
+    runs-on:
+      - ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Set up Go
+      uses: actions/setup-go@v4
+      with:
+        go-version-file: 'apptests/go.mod'
+
+    - name: Install Ginkgo
+      working-directory: apptests
+      run: go install github.com/onsi/ginkgo/v2/ginkgo
+
+    - name: Run ${{ github.event.inputs.app }} ${{ github.event.inputs.scenario }} Test
+      working-directory: apptests
+      run: |
+        ginkgo --label-filter="${{ inputs.scenario }} && ${{ inputs.app }}" appscenarios
+
--- a/.github/workflows/application-test-scenario.yaml
+++ b/.github/workflows/application-test-scenario.yaml
@@ -22,6 +22,10 @@ on:
         required: true
         type: string
 
+env:
+  UPGRADE_KAPPS_REPO_PATH: '.work/upgrade/kommander-applications'
+  UPGRADE_FROM_KOMMANDER_VERSION: 'v2.7.0'
+
 jobs:
   run-test-scenario:
     runs-on:
@@ -30,6 +34,13 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v4
+
+    - name: Checkout previous version of k-apps
+      if: ${{ github.event.inputs.scenario }} == 'upgrade'
+      uses: actions/checkout@v4
+      with:
+        path: apptests/${{env.UPGRADE_KAPPS_REPO_PATH}}
+        ref: ${{env.UPGRADE_FROM_KOMMANDER_VERSION}}
 
     - name: Set up Go
       uses: actions/setup-go@v4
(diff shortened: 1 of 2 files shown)


cwyl02:
testing:

successful run: https://github.com/mesosphere/kommander-applications/actions/runs/8545732022
skipped(when no label) run: https://github.com/mesosphere/kommander-applications/actions/runs/8545726018

previously this PR was based on https://github.com/mesosphere/kommander-applications/pull/2042, but now those commits with actual testing code are removed

.github/workflows/application-test.yaml
```diff
@@ -0,0 +1,41 @@
+name: Application Specific Test
+on:
+  pull_request:
```
mikolajb, line 3: Maybe it would be good to have a `workflow_dispatch` to run a test for any arbitrary application? wdyt?

.github/workflows/application-test.yaml
```diff
@@ -0,0 +1,41 @@
...
+          KAPP_NAME=""
+          for label in $PR_LABELS; do
+            echo $label
+            if [[ "$label" == "services/"* ]]; then
+              KAPP_NAME=$(echo "$label" | cut -d '/' -f 2)
+              break
+            fi
+          done
+
+          ginkgo --label-filter="install && $KAPP_NAME" appscenarios
```
mikolajb, line 41: I'm wondering why not just not running it on services that have changes, wdyt?

lukeogg: We have [this ticket](https://d2iq.atlassian.net/browse/D2IQ-100345) for running on the changes. Technically, if we have that, we don't need the labels to trigger specific individual tests explicitly.

I am thinking we need three ways to run the tests:
- An auto-triggered action that runs on PRs - this will run tests apps that have changed.
- A label/action that will trigger a specific test. (This would be helpful to override the change detection and run or re-run other tests.)
- A label/actions that will run all suites/tests.
Technically, the first two have some overlapping functionality.

wdyt?

cwyl02: I think relying on label is an easy way out since bump PRs like https://github.com/mesosphere/kommander-applications/pull/1964 is automatically labeled. 

vs. 

PR like this one(no change on k-apps but only GHA) will not be automatically labeled -- you can see the history I manually labeled this PR.

lukeogg: I think this works well. We already have the change detection and just use the labels to run the actions.

lukeogg (approved):

cwyl02:
test run successfully triggered maxtrix of 2 apps for both install and upgrade scenarios: https://github.com/mesosphere/kommander-applications/actions/runs/8559880372/job/23457616623?pr=2074

.github/workflows/application-test.yaml
cwyl02: addressed to refactor to trigger GHA via matrix of (app, install || upgrade).

Also added workflow dispatch on both the parent job to trigger for ALL apps, 

as well as to the individual scenario GHA of a single app on a single action.

.github/workflows/application-test.yaml
cwyl02: added

.github/workflows/application-test-scenario.yaml
```diff
@@ -0,0 +1,46 @@
...
+    inputs:
+      scenario:
+        description: 'The scenario to run, e.g., install or upgrade'
+        required: true
+        type: string
+      app:
+        description: 'The kommander application to test, e.g. reloader,kube-prometheus-stack'
+        required: true
+        type: string
+
```
lukeogg, line 24:
```suggestion

env:
  UPGRADE_KAPPS_REPO_PATH: '.work/upgrade/kommander-applications'
  UPGRADE_FROM_KOMMANDER_VERSION: 'v2.7.0'

```

.github/workflows/application-test-scenario.yaml
```diff
@@ -0,0 +1,46 @@
...
+jobs:
+  run-test-scenario:
+    runs-on:
+      - ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
```
lukeogg, line 33:
```suggestion

    - name: Checkout previous version of k-apps
      if: ${{ github.event.inputs.scenario }} == 'upgrade'
      uses: actions/checkout@v4
      with:
        path: apptests/${{env.UPGRADE_KAPPS_REPO_PATH}}
        ref: ${{env.UPGRADE_FROM_KOMMANDER_VERSION}}

```

lukeogg: And this? It will be needed for the upgrade tests.

.github/workflows/application-test-scenario.yaml
lukeogg: Do you mind adding this?

lukeogg (approved):

.github/workflows/application-test-scenario.yaml
mikolajb: Or even better:

```
$ gh dkp generate upgrade-matrix --json | jq --raw-output ".upgrades | to_entries | [last][0].value.from"
v2.8.0-dev
```

mikolajb: Actually, we need to upgrade from multiple version, e.g, right now 2.7 branch has to be tested with the following upgrade paths:

```
$ gh dkp generate upgrade-matrix                                                                   
From version | To version | k8s Version
v2.6.0       | v2.7.1-dev | 1.26
v2.6.1       | v2.7.1-dev | 1.26
v2.6.2       | v2.7.1-dev | 1.26
v2.6.3-dev   | v2.7.1-dev | 1.26
v2.7.0       | v2.7.1-dev | 1.27
```

The following `gh-dkp` commands might give you the information necessary to compute this:

```
gh dkp generate dev-version
gh dkp generate dev-versions
gh dkp generate upgrade-matrix
```

They all have json output.

cwyl02: @mikolajb let's address this in another PR

cwyl02:
admin merging to make other app specific PRs easier. the CI error is irrelevant.

PR #2059, opened 2024-04-03
feat: Bump reloader to 1.0.74

johnliu25:
Upgrades the following apps to use version 1.0.74 of chart reloader:
* reloader from version 1.0.65 to 1.0.74


gracedo:
https://github.com/mesosphere/kommander-applications/pull/2093

PR #2067, opened 2024-04-03
feat: Bump rook-ceph-cluster to v1.13.7

johnliu25:
Upgrades the following apps to use version v1.13.7 of chart rook-ceph-cluster:
* rook-ceph-cluster from version 1.12.7 to 1.13.7


gracedo:
https://github.com/mesosphere/kommander-applications/pull/2094

PR #2038, opened 2024-03-21
chore: bump ai nav and use priorityclass

cwyl02:
**What problem does this PR solve?**:
Bump the 2 ai navigator charts and use PriorityClass for all 3 ai navigator deployments:

- app -- this is defined in k-apps as a plain deployment
- agent -- this is defined in a helm chart
- API -- this is defined in a helm chart

depends on:
https://github.com/mesosphere/ai-navigator-cluster-info-agent/pull/16
https://github.com/mesosphere/ai-navigator-cluster-info-api/pull/22

**Which issue(s) does this PR fix?**:
https://d2iq.atlassian.net/browse/D2IQ-100289

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Checklist**

- [ ] If the PR adds a version bump, ensure there is no breaking change in Licensing model (or NA).
- [ ] If a chart is changed or app configuration is significantly changed, the chart version is correctly incremented (so that apps are not automatically upgraded from a previous version of DKP).

PR #2075, opened 2024-04-05
chore: bump ai nav and use priorityclass (#2038)

cwyl02:
* chore: bump ai nav and use priorityclass

* fix: path of hr

```release-note

```

**Checklist**


services/ai-navigator-app/0.2.2/helmrelease/helmrelease.yaml
```diff
@@ -11,7 +11,7 @@ spec:
         kind: HelmRepository
         name: mesosphere.github.io-ai-navigator-cluster-info-api-charts
         namespace: kommander-flux
-      version: 0.1.0
+      version: 0.2.1
```
gracedo, line 14: why is the app version 0.2.2 but chart version is 0.2.1? Can we match the app version to be 0.2.1 also?

cwyl02: on `main` the k-app is `0.2.2` while the chart version is `0.2.1`. for 2.7 I guess yes we can make k-app version 0.2.1 if we want

lukeogg (approved):

services/ai-navigator-app/0.2.2/helmrelease/helmrelease.yaml
gracedo: yeah let's do that. especially if `main` (or 2.8) has config only changes on top of `0.2.1` then we will want 2.7 to have a lower version so the upgrade is not triggered automatically (if 2.7 and 2.8 have the same app version).

gracedo:
rebasing should hopefully get the manifest validation check passing

services/ai-navigator-app/0.2.2/helmrelease/helmrelease.yaml
cwyl02: done

gracedo (commented):
`No mapping for image "docker.io/bitnami/kubectl:1.29.2"`
the new version of the ai nav chart bumps kubectl to 1.29.2. If we don't want to ship that version of kubectl in 2.7 you can override it in values.yaml to use the previous version

cwyl02:
> `No mapping for image "docker.io/bitnami/kubectl:1.29.2"` the new version of the ai nav chart bumps kubectl to 1.29.2. If we don't want to ship that version of kubectl in 2.7 you can override it in values.yaml to use the previous version

you read my mind! fixed

PR #2082, opened 2024-04-08
fix: centr grafana chart bump 2.7

cwyl02:
backporting the fix for the excessive RBAC permision of kommander-operator chart

https://d2iq.atlassian.net/browse/D2IQ-100406

```release-note

```

**Checklist**


gracedo (approved):

gracedo:
we'll need it in `release-2.8` too

PR #2088, opened 2024-04-09
refactor: dkp to nkp

nataliepina:
Application text refactoring to rename anything from DKP to NKP.

We need to rename things from DKP to NKP for the .NEXT demo / NKP 3.0.
This should be going into 3.0.

We will need new images from design for the application icons, so these will need to be changed separately.

We can merge this in whenever we're ready to have it updated. I wanted to prep the changes in case we should get this in for the .next demo. 

These changes should align with the changes in kommander-ui -- https://github.com/mesosphere/kommander-ui/pull/5535

```release-note
DKP has been renamed to NKP. 
```

**Checklist**


services/traefik/26.0.0/defaults/cm.yaml
```diff
@@ -197,7 +197,7 @@ data:
     pilot:
       dashboard: false
     # This value should be equal to release name. Justification with timeline of changes:
-    # 1. This was set to `kommander-treafik` in an old chart version - https://github.com/traefik/traefik-helm-chart/blob/v10.30.1/traefik/templates/deployment.yaml#L42 (Shipped in DKP 2.4)
+    # 1. This was set to `kommander-treafik` in an old chart version - https://github.com/traefik/traefik-helm-chart/blob/v10.30.1/traefik/templates/deployment.yaml#L42 (Shipped in NKP 2.4)
```
nataliepina, line 200: I was conflicted on changing things like this that were technically DKP to NKP. Let me know what you all think and if it makes sense to still have DKP here instead for anything prior to 3.0.

gracedo: maybe check with product?

nataliepina: Confirmed with Sonali, it should remain DKP for past versions. Reverting this change.

gracedo (approved):

PR #2105, opened 2024-04-11
fix: add general kommander overrides cm

mhrabovcin:
Remove versioning from the optional kommander chart overrides CM.

https://d2iq.atlassian.net/browse/D2IQ-100445

```release-note

```

**Checklist**


cwyl02:
looks like we need to fix the [unit test](https://github.com/mesosphere/kommander-applications/blob/main/hack/release/pkg/appversion/appversion_test.go#L79) too

cwyl02 (approved):

PR #2119, opened 2024-04-11
fix: Fix Kubernetes Audit Dashboard failing in Grafana Logging v9+

johnliu25:
Kubernetes Audit Dashboard failing in Grafana Logging v9+

https://d2iq.atlassian.net/browse/D2IQ-99728

Upgraded the Grafana version to version 10.4.2 which is the latest version. 

Test result in Grafana 10.3.3

Screenshot 2024-04-11 at 3 57 18 PM


services/grafana-logging/6.60.2/defaults/cm.yaml
```diff
@@ -7,9 +7,8 @@ data:
   values.yaml: |
     ---
     priorityClassName: "dkp-critical-priority"
-    # pinning grafana to 8.5 since it appears that v9+ is causing the kubernetes audit dashboard to crash
     image:
-      tag: 8.5.26
+      tag: 10.4.2
```
gracedo, line 11: what is the default version of grafana shipped with this helm chart? can we just use the default instead of overriding?

You'll have to make this same change to `project-grafana-logging` default values as well

johnliu25: yes. We need to make the same change to project-grafana-logging.

johnliu25:
Found error: The mapping for "docker.io/grafana/grafana:8.5.26" is unused and must be removed

cwyl02 (approved):

PR #2134, opened 2024-04-16
feat: cert-manager acme workarround test

lukeogg:

Adds test for workaround documented in https://support.d2iq.com/hc/en-us/articles/13421831855124-Custom-Domains-ACME-certificates-and-Custom-CA-certificates

https://d2iq.atlassian.net/browse/D2IQ-100418

```release-note

```

**Checklist**


mhrabovcin (approved):

mhrabovcin:
The license check is failing because it is picking up data from the apptests directory. We need to modify our `list-image.sh` script to ignore the dir https://github.com/mesosphere/kommander-applications/blob/main/hack/list-images.sh#L87

PR #2135, opened 2024-04-16
fix(release): Omit ai-nav helmrepos from bundle

gracedo:
Separate ai-nav helmrepos to separate file so we can exclude it from the bundle. I am depending on this kcli bug https://jira.nutanix.com/browse/D2IQ-97040 to continue to exist to ensure that the `kustomize` won't fail (kustomization.yaml specifies ai-nav file which is removed)

```release-note

```

**Checklist**


kaiwalyajoshi (approved):

msdolbey (approved):
👍 
very clever

PR #2137, opened 2024-04-16
fix(external-dns): Bump to 7.1.2

gracedo:
Bump to newer version of external-dns which does not introduce new error behavior which was causing the pod to crash-loop until proper configuration was provided.

https://jira.nutanix.com/browse/D2IQ-100511

```release-note

```

**Checklist**


msdolbey (approved):
👍 
wooooo

cwyl02 (approved):

PR #2145, opened 2024-04-17
ci: trigger upgrade via matrix

cwyl02:
currently it doesn't trigger upgrade tests from a matrix

https://jira.nutanix.com/browse/D2IQ-100345

```release-note

```

**Checklist**


cwyl02:
right now it triggers 2.7.x and 2.8.0-dev because we haven't cut the dev tag for latest development

PR #2154, opened 2024-04-22
fix: Modify files in the release archive

mikolajb:
It looks we have to modify a file while creating an archive.

https://jira.nutanix.com/browse/D2IQ-100557

```release-note

```

**Checklist**

### DIFF
--- a/make/release.mk
+++ b/make/release.mk
@@ -15,6 +15,7 @@ release:
 								  ":(exclude)common/helm-repositories/ai-navigator-repos.yaml" \
 								  ":(exclude)services/ai-navigator-app" \
 								  ":(exclude)services/ai-navigator-cluster-info-agent"
+	./hack/hack/build-archive.sh $(ARCHIVE_NAME)
 	aws s3 cp --acl $(S3_ACL) $(ARCHIVE_NAME) s3://$(S3_BUCKET)/$(S3_PATH)/
 	echo "Published to $(PUBLISHED_URL)"
 ifeq (,$(findstring dev,$(GIT_TAG)))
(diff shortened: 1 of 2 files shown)


make/release.mk
```diff
@@ -15,6 +15,7 @@ release:
 								  ":(exclude)common/helm-repositories/ai-navigator-repos.yaml" \
 								  ":(exclude)services/ai-navigator-app" \
 								  ":(exclude)services/ai-navigator-cluster-info-agent"
+	./hack/hack/build-archive.sh $(ARCHIVE_NAME)
```
mhrabovcin, line 18:
```suggestion
	./hack/build-archive.sh $(ARCHIVE_NAME)
```

Is there a duplicate `hack` in the path?

mhrabovcin:
Would it be possible to add a patch to `./common/airgapped` to not include these repositories instead of rebuilding tar archive. We could target a dedicated label on `HelmRelease` resource instead of hardcoding the AI navigator specific charts.

Example: https://github.com/kubernetes-sigs/kustomize/issues/1593#issuecomment-1545727117

mikolajb:
> Would it be possible to add a patch to `./common/airgapped` to not include these repositories instead of rebuilding tar archive. We could target a dedicated label on `HelmRelease` resource instead of hardcoding the AI navigator specific charts.
> 
> Example: [kubernetes-sigs/kustomize#1593 (comment)](https://github.com/kubernetes-sigs/kustomize/issues/1593#issuecomment-1545727117)

I think the idea is to not to have any trace of `ai` in our airgapped bundles. In other words - we have to remove it in build-time not run-time. Let's also hear from @gracedo as she authored the [previous changes](https://github.com/mesosphere/kommander-applications/pull/2135).

mhrabovcin (approved):

msdolbey (approved):
👍 
do bundle up

PR #2158, opened 2024-04-23
feat: Add basic kubecost tests for install and upgrade

lukeogg:

Adds basic application specific tests for Kubecost. Covers upgrade and install.

https://jira.nutanix.com/browse/D2IQ-100304

```release-note

```

**Checklist**

### DIFF
--- a/apptests/appscenarios/suite_test.go
+++ b/apptests/appscenarios/suite_test.go
@@ -2,7 +2,11 @@ package appscenarios
 
 import (
 	"context"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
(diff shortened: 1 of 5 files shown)


apptests/appscenarios/suite_test.go
```diff
@@ -39,8 +58,8 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = AfterSuite(func() {
-	err := env.Destroy(ctx)
-	Expect(err).ToNot(HaveOccurred())
+	//err := env.Destroy(ctx)
```
cwyl02, line 61: do we want to comment this out or its only for local testing?

lukeogg: For local testsing... Good catch, thanks!

cwyl02 (approved):

PR #2161, opened 2024-04-25
fix: Use correct constant for app name

lukeogg:

Use the correct constant for the `kubecost` app name in the tests

```release-note

```

**Checklist**


cwyl02 (approved):

msdolbey (approved):
👍 
Yeah that makes sense

PR #2192, opened 2024-05-01
feat: Application specific tests for kommander-flux

lukeogg:

Adds application specific tests for kommander-flux. The test suite installed the latest Flux version as part of the cluster setup. This refactors the cluster setup and Flux install to a `BeforeEach` with an `OncePerOrdered` decorator in a parent container. This allows a specific Flux version to be installed for the commander-flux tests.

All tests can now be run without a label selector for a single test: `ginkgo appscenarios`

https://jira.nutanix.com/browse/D2IQ-100298

```release-note

```

**Checklist**


apptests/environment/environment.go
```diff
@@ -6,14 +6,15 @@ import (
 	"context"
 	"fmt"
 	"github.com/drone/envsubst"
+	runclient "github.com/fluxcd/pkg/runtime/client"
```
cwyl02, line 9: nit: how about regroup the `k8s.io` and `github.com` imports to lower sections?

apptests/appscenarios/kommanderflux.go
```diff
@@ -0,0 +1,80 @@
+package appscenarios
+
+import (
+	"context"
+	"github.com/mesosphere/kommander-applications/apptests/constants"
+	"github.com/mesosphere/kommander-applications/apptests/environment"
+	"path/filepath"
```
cwyl02, line 7: nit: same as the other comment

cwyl02 (approved):

PR #2204, opened 2024-05-07
test: traefik test

cwyl02:
add traefik app test

https://jira.nutanix.com/browse/D2IQ-100302

```release-note

```

**Checklist**


apptests/appscenarios/traefik.go
```diff
@@ -0,0 +1,120 @@
...
+
+	return err
+}
+
+func (t traefik) applyTraefikOverrideCM(ctx context.Context, env *environment.Env, cmName string) error {
+	testDataPath, err := getTestDataDir()
+	if err != nil {
+		return err
+	}
+	cmPath := filepath.Join(testDataPath, "traefik", "override-cm.yaml")
```
lukeogg, line 110: Is the `override-cm.yaml` file missing?

lukeogg (approved):

apptests/appscenarios/traefik.go
cwyl02: that file is already in `main` due to karma PR.

PR #2207, opened 2024-05-07
feat: Add External DNS tests

lukeogg:

Adds app-specific tests for External DNS.
- Tests use configuration as specified in `kommander-applications`, though a typical installation would need some configuration.
- These tests do not test an AWS or Azure install. We assume that External DNS does what it is supposed to and that our [documentation](https://docs.d2iq.com/dkp/2.7/configure-an-external-dns-management-or-essential-) is not Kommander/DKP specific.
- We are primarily looking for the metrics servers to be up and emitting metrics.

https://jira.nutanix.com/browse/D2IQ-100300

- There is an "issue" with configuration that is in `v2.7.0` version of `kommander-applications`. This has external-dns `v0.13.6` (app/chart version `v6.26.1`). This version exits with an error code in a kind cluster with the default AWS provider. There are similar problems with other providers. The upgrade tests will not work from this version.

```release-note

```

**Checklist**

### DIFF
--- a/apptests/appscenarios/externaldns_test.go
+++ b/apptests/appscenarios/externaldns_test.go
@@ -2,7 +2,6 @@ package appscenarios
 
 import (
 	"fmt"
-	corev1 "k8s.io/api/core/v1"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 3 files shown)


apptests/appscenarios/externaldns_test.go
```diff
@@ -0,0 +1,252 @@
+package appscenarios
+
+import (
+	"fmt"
+	corev1 "k8s.io/api/core/v1"
```
cwyl02, line 5: nit: regroup this import

cwyl02 (approved):

PR #2227, opened 2024-05-14
feat: Add licensing tiers to Kommander Apps.

kaiwalyajoshi:

Add Licensing Tier Annotations as per the [Feature Matrix](https://docs.google.com/spreadsheets/d/1M2qhRebO3DPcrn662D7tw4DRXk8tuM4ICIMMGRS6rWw/edit#gid=0).

JIRA: https://jira.nutanix.com/browse/NCN-100668

```release-note

```

**Checklist**


cwyl02:
how about `centralized-grafana`, `centralized-kubecost` and `karma`?

gracedo (approved):
https://github.com/mesosphere/kommander-applications/blob/2ae3273683331a4763728b715d8bf527757b32ac/services/kommander/0.12.0/defaults/cm.yaml#L69-L77
These "internal" apps are deployed by default on the mgmt cluster today when an enterprise license is applied -- do we need to do the same for an Ultimate license? I noticed that these apps also don't have the `licensing` key applied, is this because they are not exposed to the UI to begin with? Just wondering if it would make sense to add it just for "completeness" in case we end up wanting to expose them to the UI (for example, if we do not deploy them by default via controller, there is no other way for a user to enable them except manually using CLI)

kaiwalyajoshi:
@gracedo @cwyl02 added licensing tiers to all K-Apps now, including the internal ones.

PR #2240, opened 2024-05-20
feat(koperator): Add missing permissions and pod label

gracedo:
Add missing permissions and pod label to koperator

https://jira.nutanix.com/browse/NCN-100844

```release-note

```

**Checklist**

### DIFF
--- a/charts/kommander-operator/templates/rback_v1_clusterrole_operator_kommander-operator.yaml
+++ b/charts/kommander-operator/templates/rback_v1_clusterrole_operator_kommander-operator.yaml
@@ -40,6 +40,7 @@ rules:
       - gitrepositories
       - secrets
       - events
+      - configmaps
     verbs:
       - get
       - list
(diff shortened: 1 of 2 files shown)


charts/kommander-operator/templates/rback_v1_clusterrole_operator_kommander-operator.yaml
```diff
@@ -58,6 +59,7 @@ rules:
       - kustomize.toolkit.fluxcd.io
       - types.kubefed.io
       - kommander.mesosphere.io
+      - ""
```
cwyl02, line 62: what's the effect of this empty string apiGroup?

gracedo: `""` --> core api group
https://kubernetes.io/docs/reference/using-api/#api-groups
> The core (also called legacy) group is found at REST path /api/v1. The core group is not specified as part of the apiVersion field, for example, apiVersion: v1.

cwyl02 (approved):

PR #2244, opened 2024-05-23
feat: Rook Ceph application specific tests

lukeogg:

- Add tests for Rook Ceph

https://jira.nutanix.com/browse/NCN-100306

```release-note

```

**Checklist**


apptests/kind/config/kind.yaml
```diff
@@ -4,13 +4,47 @@ networking:
...
+        containerPath: /hack/scripts
+      - hostPath: ./testdata/rook-ceph/manifests/
+        containerPath: /hack/manifests
+    kubeadmConfigPatches:
+      - |
+        kind: InitConfiguration
+        nodeRegistration:
+          kubeletExtraArgs:
+            node-labels: "ingress-ready=true"
+  - role: worker
```
cwyl02, line 30: probably not adding too much overhead, but do we have to have 2 workers?

lukeogg: Yeah, still testing this. I don't think so.

lukeogg: Works with one node.

cwyl02 (approved):

PR #2249, opened 2024-05-24
fix: [backport-2.8] override with a higher memory limit (#2247)

cwyl02:
and also explicitly put the chart defaults down in our default CM

(cherry picked from commit 70b4cc28b8c040ab3cd7096403adb67d11577959)

```release-note

```

**Checklist**


msdolbey (approved):
👍 
But let us not have memory loss

gracedo (commented):
bump app version?

PR #2259, opened 2024-05-30
feat: Release artifacts to OCI registry

mikolajb:
Introduces a new job that pushes k-apps contents to OCI registry

https://jira.nutanix.com/browse/NCN-100943

```release-note

```

**Checklist**


justfile
```diff
@@ -0,0 +1,7 @@
+repository := "mesosphere"
+include_file := justfile_directory() / ".include-airgapped"
+exclude_file := justfile_directory() / ".exclude-airgapped"
+
+release-oci tag tmp_dir=`mktemp --directory`:
+    rsync --info=name --archive --recursive --files-from={{ include_file }} --exclude-from={{ exclude_file }} {{ justfile_directory() }} {{ tmp_dir }}
+    echo "${DOCKER_PASSWORD}" | oras push --password-stdin --user "${DOCKER_USERNAME}" --verbose {{ repository }}/kommander-applications:{{ tag }} {{ tmp_dir }}
```
alembiewski, line 7: `mktemp` returns an absolute path to the directory, and `oras push` will fail unless the `--disable-path-validation` is provided. If the flag is set, `oras push` will not complain but it will preserve the whole directory structure, starting from `/`, replicating what the `tmp_dir` is pointing to, so we might want to call `oras push` from the `tmp_dir` directly, e.g. `oras push <args> .`

justfile
```diff
@@ -0,0 +1,7 @@
+repository := "mesosphere"
```
alembiewski, line 1: We need to specify the target registry, otherwise `oras push` will fail
```suggestion
reqistry := "registry-1.docker.io"
repository := "mesosphere/kommander-applications"
```

alembiewski (approved):
LGTM!

PR #2298, opened 2024-06-11
fix: Revert "feat: add gitops repo structure (#2284)"

gracedo:
This reverts commit f607e858f8fab735adf522a1308e3d072ab909d8.

I think that this is breaking kcli CI -- the double install test case. I think what might be happening is that when we run kommander install (or even upgrade) we end up overwriting the kustomization.yaml files with these empty ones, which result in apps getting uninstalled. In kcli we avoided writing the bootstrap gitrepo files if we saw that the files had already been created. We would need to do something similar here, or if its too complicated, stick with bootstrap gitrepo being written in kcli and bootstrapper.

https://jira.nutanix.com/browse/NCN-100425

```release-note

```

**Checklist**


cwyl02 (approved):

gracedo:
admin merging to fix CI.. this shouldn't affect the failing checks here (right now everything's borked due to hitting 401 fetching nvidia helm chart)

PR #2318, opened 2024-06-18
Traefik 3.0.0

shubham2g:

```release-note

```

**Checklist**


gracedo:
There is actually no traefik v3 helm chart (the v3 I referred to in the ticket was the traefik app version) -- there is already a PR open with the bump to a new chart version that bumps the traefik app version to v3 https://github.com/mesosphere/kommander-applications/pull/2180 please use this PR for testing and to make any necessary changes

PR #2322, opened 2024-06-20
feat: Bump k8s and kubectl to 1.29.6

gracedo:
bump k8s to 1.29.6
bump kubectl image in apps to 1.29.6

https://jira.nutanix.com/browse/NCN-101202

```release-note

```

**Checklist**

### DIFF
--- a/licenses.d2iq.yaml
+++ b/licenses.d2iq.yaml
@@ -42,7 +42,7 @@ ignore:
   #     - url: https://github.com/bitnami/bitnami-docker-kubectl
   #       ref: ${image_tag}-debian-10-r31
   #       license_path: LICENSE
-  - docker.io/bitnami/kubectl:1.29.2
+  - docker.io/bitnami/kubectl:1.29.6
 
   # Fossa cannot scan C/CPP projects
   # See: https://docs.fossa.com/docs/c-cpp
--- a/licenses.d2iq.yaml
+++ b/licenses.d2iq.yaml
@@ -44,6 +44,11 @@ ignore:
   #       license_path: LICENSE
   - docker.io/bitnami/kubectl:1.29.6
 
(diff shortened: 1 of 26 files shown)


licenses.d2iq.yaml
```diff
@@ -42,7 +42,7 @@ ignore:
   #     - url: https://github.com/bitnami/bitnami-docker-kubectl
   #       ref: ${image_tag}-debian-10-r31
   #       license_path: LICENSE
-  - docker.io/bitnami/kubectl:1.29.2
+  - docker.io/bitnami/kubectl:1.29.6
```
gracedo, line 45: > No mapping for image "docker.io/bitnami/kubectl:1.29.2"

Been trying to track down this kubectl image... it looks like it's coming from cert-federation and cluster-observer, the two "dynamic" helmreleases that we deploy with override values via controller. so our tooling here must be using the kubectl image default in the chart values.yaml, since the overrides are "dynamically" created in the controller, not defined anywhere here. I will add the image back into the license file here to get it passing... but in practice I don't think we would actually be using 1.29.2 (I think it will continue to get bundled in the airgapped bundle unless we fix the tooling) cc @mhrabovcin 

cert-federation overrides
https://github.com/mesosphere/kommander/blob/a9bf635d1a6697c234f165133a63db85170fd4fe/federation/pkg/controllers/embedded/federated-certificates/thanos-server-tls.yaml.tmpl#L25

cluster-observer overrides which is using kubetools img
https://github.com/mesosphere/kommander/blob/a9bf635d1a6697c234f165133a63db85170fd4fe/federation/pkg/controllers/kommanderclusters/kommandercluster_observer_controller.go#L253-L257

mhrabovcin (approved):

PR #2343, opened 2024-06-27
feat(kommander): Remove SSO from default mgmt apps

gracedo:
we decided to move SSO apps into kcore to install

https://jira.nutanix.com/browse/NCN-101330

```release-note

```

**Checklist**

### DIFF
--- a/services/dex-k8s-authenticator/1.3.2/dex-k8s-authenticator.yaml
+++ b/services/dex-k8s-authenticator/1.3.2/dex-k8s-authenticator.yaml
@@ -35,4 +35,7 @@ spec:
   valuesFrom:
     - kind: ConfigMap
       name: dex-k8s-authenticator-1.3.2-d2iq-defaults
+    - kind: ConfigMap
+      name: dex-k8s-authenticator-overrides
+      optional: true
   targetNamespace: ${releaseNamespace}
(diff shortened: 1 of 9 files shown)


dkoshkin:
Konvoy is getting closer to integrating with this and waiting on this condition, anything blocking this PR from being marked ready?

gracedo:
> Konvoy is getting closer to integrating with this and waiting on this condition, anything blocking this PR from being marked ready?

@dkoshkin still testing 🥲 but the condition won't change, so this shouldn't block konvoy. The condition will just wait for more apps when this is merged

mikolajb (approved):

mhrabovcin:
@gracedo why do we need those new overrides definitions?

gracedo:
> @gracedo why do we need those new overrides definitions?

@mhrabovcin this is more to maintain backwards-compatibility with how kcli install is currently used. If a user configures an app in the installation config yaml, we should process it and apply it to these core apps (deployed by kommander-operator) automatically at install instead of installing the core apps with default config and then in later kcli step, applying the override configs (resulting in reinstall). In many of our docs we ask users to provide config overrides in installation yaml file (e.g. acme docs) and I am worried that if we do not apply overrides at install, it may cause the deployment to fail.

For nutanix clusters (using prompt-based cli) nothing changes since we do not allow app configuration. But just in case if we ever needed a workaround to apply overrides at install, we can provide this (unofficially) by asking them to manually create these override cms before we hit the deployment step.

Longer-term, in a later release we will design a solution to allow passing user configuration for apps into the operator

mhrabovcin:
Thanks @gracedo

PR #2367, opened 2024-07-01
ci: Clone kommander repo for flux update script

mikolajb:
I don't know how it worked before...

https://github.com/mesosphere/kommander-applications/actions/runs/9747457924/job/26900250612#step:7:140

```release-note

```

**Checklist**


takirala (approved):
I imagine this didn't work since this was moved to GHA from TC.

PR #2370, opened 2024-07-01
feat: Conditional cluster teardown in app tests

lukeogg:

Skips the teardown of the test cluster in the CI. THis allows the support bundle to be collected as needed.

```release-note

```

**Checklist**


cwyl02 (approved):

gracedo (approved):
can cluster teardown be run after support bundle upload somehow?

PR #2377, opened 2024-07-05
feat: giving kommander-operator cluster-admin

cwyl02:
add cluster-admin CRB needed for kommander operator

```release-note

```

**Checklist**

### DIFF
--- a/charts/kommander-operator/templates/rback_v1_clusterrole_operator_kommander-operator.yaml
+++ b/charts/kommander-operator/templates/rback_v1_clusterrole_operator_kommander-operator.yaml
@@ -113,6 +113,7 @@ rules:
       - helm.toolkit.fluxcd.io
       - kustomize.toolkit.fluxcd.io
       - apps.kommander.d2iq.io
+      - rbac.authorization.k8s.io
     resources:
       - events
       - configmaps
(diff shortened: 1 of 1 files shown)


charts/kommander-operator/templates/rback_v1_clusterrole_operator_kommander-operator.yaml
```diff
@@ -113,13 +113,26 @@ rules:
...
       - events
       - configmaps
       - secrets
       - helmreleases
       - kustomizations
       - appdeployments
+      - clusterrolebindings
     verbs:
       - create
+  # cluster-admin permissions needed here
```
gracedo, line 127: why do we need cluster-admin?

cwyl02: because now the kommander-operator needs to create a CRB that binds the default user with `cluster-admin` CR, and without this permissions, it is a permission escalation and k8s API server denies the CRB creation. right not with these changes KUTTL passed, except bootstrapper Kuttl.

gracedo (approved):

mhrabovcin (approved):
I think we can remove other permissions that are now covered by having cluster admin perms from the rules definition?

cwyl02:
> I think we can remove other permissions that are now covered by having cluster admin perms from the rules definition?

agree. I plan to do that in a separate PR

PR #2378, opened 2024-07-08
fix: kubecost patched image with fixed CVEs

SandhyaRavi2403:
   Patch kubecost image with fixed CVE's

workflow link:
https://github.com/mesosphere/dkp-container-images/actions/runs/9833998335

https://jira.nutanix.com/browse/NCN-101394

```release-note

```

**Checklist**


licenses.d2iq.yaml
```diff
@@ -10,7 +10,7 @@ ignore:
   - docker.io/bitnami/memcached:1.6.19-debian-11-r7
   - docker.io/library/busybox:1
   - gcr.io/kubecost1/cost-model:prod-1.106.7
-  - gcr.io/kubecost1/frontend:prod-1.106.7
```
mhrabovcin, line 13: The new image location is `ghcr.io/mesosphere/dkp-container-images/cr.fluentbit.io/fluent/fluent-bit:2.2.3-d2iq.0` so we need to change the reference in the kubecost config map.

PR #2391, opened 2024-07-10
fix: [backport-2.7] velero with bumped limits 

cwyl02:
missed bumps for velero limits fix

```release-note

```

**Checklist**


services/velero/4.1.2/helmrelease/helmrelease.yaml
```diff
@@ -11,7 +11,7 @@ spec:
         kind: HelmRepository
         name: vmware-tanzu.github.io
         namespace: kommander-flux
-      version: 4.1.1
+      version: 4.1.2
```
gracedo, line 14: didn't technically need to bump the helm chart version but I guess since this version does exist doesn't hurt 🤷🏻‍♀️

gracedo (approved):

PR #2409, opened 2024-07-15
fix: make git-operator images appear in the airgapped bundle

asekretenko:
Makes the airgapped installation (incl. airgapped validation tests) work with Git Operator.

A temporary solution to https://jira.nutanix.com/browse/NCN-101519

The local docker registry used in Kommander airgapped validation tests (and probably some other docker registry) exhibits a weird behaviour where it fails to provide an image when the image ref uses the `image:tag@sha256...` syntax; this is why this PR strips the SHA part in `git-operator` manifests.
 

```release-note

```

**Checklist**

### DIFF
--- a/licenses.d2iq.yaml
+++ b/licenses.d2iq.yaml
@@ -1,8 +1,10 @@
 ignore:
-  - ghcr.io/mesosphere/git-operator:v0.2.0
+  - docker.io/mesosphere/git-operator:v0.8.3
+  - docker.io/mesosphere/gitwebserver:v0.8.3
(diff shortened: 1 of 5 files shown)


licenses.d2iq.yaml
```diff
@@ -1,8 +1,10 @@
 ignore:
-  - ghcr.io/mesosphere/git-operator:v0.2.0
+  - docker.io/mesosphere/git-operator:v0.8.3
+  - docker.io/mesosphere/gitwebserver:v0.8.3
```
mhrabovcin, line 3: Can you add mappings for these?

asekretenko: Thanks, completely forgot that we do have mappings for components we build as well.

PR #2460, opened 2024-07-25
fix: Add contents to git-operator defaults kustomization

mikolajb:
An empty kustomization causes the flux kustomization failure.

https://jira.nutanix.com/browse/NCN-101714

```release-note

```

**Checklist**


gracedo (approved):
the field I've been adding to these empty kustomizations has been
```
buildMetadata:
  - originAnnotations
 ```

PR #2463, opened 2024-07-25
feat: Fix Traefik rewrite for api url

lukeogg:

Work around for Traefik issue: https://github.com/traefik/traefik/issues/5853

URLs with `api` later in the URL are also being rewritten resulting in a 404.

As a workaround, pin the match to the beginning of the URL by matching the label too:

```
      rewrites:
      - regex: apiUrl:\"/api\"
        replacement: apiUrl:"/dkp/traefik/api"
```

See Slack: https://nutanix.slack.com/archives/C067BPT2MQ9/p1719940306056019

https://jira.nutanix.com/browse/NCN-101671

```release-note

```

**Checklist**

### DIFF
--- a/services/traefik/27.0.2/defaults/cm.yaml
+++ b/services/traefik/27.0.2/defaults/cm.yaml
@@ -135,8 +135,8 @@ data:
               plugin:
                 plugin-rewritebody:
                   rewrites:
-                    - regex: "/api"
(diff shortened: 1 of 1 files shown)


cneth (approved):
LGTM, and I tested this fix on a cluster and it resolved the issue.

cwyl02 (approved):

PR #2583, opened 2024-08-26
fix(ceph): Set priorityclass for ceph-exporter

gracedo:
the new ceph-exporter pods were missing priority class

https://jira.nutanix.com/browse/NCN-102111

```release-note

```

**Checklist**

### DIFF
--- a/services/rook-ceph-cluster/1.14.5/defaults/cm.yaml
+++ b/services/rook-ceph-cluster/1.14.5/defaults/cm.yaml
@@ -62,6 +62,7 @@ data:
       # priority classes to apply to ceph resources
       priorityClassNames:
         crashcollector: system-cluster-critical
+        exporter: system-node-critical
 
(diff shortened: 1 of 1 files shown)


services/rook-ceph-cluster/1.14.5/defaults/cm.yaml
```diff
@@ -62,6 +62,7 @@ data:
       # priority classes to apply to ceph resources
       priorityClassNames:
         crashcollector: system-cluster-critical
+        exporter: system-node-critical
```
takirala, line 65: osd's and mgr's are the most important things and are needed for healthy quorum so maybe we can do this instead:

```suggestion
      priorityClassNames:
        all: system-cluster-critical
        osd: system-node-critical
        mgr: system-node-critical
```

takirala (approved):

PR #2665, opened 2024-09-26
build(gha): Flux update fixes

gracedo:
Fixed up a few things in the automated flux bump workflow. Working test run https://github.com/mesosphere/kommander-applications/actions/runs/11060715919 which produced PR https://github.com/mesosphere/kommander-applications/pull/2664 (closed since I was triggering it against this branch so an extra commit was on there)

If the branch already exists then it should exit gracefully as a success instead of looking like a failed run.

https://jira.nutanix.com/browse/NCN-102720

```release-note

```

**Checklist**

### DIFF
--- a/hack/flux/update-flux.sh
+++ b/hack/flux/update-flux.sh
@@ -11,7 +11,7 @@ readonly CURRENT_FLUX_VERSION
 KOMMANDER_REPO_PATH="${REPO_ROOT}/kommander" # Override in CI to path of kommander repository.
 
 function check_remote_branch() {
-    if [[ -n $(git ls-remote --exit-code --heads git@github.com:mesosphere/"$1".git "$2") ]]; then
+    if [[ -n $(git ls-remote --exit-code --heads https://github.com/mesosphere/"$1".git "$2") ]]; then
         echo "Flux update PR is already up!"
         exit 0
     fi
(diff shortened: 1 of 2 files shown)


hack/flux/update-flux.sh
```diff
@@ -45,24 +45,16 @@ function update_flux() {
...
-
     readonly COMMIT_MSG="feat: Upgrade flux to ${LATEST_FLUX_VERSION}"
 
     git commit -m "${COMMIT_MSG}"
 
     git push --set-upstream origin "${BRANCH_NAME}"
 
     git fetch origin main
-    KOMMANDER_APPLICATIONS_PR=$(gh pr create --base main --fill --head "${BRANCH_NAME}" -t "${COMMIT_MSG}" -l ready-for-review -l ok-to-test -l slack-notify -l open-kommander-pr -l update-licenses)
+    KOMMANDER_APPLICATIONS_PR=$(gh pr create --base main --fill --head "${BRANCH_NAME}" -t "${COMMIT_MSG}" -l ready-for-review -l ok-to-test -l slack-notify -l update-licenses)
```
takirala, line 57: Is the slack notify label still working? Either way we can drop the ok to test label and perhaps add test/kuttl label

takirala (approved):

cwyl02 (approved):

hack/flux/update-flux.sh
gracedo: these are labels for the k-apps PR I believe -- for the kommander PR I added all the `test/*` labels. I don't think slack-notify works but maybe we should fix that eventually so I'll just leave it here until then 😬

PR #2666, opened 2024-09-27
feat: bump bloodhound to validate against 1.31.0

rajgitnutanix:
bloodhound to validate against 1.31.0
(https://jira.nutanix.com/browse/NCN-102262)


.bloodhound.yml
```diff
@@ -4,6 +4,8 @@ k8s_versions:
   - 1.28.7
   - 1.29.6
   - 1.30.1
```
mhrabovcin, line 6: Can you remove this line, we don't need to validate against `1.30.1` anymore.

rajgitnutanix: Ok removed as suggested , Martin.

.bloodhound.yml
```diff
@@ -3,7 +3,8 @@
 k8s_versions:
   - 1.28.7
   - 1.29.6
-  - 1.30.1
+  - 1.30.5
+  - 1.31.0
```
gracedo, line 7:
```suggestion
  - 1.31.1
```

gracedo (approved):

PR #2680, opened 2024-09-30
feat: Add server that provides k-apps contents

mikolajb:
Produce a docker image with http server.

https://jira.nutanix.com/browse/NCN-102822

To move things quickly, I'm using github.com/sigoden/dufs. I use it because it supports tls but I'm planning to create our own (maybe a few lines in Python - let's see what's the fastest).

```release-note

```

**Checklist**

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,4 +1,4 @@
-name: Go test
+name: Tests
 on:
   push:
     tags:
@@ -31,6 +31,9 @@ jobs:
         run: |
           devbox run just git-operator-fetch-manifests
           git diff --exit-code
+
+      - name: Check if kommander application server works
+        run: devbox run -- just test-server
 
       - name: Report Coveralls
         uses: coverallsapp/github-action@v2
--- a/devbox.json
+++ b/devbox.json
@@ -1,8 +1,10 @@
 {
   "packages": {
     "awscli2":         "latest",
+    "curl":            "latest",
     "direnv":          "latest",
     "docker":          "latest",
+    "docker-buildx":   "latest",
     "envsubst":        "latest",
     "fluxcd":          "latest",
     "github-cli":      "latest",
--- a/devbox.lock
+++ b/devbox.lock
@@ -65,6 +65,130 @@
         }
       }
     },
+    "curl@latest": {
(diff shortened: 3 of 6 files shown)


takirala (commented):
Will this server be a thing in 2.14 ? If not, wdyt of using just a simple container image that has the necessary artifacts at a certain known path. pulling it as init container will be enough. I am not against the idea of server but this seems like an overkill for a soon to be dead code.

mikolajb:
> Will this server be a thing in 2.14 (or if/when we have OCI support) ? If not, wdyt of using just a simple container image that has the necessary artifacts at a certain known path. pulling it as init container will be enough. I am not against the idea of server but this seems like an overkill for a soon to be dead code (and for something that is very short lived - bootstrap job takes less than 10m in happy path). 

Yes, I get it, the choice is between having a server and relying on a git operator way of initializing the repository (that won't change when we move to OCI) or having an init container and creating a custom way of initializing the repository that will go away when we move to the OCI. I would favor the former.

In other words, this web server fits in the APIs and git operator not the other way around.

takirala:
So in the current way, we would have to:
- refactor kapps tooling to publish oci image in 2.14
- refactor bootstrap job to consume oci repo (instead of fetching from http server)

But, if we can publish a docker image (that is very close a future oci image, basically a simple image that has some file system) then we can keep the future refactoring minimal by
- refactoring kapps by pushing the oci image instead of the plain docker image (this refactor ll be much more minimal compared to whats being done in this PR)
- refactoring bootstrap job to get the content using https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/ in 2.14 or getting the same contents from `.status.artifact.url` of `OCIRepository` which will again be a signifcantly smaller refactor compared to moving away from hitting an http endpoint.

mikolajb:
> So in the current way, we would have to:
> - refactor kapps tooling to publish oci image in 2.14
> - refactor bootstrap job to consume oci repo (instead of fetching from http server)
> 

We discussed it on Slack but for someone reading it: both are already happening actually. The oci artifact is already being published since 2.12. Git operator is being able to consume OCI artifact when initializing a repository.

In other words: this method is reducing the future refactoring to the absolute minimum.

takirala (approved):

PR #2744, opened 2024-10-16
feat: kube-rbac-proxy version bump

ArvinderPal09:
Bumps kube-rbac-proxy version to 0.18.1

https://jira.nutanix.com/browse/NCN-102774


ArvinderPal09:
@mhrabovcin 
I tried bumping the kube-rbac-proxy version in the two files ( services/dex/2.13.14/dex.yaml
/services/kube-prometheus-stack/60.4.1/helmrelease/kube-prometheus-stack.yaml) but unfortunately didn't have any success with it. 
This is the reference link for rbac-proxy : https://github.com/prometheus-community/helm-charts/blob/848f830e5dabb8027e42e090ba94e05e5f1b6c75/charts/kube-state-metrics/values.yaml#L106

There seems to be no visible effect or improvement after the update. If you've got any suggestions or alternate steps we could try, I'd appreciate your input.

mhrabovcin:
@ArvinderPal09 the `dex` kube-rbac-proxy comes from the bundled `dex-controller` chart - https://github.com/mesosphere/dex-controller/blob/master/charts/dex-controller/values.yaml#L21-L26

You need to bump it in the dex configmap overrides with the right nesting so that value gets picked up - https://github.com/mesosphere/kommander-applications/blob/main/services/dex/2.14.0/defaults/cm.yaml#L105-L117

mhrabovcin:
@ArvinderPal09 it seems that dex is currently last instance of outdated `kube-rbac-proxy` container image. So lets not change any `kube-prometheus-stack` values and focus on bumping image in `dex` app.

licenses.d2iq.yaml
```diff
@@ -313,7 +313,7 @@ resources:
       - license_path: LICENSE
         ref: knative-v1.10.0
         url: https://github.com/knative/serving
-  - container_image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
+  - container_image: quay.io/brancz/kube-rbac-proxy:v0.18.1
```
mhrabovcin, line 316: Could you remove duplicate records so that we have only 1 entry for the `quay.io/brancz/kube-rbac-proxy:v0.18.1`

ArvinderPal09: Sure

mhrabovcin (approved):

PR #2772, opened 2024-10-23
revert: "feat: bump cert-manager to 1.16.1"

cwyl02:
Reverts mesosphere/kommander-applications#2743


jimmidyson:
@cwyl02 Why is this being reverted?

cwyl02:
> @cwyl02 Why is this being reverted?

closed. have been having issue with kommander CI but now confirmed not related to this bump

PR #2786, opened 2024-10-30
fix: cve bump from v0.0.5 to v0.0.6

saikumar98419:

```release-note

```

**Checklist**


services/logging-operator/4.2.5/extra-images.txt
```diff
@@ -1,4 +1,4 @@
 {{ .Values.fluentbit.image.repository }}:{{ .Values.fluentbit.image.tag }}
```
mhrabovcin, line 1: We need to bump the service version from `4.2.5` to `4.2.6` and all references in the service file.

SandhyaRavi2403:
@mhrabovcin,

The above 2 commits are not signed. Is it fine or should a new PR be raised ?

PR #2860, opened 2024-11-22
feat: Patch AI Navigator to the latest version (backport to 2.12)

lukeogg:

Backports the model size issue fixes to 2.12. Brings AI Navigator up to date in 2.12.

https://jira.nutanix.com/browse/NCN-104316 is the primary one.

```release-note

```

**Checklist**


licenses.d2iq.yaml
```diff
@@ -630,3 +631,13 @@ resources:
       - url: https://github.com/kubernetes/kubernetes
         ref: master
         license_path: LICENSE
+  - container_image: docker.io/bitnami/kubectl:1.30.5
```
kaiwalyajoshi, line 634:
```suggestion
  - container_image: docker.io/bitnami/kubectl:1.29.6
```
NKP v2.12.0 is based on k8s v1.29.6

licenses.d2iq.yaml
```diff
@@ -630,3 +631,13 @@ resources:
...
+    sources:
+      - url: https://github.com/kubernetes/kubectl
+        ref: v0${image_tag#1}
+        license_path: LICENSE
+  - container_image: docker.io/bitnami/postgresql:15.8.0-debian-12-r14
```
kaiwalyajoshi, line 639: There's a patched d2iq container found [here](https://github.com/mesosphere/kommander-applications/blob/012ea26aecbaf19141e210b0f4c4b42aa6db06e9/licenses.d2iq.yaml#L608) from v2.12.0 : `ghcr.io/mesosphere/dkp-container-images/docker.io/bitnami/postgresql:15.2.0-debian-11-r21-d2iq.0` use this one instead?

services/ai-navigator-app/0.2.8/defaults/cm.yaml
```diff
@@ -112,4 +90,4 @@ data:
     tolerations: []
 
     affinity: {}
-    kubectlImage: ${kubetoolsImageRepository:=bitnami/kubectl}:${kubetoolsImageTag:=1.29.6}
+    kubectlImage: ${kubetoolsImageRepository:=bitnami/kubectl}:${kubetoolsImageTag:=1.30.5}
```
kaiwalyajoshi, line 93: We should revert to the v1.29.6 bitnami image that ships with NKP 2.12.x

kaiwalyajoshi (changes requested):
Requesting changes to use the v1.29.6 kubectl images and using the patched version of Postgres that ships with NKP 2.12.x

services/ai-navigator-app/0.2.8/defaults/cm.yaml
lukeogg: I suspect `1.30` is backward compatible, but I suppose consistency is good.

licenses.d2iq.yaml
lukeogg: `kubectl` @ 1.30 is surely backward compatible. The idea was that we tested with these dependencies. I doubt `kubectl` matters though - I'll back it out.

services/ai-navigator-app/0.2.8/ai-navigator-app.yaml
```diff
@@ -60,7 +63,7 @@ spec:
       serviceAccountName: ai-navigator-app
       containers:
         - name: ai-navigator-app
-          image: mesosphere/ai-navigator-app:v0.1.1
+          image: mesosphere/ai-navigator-app:v0.2.0
```
alembiewski, line 66: I don't think we can backport the agent - it might not work well with the version of Flux shipped with 2.12. We don't need to patch the agent for cluster-info-api, so I would suggest reverting these changes

alembiewski (commented):
We changed the version of flux CRD in the latest agent release, and I'm not sure it's compatible with the flux we shopped in 2.12. I suggest exclude the agent from the patch since there is no need to patch it, only the api

services/ai-navigator-app/0.2.8/ai-navigator-app.yaml
```diff
@@ -15,6 +15,9 @@ rules:
   - apiGroups: [""]
     resources: ["secrets", "nodes"]
     verbs: ["get", "list"]
+  - apiGroups: [ "apiextensions.k8s.io" ]
+    resources: [ "customresourcedefinitions" ]
+    verbs: ["get", "list"]
```
alembiewski, line 20: This is not required for the api, but for the app only - let's remove it

services/ai-navigator-cluster-info-agent/0.1.2/defaults/cm.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: ai-navigator-cluster-info-agent-0.1.1-d2iq-defaults
+  name: ai-navigator-cluster-info-agent-0.1.2-d2iq-defaults
   namespace: ${releaseNamespace}
 data:
   values.yaml: |
```
alembiewski, line 7: Let's rollback all the changes for `ai-navigator-cluster-info-agent` since they are not needed for the cluster info API patches

alembiewski (approved):

PR #2959, opened 2024-12-18
feat: upgrade to kubecost v2

takirala:

Upgrade kubecost to v2.

- [x] Depends on https://github.com/mesosphere/charts/pull/1529 for cosi helm chart
- [x] Depends on https://github.com/mesosphere/charts/pull/1534 for cosi bucket kit helm chart
- [ ] Depends on https://github.com/mesosphere/kommander/pull/5374 for attached clusters. Will be done in a follow up PR.

From 2.14.x, we no longer need centralized-kubecost as there is no need to run multiple instance of kubecost on management cluster. We will have just one `kubecost` service which is controlled via the `valuesFrom` field of kubecost `HelmRelease`. I have added comments where needed and there are some TODO's left in the code that will be follow up PRs. 

[NCN-104266](https://jira.nutanix.com/browse/NCN-104266)
[NCN-104267](https://jira.nutanix.com/browse/NCN-104267)

```release-note

```

**Checklist**


services/kubecost/metadata.yaml
```diff
@@ -2,7 +2,7 @@ displayName: Kubecost
 description: Provides real-time cost visibility and insights for teams using Kubernetes, helping organizations continuously reduce cloud costs.
 category:
   - monitoring
-type: partner
+type: platform
```
mhrabovcin, line 5: Why is this changing?

services/centralized-kubecost/2.5.0/cosi-storage/cosi-bucket.yaml
```diff
@@ -0,0 +1,35 @@
...
+metadata:
+  name: kubecost-cosi-storage
+  namespace: ${releaseNamespace}
+spec:
+  chart:
+    spec:
+      chart: cosi-bucket-kit
+      sourceRef:
+        kind: HelmRepository
+        name: takirala
```
mhrabovcin, line 12: revert?

services/kubecost/metadata.yaml
cwyl02: https://nutanix.slack.com/archives/C06A78Y063B/p1734635813445609?thread_ts=1734623086.853099&cid=C06A78Y063B

services/centralized-kubecost/2.5.0/cosi-storage/cosi-bucket.yaml
takirala: right, was waiting for https://github.com/mesosphere/charts/pull/1534 to merged. now that its merged, i dropped this custom helm repo reference.

common/helm-repositories/kubecost.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: source.toolkit.fluxcd.io/v1beta2
```
mhrabovcin, line 2: Can we use the latest api?

```suggestion
apiVersion: source.toolkit.fluxcd.io/v1
```

mhrabovcin (approved):

common/helm-repositories/kubecost.yaml
takirala: maybe we can but the rest of the helmrepository resources use v1beta2 so i ll just keep it as is for now to avoid any unexpected issues (like i am not sure if the version agnostic cli or the chart museum parser or the tooling that downloads helm charts etc., will be affected. most likely not, but i am being extra cautious here). we can update all helmrepos at once in future.

licenses.d2iq.yaml
```diff
@@ -454,13 +448,13 @@ resources:
       - license_path: LICENSE
         ref: ${image_tag%-distro}
         url: https://github.com/kiali/kiali
-  - container_image: quay.io/prometheus-operator/prometheus-config-reloader:v0.79.2
+  - container_image: quay.io/prometheus-operator/prometheus-config-reloader:v0.78.2
```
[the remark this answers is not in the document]
takirala: thats what i had initially, but then i realized we would need to keep bumping the image overrides everytime we bump the kubecost app. is that something that we can account for ? i feel like we end up with old image overrides that are no longer necessary or are out of date in the end if we do that. 

also, the upstream chart `main` branch has already moved to v0.79.2 so we ll get rid of this duplicate image eventually (unless of course if kps moves ahead again) so maybe this is fine as is ?

cwyl02 (approved):

licenses.d2iq.yaml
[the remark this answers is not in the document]
takirala: merging this for now, feel free to respond later and i can follow up in another pr.

PR #2973, opened 2024-12-27
feat: bump bloodhound to validate against 1.32.0

rajgitnutanix:
bumping 1.32.0 in commander-application for checking incompatible resources

https://jira.nutanix.com/browse/NCN-103535

### DIFF
--- a/.bloodhound.yml
+++ b/.bloodhound.yml
@@ -5,6 +5,7 @@ k8s_versions:
   - 1.29.9
   - 1.30.5
   - 1.31.1
+  - 1.32.9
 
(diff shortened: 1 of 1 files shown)


.bloodhound.yml
```diff
@@ -5,6 +5,7 @@ k8s_versions:
   - 1.29.9
   - 1.30.5
   - 1.31.1
+  - 1.32.0
```
mhrabovcin, line 8: Lets remove 1.28 and 1.29

rajgitnutanix: Sure, I will do that

PR #2994, opened 2025-01-07
fix: list image shell missing cronjob

cwyl02:
list-images.sh skips Cronjob images

https://jira.nutanix.com/browse/NCN-105004

tested running the script and saw the missing `docker.io/bitnami/kubectl:1.29.6` image

```release-note

```

**Checklist**

### DIFF
--- a/hack/list-images.sh
+++ b/hack/list-images.sh
@@ -129,7 +129,7 @@ done
 # These services use raw manifests rather than Helm charts so list the images directly from the manifests.
 # If more raw manifest services are added, then they should be added to the list of paths below.
 gojq --yaml-input --raw-output 'select(.kind | test("^(?:Deployment|Job|CronJob|StatefulSet|DaemonSet)$")) |
-                                .spec.template.spec |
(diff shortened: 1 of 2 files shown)


hack/list-images.sh
```diff
@@ -129,7 +129,7 @@ done
 # These services use raw manifests rather than Helm charts so list the images directly from the manifests.
 # If more raw manifest services are added, then they should be added to the list of paths below.
 gojq --yaml-input --raw-output 'select(.kind | test("^(?:Deployment|Job|CronJob|StatefulSet|DaemonSet)$")) |
-                                .spec.template.spec |
+                                (.spec.template.spec // .spec.jobTemplate.spec.template.spec) |
```
takirala, line 132: now you need to add the missing `docker.io/bitnami/kubectl:1.29.6` image in license manifest as this change is picking up the missing image 😄

takirala (approved):

hack/list-images.sh
cwyl02: done

PR #3013, opened 2025-01-14
feat: Move cloudnative-pg under harbor

jimmidyson:


takirala:
If we want to deploy postgresql by itself in the future, this rearchitecture would make that difficult (maybe we can, but i can't think of a way without some data loss).

mhrabovcin (approved):

PR #3022, opened 2025-01-16
feat: add monitoring to cnpg and default harbor db

tuxtof:

- configure CloudNativePG Operator to pass labels used by NKP Prom instance to select monitors
- enable monitoring on harbor databases


jimmidyson (approved):
Thanks!

PR #3027, opened 2025-01-16
fix: third-party apps should use the qualified certification

nataliepina:
Third-party applications should use 'qualified' instead of 'supported'.
Supported is now reserved for Nutanix supported apps - NKP Insights, NAI, etc. 

https://jira.nutanix.com/browse/NCN-105258

```release-note
CloudNativePG and Harbor applications will have a qualified badge and not a certified badge.
```

**Checklist**

### DIFF
--- a/services/cloudnative-pg/metadata.yaml
+++ b/services/cloudnative-pg/metadata.yaml
@@ -12,7 +12,6 @@ licensing:
   - Enterprise
 certifications:
   - qualified
-  - supported
   - airgapped
 overview: |-
   # Overview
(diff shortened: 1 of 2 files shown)


nataliepina:
> I agree an app shouldn't have both supported/qualified.
> 
> The trick with Nutanix registry is that we are adding extra configuration to make it HA, eg, deploying PG and Valkey(Redis) as dependencies, so this is not an app that we simply bundle.
> 
> I'm not sure where the boundary between Qualified and Supported apps lives.

Yeah, I am following the notes from product on this. There may need to be further discussion around this.

mhrabovcin (approved):

PR #3032, opened 2025-01-17
feat: enable monitoring and network policy for valkey

mhrabovcin:
Adds the `NetworkPolicy` and metrics to valkey installation.

```release-note

```

**Checklist**

### DIFF
--- a/services/harbor/1.16.0/valkey/valkey.yaml
+++ b/services/harbor/1.16.0/valkey/valkey.yaml
@@ -31,3 +31,14 @@ spec:
     - kind: ConfigMap
       name: harbor-valkey-overrides
       optional: true
+  postRenderers:
+    - kustomize:
+        patches:
+          - patch: |
+              - op: add
+                path: /metadata/labels/servicemonitor.kommander.mesosphere.io~1path
+                value: metrics
+            target:
+              kind: Service
+              name: harbor-valkey-metrics
+              version: v1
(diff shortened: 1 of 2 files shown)


services/harbor/1.16.0/valkey/valkey.yaml
```diff
@@ -31,3 +31,14 @@ spec:
     - kind: ConfigMap
       name: harbor-valkey-overrides
       optional: true
+  postRenderers:
```
tuxtof, line 34: @mhrabovcin why not directly use the `metrics.service.annotations` value in the helm chart ??

https://github.com/mesosphere/charts/blob/10f659401f8163ac03d8e811c3e29b00181b83b5/stable/valkey/templates/metrics-svc.yaml#L15

mhrabovcin: I think that `ServiceMonitor` selects endpoints by labels, `annotations` wouldn't work here.

jimmidyson (approved):

services/harbor/1.16.0/valkey/valkey.yaml
tuxtof: agree missread the sevice tempalte in valkey chart
only solution is `commonLabels` in the value , but will change it on all object

tuxtof (approved):

PR #3044, opened 2025-01-21
fix: add airgapped certification to istio, jaeger, and kiali

nataliepina:
Istio, Jaeger, and Kiali are available in airgapped environments.
Link to discussion - https://nutanix.slack.com/archives/C068HGFS19B/p1737181472843889

https://jira.nutanix.com/browse/NCN-105312

```release-note
Istio, Jaeger, and Kiali will have a badge indicating they are available in airgapped environments.
```

**Checklist**

### DIFF
--- a/services/istio/metadata.yaml
+++ b/services/istio/metadata.yaml
@@ -9,6 +9,7 @@ scope:
   - workspace
 certifications:
   - qualified
+  - airgapped
 licensing:
   - Pro
   - Ultimate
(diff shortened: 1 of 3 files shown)


guillermoz-nutanix (approved):
Is this change only cosmetic?

I mean, how we decide if an app would make it to the final airgapped bundle? Can't we use the same data to show the corresponding badge?

mhrabovcin (approved):
Thanks @nataliepina !

PR #3105, opened 2025-02-03
feat: bump new version postgres-exporter 0.16.0-debian-12-r3

rajgitnutanix:
Upgrades the following apps to use version 0.15.0-debian-12-r44 to 0.16.0-debian-12-r3

* postgres-exporter from 0.15.0-debian-12-r44 to 0.16.0-debian-12-r3


rajgitnutanix:
As per @mhrabovcin inputs there will be a new release of insights that will bring this fix to k-apps and there will be a new pr where the nkp-insights new release will be crated. so closing the PR for now .. 

Thank you

PR #3108, opened 2025-02-04
feat: update chartmuseum chart to 3.10.3

saikumar98419:
Upgrades the following apps to use version chart museum to 0.16.2:
* chart museum to 0.16.2


mhrabovcin (dismissed):
Can we bump the whole chart to 3.10.3 ? https://artifacthub.io/packages/helm/chartmuseum/chartmuseum

services/chartmuseum/3.10.3/defaults/cm.yaml
```diff
@@ -2,11 +2,14 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: chartmuseum-3.10.2-d2iq-defaults
+  name: chartmuseum-3.10.3-d2iq-defaults
   namespace: ${releaseNamespace}
 data:
   values.yaml: |
     priorityClassName: "dkp-critical-priority"
+    image:
```
takirala, line 10: you can drop this overrides and use the default upstream values.

services/chartmuseum/3.10.3/defaults/cm.yaml
```diff
@@ -2,11 +2,14 @@
...
+      repository: ghcr.io/helm/chartmuseum
+      tag: v0.16.2
```
takirala, line 12:
```suggestion
```

takirala:
@saikumar98419 For future reference, the helm chart `.tgz` is not expected to be simply renamed - we should fetch the latest tar file from upstream (like https://github.com/chartmuseum/charts/blob/gh-pages/index.yaml) and replace the old tar file with new tar file. I have done that and pushed the changes to this PR for now.

takirala (approved):

PR #3120, opened 2025-02-05
fix: update cosi & harbor metadata fields

kishankumarnutanix:
The category and the dependency field of metadata yaml of the cosi driver nutanix app and harbor app was set to empty array. However, we do not need dependency fields for either of the app and category needs to be general. Hence removing dependency field and adding category.
Slack Discussion Thread :- https://nutanix.slack.com/archives/C06A78Y063B/p1738775586636359?thread_ts=1738773226.964009&cid=C06A78Y063B

https://jira.nutanix.com/browse/NCN-105574

```release-note

```

**Checklist**

### DIFF
--- a/services/cosi-driver-nutanix/metadata.yaml
+++ b/services/cosi-driver-nutanix/metadata.yaml
@@ -1,7 +1,7 @@
 displayName: COSI Driver For Nutanix
 description: Seamlessly manage provisioning of object storage resources on Nutanix platforms within Kubernetes environments adhering to COSI standard.
-category: []
-dependencies: []
+category:
+  - general
 type: platform
 scope:
   - workspace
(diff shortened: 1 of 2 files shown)


jimmidyson (approved):

services/cosi-driver-nutanix/metadata.yaml
```diff
@@ -1,7 +1,7 @@
 displayName: COSI Driver For Nutanix
 description: Seamlessly manage provisioning of object storage resources on Nutanix platforms within Kubernetes environments adhering to COSI standard.
-category: []
-dependencies: []
+category:
+  - general
```
jimmidyson, line 4: Just to double check this should this be capitalised?

kishankumarnutanix: I checked with the other k-apps it should be small

PR #3134, opened 2025-02-11
docs: highlights kubecost v2 breaking change

cwyl02:
currently kubecost metadata overview doesn't include kubecost v2 breaking change

https://jira.nutanix.com/browse/NCN-105479

```release-note

```

**Checklist**

### DIFF
--- a/services/kubecost/metadata.yaml
+++ b/services/kubecost/metadata.yaml
@@ -17,6 +17,11 @@ overview: |-
 
   Today, Kubecost empowers more than 1,000 teams across companies of all sizes to monitor and reduce costs, while balancing cost, performance, and reliability. Kubecost is tightly integrated with the open source cloud native ecosystem and built for engineers and developers first, making it easy to drive adoption within your organization.
 
+  ## Kubecost V2 Breaking Changes  
+  Kubecost V2 now requires S3-compatible object storage to persist long-term data. If you do not enable rook-ceph-cluster AppDeployment in NKP, you would need to provide the endpoint to your own S3-compatible storage in order to use Kubecost. **Otherwise, Kubecost will not be able to start.**
+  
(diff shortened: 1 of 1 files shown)


services/kubecost/metadata.yaml
```diff
@@ -17,6 +17,11 @@ overview: |-
 
   Today, Kubecost empowers more than 1,000 teams across companies of all sizes to monitor and reduce costs, while balancing cost, performance, and reliability. Kubecost is tightly integrated with the open source cloud native ecosystem and built for engineers and developers first, making it easy to drive adoption within your organization.
 
+  ## Kubecost V2 Breaking Changes
```
takirala, line 20: fluent-bit has this wording:

```
## Notes
  Please ensure your S3-compatible object storage provider is deployed and has enough storage available.
  If deployed with the default option (rook-ceph), please ensure that the rook-ceph and rook-ceph-cluster applications are enabled in the workspace, and that rook-ceph-cluster is configured with sufficient storage space.
```

maybe we should be consistent and mention the dependency against two appdeployments instead of one (rook-ceph and rook-ceph-cluster) ?

cneth: Yeah, I agree with Tarun, I think the rook-ceph 'Notes' is better.    

How about this??

"## Notes 

Kubecost requires S3-compatible object storage to persist long-term data.

Please ensure your S3-compatible object storage provider is deployed and has enough storage available. 

To use the default S3-compatible object storage available in NKP (rook-ceph), ensure that the rook-ceph and rook-ceph-cluster applications are enabled in the workspace, and that the rook-ceph-cluster is configured with sufficient storage space.

For more information on configuring Kubecost to use other S3-compatible storage providers, see: https://docs.kubecost.com/install-and-configure/install/multi-cluster/long-term-storage-configuration.
"

cwyl02: updated

msdolbey (approved):
👍 
straight forward

PR #3142, opened 2025-02-12
fix: use fluentbit image from dockerhub

mhrabovcin:

```release-note

```

**Checklist**

### DIFF
--- a/services/fluent-bit/0.48.5/defaults/cm.yaml
+++ b/services/fluent-bit/0.48.5/defaults/cm.yaml
@@ -8,7 +8,7 @@ data:
     ---
     # overriding the default image tag to be consistent with logging-operator
     image:
-      repository: cr.fluentbit.io/fluent/fluent-bit
+      repository: fluent/fluent-bit
       tag: 3.2.4
 
     priorityClassName: "dkp-critical-priority"
--- a/services/fluent-bit/0.48.5/defaults/cm.yaml
+++ b/services/fluent-bit/0.48.5/defaults/cm.yaml
@@ -8,7 +8,7 @@ data:
     ---
     # overriding the default image tag to be consistent with logging-operator
     image:
-     # Pull from docker hub rather than cr.fluentbit.io to avoid rate limiting.
+      # Pull from docker hub rather than cr.fluentbit.io to avoid rate limiting.
       repository: docker.io/fluent/fluent-bit
       tag: 3.2.4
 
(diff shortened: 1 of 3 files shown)


services/fluent-bit/0.48.5/defaults/cm.yaml
```diff
@@ -8,7 +8,7 @@ data:
     ---
     # overriding the default image tag to be consistent with logging-operator
     image:
-      repository: cr.fluentbit.io/fluent/fluent-bit
+      repository: fluent/fluent-bit
```
jimmidyson, line 11: can we prefix it with `docker.io/` please?

services/logging-operator/5.0.1/defaults/cm.yaml
```diff
@@ -111,7 +111,7 @@ data:
       image:
         # Explicitly specify the version here. This should be updated when logging-operator is upgraded.
         # Also, update the image in fluent-bit configuration if the image is upgraded here.
-        repository: cr.fluentbit.io/fluent/fluent-bit
+        repository: fluent/fluent-bit
```
jimmidyson, line 114:
```suggestion
        repository: docker.io/fluent/fluent-bit
```

services/fluent-bit/0.48.5/defaults/cm.yaml
jimmidyson: And add a comment in why this deviates from upstream default.

mhrabovcin, line 11:
```suggestion
     # Pull from docker hub rather than cr.fluentbit.io to avoid rate limiting.
      repository: docker.io/fluent/fluent-bit
```

services/fluent-bit/0.48.5/defaults/cm.yaml
```diff
@@ -8,7 +8,8 @@ data:
     ---
     # overriding the default image tag to be consistent with logging-operator
     image:
-      repository: cr.fluentbit.io/fluent/fluent-bit
+     # Pull from docker hub rather than cr.fluentbit.io to avoid rate limiting.
+      repository: docker.io/fluent/fluent-bit
```
mhrabovcin, line 12:
```suggestion
      # Pull from docker hub rather than cr.fluentbit.io to avoid rate limiting.
      repository: docker.io/fluent/fluent-bit
```

PR #3145, opened 2025-02-12
chore: add priorityclass for new pods

takirala:

- harbor has critical priority so cloudnative-pg gets the same priority.
- kubecost has always been at high priority.

https://jira.nutanix.com/browse/NCN-105638
https://jira.nutanix.com/browse/NCN-105641

```release-note

```

**Checklist**


msdolbey (approved):
👍 
We are all Dwight

PR #3186, opened 2025-02-18
feat: bump k8s 1.32.2 version to bloodhound

rajgitnutanix:

bump k8s 1.32.2 to bloodhound 

https://jira.nutanix.com/browse/NCN-105344

### DIFF
--- a/.bloodhound.yml
+++ b/.bloodhound.yml
@@ -3,7 +3,7 @@
 k8s_versions:
   - 1.30.5
   - 1.31.4
-  - 1.32.0
+  - 1.32.2
 
 # paths to load (in that order)
 paths:


.bloodhound.yml
```diff
@@ -1,9 +1,9 @@
 # Kubernetes versions
 # (e.g. 1.23.0 - has to match versions in https://github.com/yannh/kubernetes-json-schema)
 k8s_versions:
-  - 1.30.5
   - 1.31.4
   - 1.32.0
```
mhrabovcin, line 5: Remove this line

.bloodhound.yml
```diff
@@ -1,10 +1,9 @@
 # Kubernetes versions
 # (e.g. 1.23.0 - has to match versions in https://github.com/yannh/kubernetes-json-schema)
 k8s_versions:
-  - 1.30.5
-  - 1.31.4
-  - 1.32.0
-
+ - 1.31.4
+ - 1.32.0
+ - 1.32.2
```
takirala, line 6: this should be a 1.33.x version. there is no point in having two 1.32.x versions.
We use this to future proof for 1.33.x bump.

takirala: looks like i got ahead of myself - there is no 1.33.x yet!

lets remove duplicate 1.32.x versions and retain the 1.30.x here.

rajgitnutanix: @takirala  As disciussed - 1.30.5
  - 1.31.4
  - 1.32.2 these version I am adding on it thank you for confirmation

takirala (approved):

PR #3197, opened 2025-02-20
chore: make the kubectl image tag consistent

takirala:

Adjusting the kubectl image to be consistent with what we add in the bundle https://github.com/mesosphere/kommander-applications/blob/release-2.14/licenses.d2iq.yaml#L524-L528 

```release-note

```

**Checklist**


msdolbey (approved):
👍 
shhhh

PR #3206, opened 2025-02-24
fix: gpu dashboard temp graph

tuxtof:

in GPU dashboard the GPU Temperatur is not displaying correctly, we need to remove the instant flag

image

also add some tags on the dashboard for cosmetic

image


faiq (approved):

tuxtof:
@jimmidyson something to backport ??

PR #3207, opened 2025-02-24
chore: bumping dex image

msdolbey:

Bumping dex to include a11y changes and some other recent theme changes
https://github.com/mesosphere/dex/tree/refs/tags/v2.41.1-d2iq.2

n/a

solely bumping dex image

n/a - ui changes

**Checklist**


takirala (approved):

takirala:
We already bumped dex version since 2.13.1 so its safe to bump the image version without bumping dex version.

PR #3225, opened 2025-02-27
feat: bump kubecost to 2.6.3

takirala:

Bumps kubecost to latest version

https://github.com/kubecost/cost-analyzer-helm-chart/releases/tag/v2.6.3

```release-note

```

**Checklist**


licenses.d2iq.yaml
```diff
@@ -470,7 +470,7 @@ resources:
         notice_path: NOTICE
         ref: ${image_tag}
         url: https://github.com/prometheus/node_exporter
-  - container_image: quay.io/prometheus/prometheus:v2.55.1
+  - container_image: quay.io/prometheus/prometheus:v3.2.0
```
cwyl02, line 473: let's 🤞 no bad breaking changes

cwyl02 (approved):

licenses.d2iq.yaml
takirala: Yes I wanted to test sth with prom v3 and that's kinda main reason why I am doing this! Fwiw this is tested upstream so at least kubecost should be fine. I will keep an eye out for other impact if any.

PR #3289, opened 2025-03-20
feat: Introduce cilium-hubble-relay-traefik

dlipovetsky:
This deploys the helm chart that configures a traefik route for the Cilium Hubble Relay service.

Adds an install apptest for the service, and an upgrade test that is skipped, for now, because no previous version of the service exists, and the upgrade test framework requires a previous version.

The install apptest verifies that the IngressRouteTCP created by cilium-hubble-relay-traefik does correctly route requests to the hubble-relay service with TLS passthrough. The test:
- Installs traefik and its dependencies
- Installs cilium-hubble-relay-traefik
- Runs an HTTPS server that presents a self-signed server cert with a specific CN.
- Routes the hubble-relay Service to that HTTPS server.
- Verifies that a request to traefik using the expected SNI is routed to the service, and that server cert has the expected CN.
-     

https://jira.nutanix.com/browse/NCN-105953

This service is internal.

```release-note

```

**Checklist**

### DIFF
--- a/services/cilium-hubble-relay-traefik/metadata.yaml
+++ b/services/cilium-hubble-relay-traefik/metadata.yaml
@@ -1,6 +1,9 @@
 type: internal
 scope:
   - workspace
+requiredDependencies:
+  - traefik
+allowMultipleInstances: false
 licensing:
   - Pro
   - Ultimate
--- a/services/cilium-hubble-relay-traefik/metadata.yaml
+++ b/services/cilium-hubble-relay-traefik/metadata.yaml
@@ -1,0 +1,8 @@
+type: internal
+scope:
+  - workspace
+licensing:
+  - Pro
+  - Ultimate
+  - Essential
+  - Enterprise
--- a/.github/service-labeler.yaml
+++ b/.github/service-labeler.yaml
@@ -18,6 +18,10 @@ services/chartmuseum:
 - changed-files:
   - any-glob-to-any-file:
     - services/chartmuseum/**
+services/cilium-hubble-relay-traefik:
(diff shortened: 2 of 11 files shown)


services/cilium-hubble-relay-traefik/metadata.yaml
```diff
@@ -0,0 +1,8 @@
+type: internal
```
takirala, line 1: Its `internal` so not really matters ig but lets add 
```
requiredDependencies
- traefik
```
or atleast 
```
dependencies:
  - traefik
```
if we want to keep things loosely coupled.

We can also add 
```
allowMultipleInstances: false
```
for future consumption.

takirala (approved):

services/cilium-hubble-relay-traefik/metadata.yaml
dlipovetsky: Thanks!

dlipovetsky: Since the app creates a resource defined by a traefik CRD, the `requiredDependencies` makessense.

dlipovetsky: Fixed in https://github.com/mesosphere/kommander-applications/pull/3289/commits/a16cf884069f9ba4bd619e1feb52f5ef5a902440

dlipovetsky:
Waiting for an answer on whether to deploy for all licenses

dlipovetsky:
Manifest validation job failed without any informative error. It passes locally. Re-running.

PR #3295, opened 2025-03-21
fix: preserve user labels on flux ns during upgrades

takirala:

Whenever the `flux-configuration` `Kustomization` in `kommander` namespace is reconciled, we overwrite any user given labels on flux namespace. This breaks proxied environments cuz our docs currently ask user to label flux namespace manually. We should fix this manual ask from the user in long run but until then, we should use `merge` strategy in flux ssa. 

I am not bumping the flux component version here cuz this change is only limited to adding an annotation on namespace and does not cause any restarts or such.

This was first added in https://github.com/fluxcd/kustomize-controller/blob/main/CHANGELOG.md#0220 so we can safely backport it all way back to 2.8.x (which uses a 1.x.y version of kustomize-controller)

https://jira.nutanix.com/browse/NCN-106215

```release-note

```

**Checklist**

### DIFF
--- a/hack/flux/flux-update-kustomization.yaml
+++ b/hack/flux/flux-update-kustomization.yaml
@@ -4,6 +4,15 @@ resources:
   - flux.yaml
 patches:
   - target:
+      kind: Namespace
+      name: kommander-flux
+    patch: |-
+      - op: add
+        path: /metadata/annotations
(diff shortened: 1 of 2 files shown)


msdolbey (approved):
👍 
flux it

PR #3328, opened 2025-03-31
fix(kps): Missing node-exporter rbac proxy rule

gracedo:
follow up to https://github.com/mesosphere/kommander-applications/pull/2624
https://github.com/prometheus-community/helm-charts/issues/3338
Was fixed in KPS 57.1.1 https://github.com/prometheus-community/helm-charts/pull/3946 which is higher than the version of KPS in 2.8, so manually adding the rule via default values.

After applying the rule, prometheus was able to access node metrics and error disappeared from UI

```release-note

```

**Checklist**

### DIFF
--- a/services/kube-prometheus-stack/48.3.3/defaults/cm.yaml
+++ b/services/kube-prometheus-stack/48.3.3/defaults/cm.yaml
@@ -43,6 +43,12 @@ data:
           - name: grpc
             port: 10901
             targetPort: grpc
+      additionalRulesForClusterRole:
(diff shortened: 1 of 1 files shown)


takirala (approved):

takirala:
Is this fix present in 2.12 versions already via upstream chart or does this need backporting ?

gracedo:
> Is this fix present in 2.12 versions already via upstream chart or does this need backporting ?

@takirala our charts in 2.12+ include the fix!

PR #3347, opened 2025-04-01
fix(kps): Use correct metrics path for grafana-logging

gracedo:
image

grafana-logging metrics is behind the path `/dkp/logging/grafana/metrics` not `/metrics`. needs https://github.com/mesosphere/kommander-cli/pull/1592 as well, and forwardports to newer versions

image

```release-note

```

**Checklist**


mikolajb (approved):
It would be good to create a test for that path 🙏🏻

PR #3368, opened 2025-04-03
fix: CVE-fix k8s-sidecar:1.30.3

mohammad-irfan3:
Upgrades the following apps versions

* k8s-sidecar from version 1.28.0 to 1.30.3
* Kube-prometheus-stack 69.8.0 to 70.4.2

Jira:https://jira.nutanix.com/browse/NCN-106684


services/centralized-grafana/69.8.0/defaults/cm.yaml
mhrabovcin, line 1: When changing the default configuration then the app version needs to be bumped. Please consider making the bump in these services as part of app bump itself.

mohammad-irfan3: ok, I'll do as suggested

mhrabovcin (approved):

PR #3372, opened 2025-04-04
feat: kommander charts use oci

cwyl02:
currently the kommander charts are still pushed to chartmuseum, which we are going to move off from.

https://jira.nutanix.com/browse/NCN-104792

```release-note

```

**Checklist**


services/kommander-appmanagement/0.16.0/kommander-appmanagement.yaml
```diff
@@ -1,18 +1,25 @@
 ---
+apiVersion: source.toolkit.fluxcd.io/v1beta2
+kind: OCIRepository
+metadata:
+  name: kommander-appmanagement-chart
+  namespace: kommander-flux
```
mikolajb, line 6: I'm wondering if it should go to the `${releaseNamespace}`

services/kommander-appmanagement/0.16.0/kommander-appmanagement.yaml
```diff
@@ -1,18 +1,25 @@
...
+spec:
+  interval: 1m
+  url: "${ociRegistryURL:=oci://docker.io/mesosphere/kommander-appmanagement-chart}"
```
mikolajb, line 9:
```suggestion
  url: "${ociRegistryURL:=oci://docker.io/mesosphere}/kommander-appmanagement-chart"
```

services/kommander/0.16.0/kommander.yaml
```diff
@@ -1,4 +1,15 @@
 ---
+apiVersion: source.toolkit.fluxcd.io/v1beta2
+kind: OCIRepository
+metadata:
+  name: kommander-chart
+  namespace: kommander-flux
+spec:
+  interval: 1m
+  url: "${ociRegistryURL:=oci://docker.io/mesosphere/kommander-chart}"
```
mikolajb, line 9:
```suggestion
  url: "${ociRegistryURL:=oci://docker.io/mesosphere}/kommander-chart"
```

services/kommander-appmanagement/0.16.0/kommander-appmanagement.yaml
cwyl02: done

cwyl02:
currently the Go Test has to fail because we are introducing this new OCIRepo object, where the [test code](https://github.com/mesosphere/kommander-applications/blob/main/hack/release/pkg/chartversion/chartversion_test.go#L50-L57) is comparing to HelmRelease. In this PR particularly, its should compare an HR against an OCIRepo, but it will be OCIRepo vs OCIRepo moving forward. 

created https://jira.nutanix.com/browse/NCN-107233 

We have to admin merge this

mikolajb (approved):
Do the pre and post release scripts work?

cwyl02:
> Do the pre and post release scripts work?

hmm the version bump works but it also change the URL and the $ variable got dropped.. let me take a look

services/kommander-appmanagement/0.16.0/kommander-appmanagement.yaml
```diff
@@ -1,18 +1,25 @@
...
   namespace: ${releaseNamespace}
 spec:
-  chart:
-    spec:
-      chart: kommander-appmanagement-chart
-      sourceRef:
-        kind: HelmRepository
-        name: mesosphere.github.io-kommander-charts
-        namespace: kommander-flux
-      version: "${kommanderChartVersion:=v2.16.0-dev}"
```
Indrajeetsi, line 15: Hi @kaiwalyajoshi , 

UI was reading NKP version from kommander-chart. Because of this change, in 2.16 is not available anymore.

PR #3409, opened 2025-04-15
feat: update the valkey-sentinel:8.1.0

saikumar98419:
update the valkey-sentinel to latest

* update to 8.1.0


services/harbor/1.16.2/defaults/valkey.yaml
mhrabovcin, line 1: For the future - The harbor app version directory needs to be bumped when some of the default configuration changes. I think this will be done when @mohammad-irfan3 will do the overall harbor bump so we can skip it for this specific PR.

PR #3423, opened 2025-04-16
chore: make testapp module compatible to be used by other projects

NRoshanRai:


apptests/environment/environment.go
```diff
@@ -59,17 +64,25 @@ func (e *Env) Provision(ctx context.Context) error {
...
 		return err
 	}
+	fmt.Printf("===applied calico yaml : env : %+v\n", e)
 
 	subnet, err := e.Network.Subnet()
 	if err != nil {
 		return err
 	}
-	_ = InstallMetallb(ctx, e.Cluster.KubeconfigFilePath(), subnet)
+	fmt.Println("=== subnet")
```
NRoshanRai, line 83: remove this

apptests/environment/environment.go
```diff
@@ -437,3 +450,21 @@ func applyYAMLFile(ctx context.Context, genericClient genericCLient.Client, path
...
+	content, err := calicoFS.ReadFile("calico.yaml")
+	if err != nil {
+		return "", err
+	}
+
+	// Write to a temporary file
+	tmpDir := os.TempDir()
+	tmpFile := filepath.Join(tmpDir, "calico.yaml")
+
+	if err := os.WriteFile(tmpFile, content, 0644); err != nil {
```
NRoshanRai, line 465: what will happen if file already exists ?

apptests/environment/prerequisites.go
```diff
@@ -108,3 +114,26 @@ func NewClient(kubeConfigPath string) (client.Client, error) {
...
+	// Write to a temporary file
+	tmpDir := os.TempDir()
+	tmpFile := filepath.Join(tmpDir, filePath)
+
+	// Ensure parent directories exist
+	if err := os.MkdirAll(filepath.Dir(tmpFile), 0755); err != nil {
+		return "", err
+	}
+
+	if err := os.WriteFile(tmpFile, content, 0644); err != nil {
```
NRoshanRai, line 134: what will hapen if file already exists ?

apptests/environment/environment.go
NRoshanRai: it will replace the content of the file

apptests/environment/prerequisites.go
NRoshanRai: .

apptests/kind/scripts/loopback-storage-creator.sh
```diff
@@ -0,0 +1,138 @@
...
+set -euox pipefail
+
+CI=${CI:-"true"}
+
+ensure_prerequisites() { (
+  # wait for hack/kind/gha/create-loop-devs_daemonset.yaml to finish running
+  # currently, the e2e tests run the loopback creator daemonset if the CI env var is "true"
+  echo "is this running in CI?: ${CI}"
+  if [ "${CI}" == "true" ]; then
+    timeout 5m bash -c "while ! test -e /dev/loop100; do echo 'waiting for loopback devices to be created'; sleep 5; done"
```
shivendra-ntnx, line 12: Are these paths are compatible with all envs where this script is expected to run?

shivendra-ntnx (commented):
Just make sure, whatever temp folders are you creating, those should be cleanup incase of failures etc.

apptests/environment/environment.go
```diff
@@ -437,3 +448,21 @@ func applyYAMLFile(ctx context.Context, genericClient genericCLient.Client, path
 
 	return nil
 }
+
+// GetCalicoYAMLPath writes the embedded YAML to a temp file and returns the path.
+func GetCalicoYAMLPath() (string, error) {
```
takirala, line 453:
```suggestion
func getCalicoYAMLPath() (string, error) {
```

takirala (commented):
Similar to 
```
ApplyYAMLWithoutSubstitutions
ApplyYAML
```
functions, what do we think of creating a new function like
```
ApplyYAMLRaw
```
that accepts a `[]byte` content instead of `string` paths.

This would remove the need for calico yaml tempory directory (and paves way for future yaml manifests too).

Metallb is a bit difficult (but maybe we can switch to `oci://<>` in the future instead of that tar bundle.

apptests/environment/environment.go
```diff
@@ -59,8 +64,13 @@ func (e *Env) Provision(ctx context.Context) error {
...
 	e.SetK8sClient(k8sClient)
 	e.SetCluster(cluster)
+
 	// install calico CNI
-	err = e.ApplyYAML(ctx, "../environment/calico.yaml", nil)
+	calicoYamlFIlePath, err := GetCalicoYAMLPath()
+	if err != nil {
+		return err
+	}
+	err = e.ApplyYAML(ctx, calicoYamlFIlePath, nil)
```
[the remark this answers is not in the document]
NRoshanRai: removing this approach , instead using the emded file via bytes

apptests/kind/scripts/loopback-storage-creator.sh
NRoshanRai: these are path inside of a container os

NRoshanRai: this script is used , only while a rook-ceph installation is done

shivendra-ntnx (approved):

NRoshanRai:
have gone with ApplyYAMLRaw approach where the yaml files content are passed instead of file path

PR #3431, opened 2025-04-18
feat: bump rook-ceph to 1.17.0

shubham2g:
feat: bump rook-ceph to 1.17.0

https://jira.nutanix.com/browse/NCN-107129


services/rook-ceph/1.17.0/defaults/cm.yaml
```diff
@@ -1,11 +1,20 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: rook-ceph-1.16.2-d2iq-defaults
+  name: rook-ceph-1.17.0-d2iq-defaults
   namespace: ${releaseNamespace}
 data:
   values.yaml: |
     ---
+    image:
```
mhrabovcin, line 9: Why do we need this section?

shubham2g: I noticed that during license generation, the image tag was being pulled as 'master' instead of version 1.17.0. After checking the values.yaml in the Helm chart, I updated it accordingly.

mhrabovcin (approved):

PR #3433, opened 2025-04-19
feat: enable persistence for grafana

tuxtof:

Many customers have complained that after an upgrade, they have lost their custom Grafana dashboards or customized configurations. 

This PR enables persistence for the Grafana app using the default settings.

### DIFF
--- a/services/kube-prometheus-stack/70.4.2/defaults/cm.yaml
+++ b/services/kube-prometheus-stack/70.4.2/defaults/cm.yaml
@@ -341,6 +341,13 @@ data:
       enabled: true
       defaultDashboardsEnabled: true
       priorityClassName: "dkp-critical-priority"
+      persistence:
+        enabled: true
+      initChownData:
+        image:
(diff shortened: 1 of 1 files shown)


services/kube-prometheus-stack/70.4.2/defaults/cm.yaml
```diff
@@ -341,6 +341,8 @@ data:
       enabled: true
       defaultDashboardsEnabled: true
       priorityClassName: "dkp-critical-priority"
+      persistence:
+        enabled: true
```
mhrabovcin, line 345: Enabling this runs an additiona init pod for adjusting the data ownership. Can we add values for using existing `kubectl` image - https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L449-L463 

Otherwise we'd have to bring another busybox image to the airgapped bundle.

mhrabovcin (approved):

PR #3446, opened 2025-04-24
refactor: break up apptests module

takirala:

Breaking up the apptests module into two:
1. Without ginkgo where we can't really run anything and is solely meant for setting up test framework
2. With all the app specific scenarios that actually handles running a specific test

We need to refactor a lot to make the helpers more modular and consumable. This is just a preliminary refactor but still allows us to import appscenarios as a reusable module. 

```release-note

```

**Checklist**


apptests/appscenarios/traefik.go
```diff
@@ -69,6 +93,7 @@ func (t traefik) install(ctx context.Context, env *environment.Env, appPath stri
 		}
 
 		// Apply defaults for gateway-api-crds
+		fmt.Println("=== default k : ",  filepath.Join(gatewayCRDsPath, "/defaults"))
```
takirala, line 96: need to clean this up ?

shivendra-ntnx (approved):

PR #3448, opened 2025-04-24
refactor: Appscenerio interface add 2 new methods

NRoshanRai:
adding methods to interface to be used by other modules

```release-note

```

**Checklist**

### DIFF
--- a/apptests/appscenarios/appscenarios.go
+++ b/apptests/appscenarios/appscenarios.go
@@ -16,6 +16,8 @@ import (
 	"github.com/mesosphere/kommander-applications/apptests/environment"
 )
 
+var upgradeKAppsRepoPath string
+
 // AppScenario defines the behavior and name of an application test scenario
 type AppScenario interface {
 	Name() string                                    // scenario name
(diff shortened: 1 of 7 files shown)


apptests/appscenarios/appscenarios.go
```diff
@@ -16,10 +16,14 @@ import (
...
 )
 
+var upgradeKAppsRepoPath string
+
 // AppScenario defines the behavior and name of an application test scenario
 type AppScenario interface {
-	Install(context.Context, *environment.Env) error // logic implemented by a scenario
 	Name() string                                    // scenario name
+	Install(context.Context, *environment.Env) error // logic implemented by a scenario
+	InstallPreviousVersion(ctx context.Context, env *environment.Env) error
```
shivendra-ntnx, line 25: Check if it really needed or Install() can do the job, rest LGTM.

shivendra-ntnx (approved):

apptests/appscenarios/appscenarios.go
NRoshanRai: this method is already in use by each kommander application, Install could be used to achive this behaviour, but that would require a change in atleast 25 files which have their own implementation of InstallPreviousVersion method

PR #3457, opened 2025-04-28
feat(apps): Copy metadata.yaml to <app>/<version>/

gracedo:
From 2.16 we will be looking at metadata.yaml files at the `/app/version/` level (will still support files at top level `/app/` for backwards-compatibility for existing catalogs but will eventually deprecate it). There will be a follow-up PR to remove the app level metadata.yaml once the kommander controller is updated to support parsing files in the app version dir

https://jira.nutanix.com/browse/NCN-107224

```release-note

```

**Checklist**


gracedo:
failing tests not related to this change

PR #3495, opened 2025-05-06
refactor: reorder defaultProApps to deploy ai navigator towards the end.

takirala:

`deploy/kommander-licensing-cm ` is failiing in airgapped environments with the following error:

```
2025-05-06T05:52:00.787Z	ERROR	controllers.LicenseProAppDeployment	Reconciler error	{"reconcileID": "7fba97b1-dd1e-417c-ae7d-b2761b13d80c", "error": "unable to get latest version of app ai-navigator-app: no ClusterApps found for app ai-navigator-app"}
```

This is because ai-navigator-app is excluded in airgapped artifacts. As an interim fix for 2.15, we can move it down the list so that other apps are deployed and then the controller will fail on the last app. This has been the behavior since 2.12.0 - https://github.com/mesosphere/kommander-applications/blob/v2.12.0/services/kommander/0.12.0/defaults/cm.yaml#L80

This is an interim fix for 2.15 and we will figure out an ideal fix for 2.16 later.

https://jira.nutanix.com/browse/NCN-107450

```release-note

```

**Checklist**

### DIFF
--- a/services/kommander/0.15.0/defaults/cm.yaml
+++ b/services/kommander/0.15.0/defaults/cm.yaml
@@ -75,7 +75,6 @@ data:
       - "thanos"
       - "nkp-insights-management"
       defaultProApps:
-      - "ai-navigator-app"
       - "grafana-logging"
       - "grafana-loki"
       - "kube-prometheus-stack"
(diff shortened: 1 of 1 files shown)


msdolbey (approved):
👍 
elegant - class class class

gracedo (approved):
😬

PR #3512, opened 2025-05-13
fix: use pull_request target for GHA

mhrabovcin:
Don't use `pull_request_target` for GHA.

https://jira.nutanix.com/browse/NCN-107016

There are 2 workflows updates.

`automerge.yaml`: the `pull_request_target` trigger couldn't work properly as the token is missing the `pull-request: write` permission necessary for approving a PR.

`backport.yaml`: implication will be that it won't be possible to run backport action on PRs that are coming from a fork.

```release-note

```

**Checklist**

### DIFF
--- a/.github/workflows/automerge.yaml
+++ b/.github/workflows/automerge.yaml
@@ -1,6 +1,6 @@
 name: "Automerge"
 on:
-  pull_request_target:
+  pull_request:
     types:
       - labeled
       - unlabeled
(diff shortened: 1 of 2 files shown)


fatz (approved):

gracedo (approved):

gracedo:
do we need to backport this 🤔

mhrabovcin:
@gracedo I had a same question and because of the nature of the executed GHA there is no need to backport this change.

PR #3514, opened 2025-05-13
feat: bump k8s 1.33.0 version to bloodhound

SandhyaRavi2403:
bump k8s 1.33.0 version to bloodhound

https://jira.nutanix.com/browse/NCN-106613

```release-note

```

**Checklist**


.bloodhound.yml
```diff
@@ -4,6 +4,7 @@ k8s_versions:
   - 1.30.10
```
mhrabovcin, line 4: We can remove 1.30.10

PR #3524, opened 2025-05-15
feat: Use helm in v2

mikolajb:
The title says it all

```release-note

```

**Checklist**


gracedo (approved):
Did we decide how to handle v2beta2 -> v2 interim step?

cwyl02 (approved):

gracedo (changes requested):
Let's check the diff between these app versions now vs 2.15.0 and make sure that if it's the same in 2.15.0 we bump the minor version here. I think for many apps there will be upcoming bumps (lots of chart bump PRs in flight) but let's just make sure to cover the apps that don't end up having a bump

mikolajb:
> Let's check the diff between these app versions now vs 2.15.0 and make sure that if it's the same in 2.15.0 we bump the minor version here. I think for many apps there will be upcoming bumps (lots of chart bump PRs in flight) but let's just make sure to cover the apps that don't end up having a bump

@gracedo I think we will try to do it without the version bump

PR #3542, opened 2025-05-22
feat: adds mig manager to airgap bundle

faiq:
Adds mig manager image

```release-note

```

**Checklist**

### DIFF
--- a/licenses.d2iq.yaml
+++ b/licenses.d2iq.yaml
@@ -358,6 +358,11 @@ resources:
       - license_path: LICENSE
         ref: ${image_tag%-ubuntu22.04}
         url: https://github.com/NVIDIA/dcgm-exporter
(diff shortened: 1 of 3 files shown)


services/nvidia-gpu-operator/25.3.0/defaults/cm.yaml
```diff
@@ -43,3 +43,8 @@ data:
     operator:
       repository: nvcr.io/nvidia
       version: v25.3.0
+    migManager:
```
mhrabovcin, line 46: Because this is an operator our tooling cannot discover some images only by static analysis. For such cases we maintain list of additional images that are included in the bundle.

Can you add additional images to this list https://github.com/mesosphere/kommander-applications/blob/main/services/nvidia-gpu-operator/25.3.0/helmrelease/extra-images.txt

PR #3550, opened 2025-05-26
refactor: Switch HelmReleases to use OCI registries

shivendra-ntnx:
Migrated Helmreleases to use the OCI registry

https://jira.nutanix.com/browse/NCN-107634

Currently migrating those which are publically available in OCI.


.github/workflows/validate-licenses.yaml
```diff
@@ -37,6 +37,7 @@ jobs:
         run: |
           export PATH=${PWD}/.local/bin:$PATH
           devbox run -- helm plugin install https://github.com/d2iq-labs/helm-list-images
+          devbox run -- helm plugin install https://github.com/nikhilsbhat/helm-images
```
mikolajb, line 40:
```suggestion
          devbox run -- helm plugin install https://github.com/nikhilsbhat/helm-images
```

Do we need both?

services/thanos/15.8.0/defaults/cm.yaml
```diff
@@ -9,7 +9,7 @@ data:
     image:
       registry: docker.io
       repository: bitnami/thanos
-      tag: 0.38.0-debian-12-r0
+      tag: 0.36.1-debian-12-r3
```
mikolajb, line 12:
```suggestion
      tag: 0.38.0-debian-12-r0
```

Do we need this change?

hack/list-images.sh
```diff
@@ -162,15 +162,32 @@ for dir in $(find . -path "./apptests/*" -prune -o -type f -name "*.yaml" -print
...
-      >&2 echo
+          else
+            empty
+          end
+        else
+          empty
+        end' | \
+      while IFS= read -r line; do
+        if [[ "$line" == *"mesosphere"* ]]; then
+          >&2 echo "Skipping mesosphere chart: $line"
```
mikolajb, line 181: Why to skip mesosphere charts?

mikolajb (approved):

.github/workflows/validate-licenses.yaml
shivendra-ntnx: I haven't changed the older workflow, I will see if we can use just one.

services/thanos/15.8.0/defaults/cm.yaml
shivendra-ntnx: Yes based on docummentation https://artifacthub.io/packages/helm/bitnami/external-dns/8.5.0
OCI helm repo uses this image docker.io/bitnami/thanos:0.36.1-debian-12-r3

hack/list-images.sh
shivendra-ntnx: kommander-appmanagement-chart chart is breaking. For now added exception.

Error: exit status 1: Error: Unable to locate any tags in provided repository: oci://docker.io/mesosphere/kommander-appmanagement-chart

apptests/appscenarios/gatekeeper_test.go
```diff
@@ -213,16 +213,10 @@ func ensureConstraintEnforced(projectNS string) {
...
-					Chart:   "external-dns",
-					Version: "7.2.0",
-					SourceRef: fluxhelmv2beta2.CrossNamespaceObjectReference{
-						Kind:      "HelmRepository",
-						Name:      "charts.github.io-bitnami",
-						Namespace: "kommander-flux",
-					},
-				},
+			ChartRef: &fluxhelmv2beta2.CrossNamespaceSourceReference{
+				Kind:      "HelmChartTemplate",
```
cwyl02, line 217: this should change to use `OCIRepository`

services/external-dns/8.5.0/defaults/cm.yaml
```diff
@@ -7,7 +7,7 @@ data:
   values.yaml: |-
     priorityClassName: dkp-high-priority
     image:
-      tag: 0.16.1-debian-12-r5
+      tag: 0.15.0-debian-12-r2
```
cwyl02, line 10: should we really downgrade this? i see 0.16.1 is used to fix CVE https://github.com/mesosphere/kommander-applications/pull/3451

services/thanos/15.8.0/defaults/cm.yaml
mikolajb: It was set to the higher image version for a reason probably

services/external-dns/8.5.0/defaults/cm.yaml
mikolajb: Let's absolutely not downgrade it 🙏🏻 https://github.com/mesosphere/kommander-applications/pull/3550#discussion_r2112190620

hack/list-images.sh
shivendra-ntnx: With this helm plugin install https://github.com/nikhilsbhat/helm-images, 
When I am trying to list images of  kommander-appmanagement OCIRepository It's throwing this error, Although successful listing image for bitnami ones.

services/external-dns/8.5.0/defaults/cm.yaml
shivendra-ntnx: But My understanding was for each Repo version, image mapping is one to one.
If you check this it will clear the things out.
https://artifacthub.io/packages/helm/bitnami/external-dns/8.5.0?modal=values&path=image

As you suggest, I will not downgrade it.

mikolajb: Generally we favour default images (so no image spec in the config) but sometimes we have to manually set it to, for example, fix s CVE.

hack/list-images.sh
mikolajb: Like York suggested, I think you should extract `v2.16.0-dev` from `${kommanderChartVersion:=v2.16.0-dev}`

shivendra-ntnx: I have made the changes accordingly, But that doesn't seems to work locally on my setup, As mentioned above.

apptests/appscenarios/externaldns_test.go
```diff
@@ -136,6 +136,8 @@ var _ = Describe("External DNS Tests", Label("external-dns"), func() {
 
 		It("should install the previous version successfully", func() {
 			ed = externalDns{}
+			// Only needed for 2.16 version as its been migrated to use Bitnami OCIRepository
```
cwyl02, line 139: why do we still need to apply the HelmRepository?

shivendra-ntnx: The behavior of this test involves checking out a previous version, specifically v2.15.1-dev, and performing an installation of that version. In that version, the HelmRepository was expected to be present via common/kustomization.

However, in the current codebase, we have removed the Helm repositories from common/helm-repositories/, so they are no longer being applied during the initial Describe block of the test.

As a result, when the test attempts to install v2.15.1-dev, it fails to locate the required HelmRepository. To maintain compatibility with the 2.15 version installation during testing, we need to explicitly patch and apply the HelmRepository.

If we intend to change the test behavior (e.g., not supporting older version installs or modifying the install flow), I'm happy to discuss alternatives.

PR #3551, opened 2025-05-27
feat: modify types and certifications for apps

akshatjohar:

According to Figma mockups - https://www.figma.com/design/MGF4QAB843gaZtmZ9L9NhY/Applications---Partner-Catalog-and-more?node-id=0-1&p=f&t=gSSBhsrs0zrVPee4-0, new types and tags are being introduced as follows for various types of application that can exist (for both 2.16 and beyond) 

image

Following matrix (after discussion with PM and Design) will be used to modify metadata.yaml files for existing applications and also would be taken as reference for future applications so that configuration based approach works in the UI

image

I am not exactly sure at what areas does the backend team require the `airgapped` tag , hence I have kept it optional (according to usecase) in above matrix. But in the UI , we wouldn't show the `airgapped` tag even if we do receive it from the backend , which is part of this feature

**Backward Compatibility for types and certifications**

For applications that are not upgraded as part of `kommander-upgrade` process , the data for those will still be extracted from older yamls (which don't get the new types/certifications) that we introduce now. Hence inside the UI repo , we plan to create a mapping technique to read the older values as newer ones so that filtering and display of tags is consistent

- `platform` would be read as `nkp-core-platform`
- `catalog` would be read as `nkp-catalog`
- `supported` would be read as `nutanix-supported`
- `qualified` would be read as `lifecycle-managed`
- airgapped remains as it is , but won't be displayed as a tag in the UI

Also , in this PR we are also changing the type of `Knative` from `catalog` to `nkp-core-platform` as in the UI we need to recategorise it as a platform app after discussions with PM

https://jira.nutanix.com/browse/NCN-107678

N/A

```release-note
As mentioned in figma mockups , we are introducing new types and tags in the UI as part of this feature
```

**Checklist**


guillermoz-nutanix:
> we are introducing new types in the UI and following is now the expected behaviour:-
> 
> * Nutanix supported apps filtered out as `Nutanix Apps`
> * platform and catalog apps filtered out as `Certified Open Source Apps`
> * partner apps filtered out as `Preferred Partner Apps`

This are the types that UI recognizes:

https://github.com/mesosphere/kommander-ui/blob/d6a8a49b262cc0f13140011baea4d989e8512b3e/packages/frontend/src/constants/applications.ts#L163-L173

And these are the ones that UX wants (4 items):
image

What would happen to the other types?
How would this work when a user upgrades a cluster? Will the old versions of the apps still show with the old type?

msdolbey:
> > As a result , we need to change the type of nutanix supported apps from `platform` to `nutanix` to differentiate them from Open Source Apps and need to follow the same practice in the future in order to maintain the type filtering logic in the UI
> 
> Was this discussed somewhere else with `kommander` team? Is UI the only consumer of this data?
> 
> I have no context to know if the right approach is to introduce a new type or if the UI has to derive this information from other fields.
> 
> So I defer the review to the `kommander` team.

Yes - I think there are other consumers here, too - I think `kommander-cli` is one of them, and I think `kommander` as well.

I don't know if just changing this here is going to work for all scenarios here.

akshatjohar:
> @akshatjohar Can you explain what are the new supported `type` annotation values the UI is now expecting? What is the outcome of these annotations - will UI consume them and display tags on the Application cards? Let's have a list of all supported `type`s with a clear definitions of what they all mean, and we can go through the list of platform apps here and redefine their `type`s all at once. This list should also include the non-platform app `type`s - like those coming from our partner catalog or external custom catalogs. I think it would be good for everyone to be on the same page on how to best redefine and use the `type` field moving forward.
> 
> Right now, the UI is the only consumer of this type field (App annotation). On upgrades, since we only display the latest versions of the platform apps (`ClusterApp`s) in the UI, these apps would reflect the updated `type` annotation automatically when Kommander is upgraded.


Hi @gracedo here's what I thought while formulating the logic to populate new types in the UI according to figma mockups. Also I would like to understand what would be the type and certifications for `partner catalog apps` which now need to be displayed as `Preferred Partner Apps` in the UI according to mockups

- if app is Nutanix Supported , `type: nutanix`, certifications should include `supported` (here I assume that `supported` and `qualified` cannot co-exist as certifications)
- If `type: platform/catalog` , certifications should include `qualified` , then regard them as `Certified Open Source` Apps and show `platform` tag explicitly in the card for `type: platform` apps
- If `type: partner` , then regard them as `Preferred Partner Apps` and show the tag `Preferred Partner`. I am not aware what king of certifications would be included here (in this case we might show the tag based on type or not depending on what certifications looks like for partner apps)

Does this explain the various app types or am I missing something , for external custom catalogs what are the apps are you referring to here ? If they have `type: catalog` , we would handle it the way above discussed

Although there's another way which I thought to display the UI types without changing types in the backend, but was not sure about the approach. It is as follows

- For Nutanix Apps - `type: platform` and certifications should include `supported`
- For Certified Open Source Apps - `type: platform/catalog` and certifications should include `qualified`
- For Preferred Partner Apps - (I am not aware of how the `metadata.yaml` looks like for partner catalog apps) , but `type: partner` ? (can derive tags in the UI from type or unique certifications entry (other than supported/qualified) if needed)

With this approach , we might need to change types for existing apps as of now if it fits every usecase / scenario , what do you think reg this approach ?

takirala (commented):
What happens to the older versions of applications that still use the `qualified` and `airgapped` certifications (During upgrades, Customers with catalog repositories etc.,) ? UI should support both certifications for backwards compatibility. We can deprecate in 2.16 and remove in 2.17 if we want.

gracedo:
@akshatjohar I am having a hard time following all these new types/certifications -- I have added a section for app metadata fields in this google doc https://docs.google.com/document/d/161r_MhTCsOR53WT5lp3wI-QSjsjluFYV5sv1AKzaZqw/edit?pli=1&tab=t.0#heading=h.mvy5rwvcv81j let's discuss there with a wider audience? I've just listed the current values we are using, can you update with your proposal of new/removed fields and also a description of how they should be used? With the v2 catalog apps work we're doing now, we will need to document these anyways since users will need to know how to use those fields.

PR #3594, opened 2025-06-13
fix: Revert "Create applications symlink to services"

gracedo:
Reverts mesosphere/kommander-applications#3591

It is causing installs to fail -- reverting for now, will probably just have to rename services to applications altogether.

### DIFF
--- a/applications
+++ b/applications
@@ -1,1 +1,0 @@
-services


msdolbey (approved):
👍 
eek

takirala (approved):

PR #3624, opened 2025-06-25
feat: an app to run AI Navigator Chatbot locally

asekretenko:

This introduces an app that installs the chatbot chart that perviously only has been run on internal premises by D2iQ/Nutanix.

The version is tentatively set to "0.5.0", despite the fact that the `v0.1.0` chart we've published for testing purposes  
is slightly newer than the `v0.5.0` tag in the `chatbot` repo, and that we use the dev image here.

The versions need to be brought into sync separately.

- https://jira.nutanix.com/browse/NCN-107090

Tested operation with our Azure OpenAI endpoint sending queries via port forwarding - seems to work.

```release-note

```

**Checklist**


alembiewski:
We should also update tooling in the kommander repo: https://github.com/mesosphere/kommander/blob/992f1266d20575c73dd8c7f8dce36401ba44fb4a/Makefile#L504

kaiwalyajoshi (changes requested):
@asekretenko could you please update the `.exclude-airgapped` according to the section here: https://github.com/mesosphere/kommander-applications?tab=readme-ov-file#airgapped-build-exclusions

kaiwalyajoshi (approved):
Approving as the airgapped changes are happening in sibling PR:https://github.com/mesosphere/kommander/pull/6058

PR #3641, opened 2025-06-29
feat: allow image parsing via bloodhound

takirala:

```release-note

```

**Checklist**


licenses.d2iq.yaml
```diff
@@ -464,11 +463,6 @@ resources:
         notice_path: NOTICE
         ref: ${image_tag}
         url: https://github.com/kubernetes-sigs/prometheus-adapter
-  - container_image: docker.io/semitechnologies/weaviate:1.27.6
```
takirala, line 467: This image is related to ai-navigator and is not needed here.

hack/list-oci-artifacts.sh
```diff
@@ -1,42 +0,0 @@
-#!/usr/bin/env bash
```
mhrabovcin, line 1: This script is used in `kommander` repo. The bundle creation make target will require an update.

takirala:
@mhrabovcin  I added the scripts back for now so that kapps CI can be unblocked by merging this, will update kommander repo after merging this.

mhrabovcin (approved):

PR #3663, opened 2025-07-01
refactor: migrated rook-ceph & rook-ceph-cluster to OCIRepository

shubham2g:
migrated rook-ceph & rook-ceph-cluster to OCIRepository
Bump Rook Ceph & Rook Ceph Cluster to 1.17.5

helm pull oci://ghcr.io/mesosphere/charts/rook-ceph-cluster --version v1.17.5

Pulled: ghcr.io/mesosphere/charts/rook-ceph-cluster:v1.17.5
Digest: sha256:a04bf948f943773149d233dad6c7677563ac313fa9edbfe3ad97b327a26ed40e

https://jira.nutanix.com/browse/NCN-108168

### DIFF
--- a/applications/rook-ceph/1.17.5/helmrelease/helmrelease.yaml
+++ b/applications/rook-ceph/1.17.5/helmrelease/helmrelease.yaml
@@ -8,7 +8,7 @@ spec:
   interval: 1m
   url: "${ociRegistryURL:=oci://ghcr.io}/mesosphere/charts/rook-ceph"
   ref:
-    tag: 1.17.5
+    tag: v1.17.5
 ---
 apiVersion: helm.toolkit.fluxcd.io/v2
 kind: HelmRelease


applications/rook-ceph/1.17.5/defaults/cm.yaml
```diff
@@ -1,20 +1,11 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: rook-ceph-1.17.1-config-defaults
+  name: rook-ceph-1.17.5-config-defaults
   namespace: ${releaseNamespace}
 data:
   values.yaml: |
     ---
```
shivendra-ntnx, line 8: New version doesn't require image??

shivendra-ntnx (approved):
Check if removal of image is intentional.

applications/rook-ceph/1.17.5/defaults/cm.yaml
shubham2g: I think you are referring to lines L9–15. Previously, I added this because I noticed that during license generation, the image tag was being pulled as master instead of the expected version 1.17.0. After reviewing the values.yaml in the Helm chart, I updated it accordingly. However, since it is now pulling the latest version by default, this update is no longer necessary — so I’ve removed it. I hope it is okay.

PR #3693, opened 2025-07-08
feat: adds vgpu token operator application

faiq:
Adds this chart to kommander applications https://github.com/orgs/nutanix-cloud-native/packages/container/package/vgpu-token-operator-charts%2Fvgpu-token-operator

```release-note

```

**Checklist**


gracedo (commented):
as discussed in slack, `vgpu-token-operator-1.0.5-config-defaults` defaults cm needs to be created in the `defaults/` dir (see any app for examples)

PR #3741, opened 2025-07-15
feat: add flux oci mirror app

mhrabovcin:
Add new `flux-oci-mirror` deployment.

```release-note

```

**Checklist**

### DIFF
--- a/applications/kommander-flux/2.6.1/mirror/flux-oci-mirror.yaml
+++ b/applications/kommander-flux/2.6.1/mirror/flux-oci-mirror.yaml
@@ -1,0 +1,107 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: flux-oci-mirror
+  labels:
+    app: flux-oci-mirror
+spec:
+  selector:
+    matchLabels:
+      app: flux-oci-mirror
+  replicas: 1
+  strategy:
+    rollingUpdate:
+      maxSurge: 25%
+      maxUnavailable: 25%
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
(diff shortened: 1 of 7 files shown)


mhrabovcin:
> do we need to update the flux update script at all? https://github.com/mesosphere/kommander-applications/blob/main/hack/flux/update-flux.sh

@gracedo I will need to add similar script for bumping this app once we have the release pipeline for it.

applications/kommander-flux/2.6.1/mirror/flux-oci-mirror.yaml
```diff
@@ -0,0 +1,107 @@
...
+      maxSurge: 25%
+      maxUnavailable: 25%
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: flux-oci-mirror
+    spec:
+      volumes:
+        - name: mirror-ca-bundle
```
cwyl02, line 24: should we change the volume name to `mirror-config`?

cwyl02 (approved):

cwyl02:
right now the app test failed due to missing cert-manager scheme https://github.com/mesosphere/kommander-applications/blob/main/apptests/flux/apply.go#L143-L154. but we can defer that change to another PR, since it will involve adding a new dependency to apptest module

mhrabovcin:
> right now the app test failed due to missing cert-manager scheme https://github.com/mesosphere/kommander-applications/blob/main/apptests/flux/apply.go#L143-L154. but we can defer that change to another PR, since it will involve adding a new dependency to apptest module

@cwyl02 thanks, I've added cert-manager CRDs to default kind installation.

PR #3753, opened 2025-07-17
refactor: default catalog configuration

takirala:
**What problem does this PR solve?**:

Relates to https://github.com/mesosphere/kommander/pull/6213

**Which issue(s) does this PR fix?**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Checklist**

- [ ] If the PR adds a version bump, ensure there is no breaking change in Licensing model (or NA).
- [ ] If a chart is changed or app configuration is significantly changed, the chart version is correctly incremented (so that apps are not automatically upgraded from a previous version of DKP).

PR #3768, opened 2025-07-18
feat: add nkp product catalog

johnliu25:

```release-note

```

**Checklist**

### DIFF
--- a/applications/kommander/0.16.0/defaults/cm.yaml
+++ b/applications/kommander/0.16.0/defaults/cm.yaml
@@ -130,3 +130,6 @@ data:
     catalogCollections:
     - registry: ghcr.io
       repository: nutanix-cloud-native/nkp-partner-catalog
+    - registry: ghcr.io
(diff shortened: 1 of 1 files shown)


applications/kommander/0.16.0/defaults/cm.yaml
```diff
@@ -130,3 +130,6 @@ data:
     catalogCollections:
     - registry: ghcr.io
       repository: nutanix-cloud-native/nkp-partner-catalog
+    - registry: ghcr.io
+      repository: nutanix-cloud-native/nkp-nutanix-product-catalog
+      branch: v2.0.0
```
takirala, line 135: This will be calculated dynamically and needs to be dropped. 
Also, lets skip this for dev.12

johnliu25: Could you please tell me where is the logic that the branch will be calculated dynamically?

takirala:
Done in #3753

PR #3798, opened 2025-07-23
feat: update the bitnami to official dns

saikumar98419:

update the bitnami to official dns


mhrabovcin (changes requested):
@saikumar98419 we'll have to use app version `>9.0.0` otherwise the new version would be considered as lower than the 9. I am thinking we can use `10` prefix so `10.18.0`

takirala:
closing stale PR. please reopen as needed.

PR #3840, opened 2025-08-02
chore: Readd AI-Navigator RAG for release in NKP v2.16.0

kaiwalyajoshi:
**What problem does this PR solve?**

This PR is a revert of #3773 and land the following additional fixes atop the revert patch:
- Rename `ai-navigator-gateway` to `ai-navigator-rag`
- Bump versions to `0.6.0`
- Use `pullthrough.infra.nkp.sh` to pull the private docker image till we Dockerhub repo Public.

#3773 had a companion reverted PR in Kommander [PR-6220](https://github.com/mesosphere/kommander/pull/6220) which due to new NKP-Full-Bundle doesn't need to be reverted anymore.

```release-note

```

**Checklist**


kaiwalyajoshi:
@gracedo I don't think we need to do anything further to include `ai-navigator-rag` into the NKP Full Bundle right?

PR #3851, opened 2025-08-04
fix: NCN-108862 adds ubuntu prebuilt driver container for pro

faiq:
For the Ubuntu OOTB OS, we need to add the nvidia precompiled drivers to the NKP airgap package. 

This will be required to allow NKP customers to use GPU workloads in airgap scenarios in combination with Ubuntu OOTB OS

https://jira.nutanix.com/browse/NCN-108862

```release-note

```

**Checklist**


msdolbey (approved):
👍 
yeah sure

PR #3873, opened 2025-08-08
feat: Add pod anti-affinity rules for eventing-webhook deployment

lukeogg:

Made adjustments to the Knative `eventing-webhook` deployment:

- Added a `podDisruptionBudgets` override in the config. This is set to `maxUnavailable: 1` to be consistent with our original Knative chart. This setting allows the replicas to be scaled from 1 to n and will still allow nodes to drain.
- Added `replicas: 3` as a default. Consistent with the idea, Knative would remain available during an upgrade and eventing is generally a HA component.
- Added preferred `podAntiAffinity` to spread pods among nodes.

https://jira.nutanix.com/browse/NCN-108946

Tested by draining nodes in a dev cluster.

```release-note

```

**Checklist**

### DIFF
--- a/applications/knative/1.18.1/defaults/cm.yaml
+++ b/applications/knative/1.18.1/defaults/cm.yaml
@@ -69,6 +69,23 @@ data:
       namespace: knative-eventing
       manifest:
         spec:
+          deployments:
+            - name: eventing-webhook
+              replicas: 3
+              podTemplate:
+                spec:
+                  affinity:
+                    podAntiAffinity:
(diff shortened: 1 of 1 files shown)


kaiwalyajoshi (approved):

msdolbey:
I think generally too, if someone has KNative in their cluster, they will have a pretty beefy setup, so I think this is okay to bump the number of replicas here.

msdolbey (approved):
👍 
The real disruptor

PR #3913, opened 2025-08-20
fix: remove grafana cpu limit & use recommended memory for limit

cwyl02:
Grafana ran into OOMKilled couple time with the default memory setting, turns out grafana minimum requirement is 512M memory and 1 cpu core.

https://jira.nutanix.com/browse/NCN-109257

```release-note

```

**Checklist**


fatz (approved):
awesome!

gracedo (approved):

PR #3929, opened 2025-08-24
fix: re-revert git-operator manifests

msdolbey:
Git Operator Admin Creds Rotate Job uses the bitnami/kubectl image
It should use bitnamilegacy/kubectl 

https://jira.nutanix.com/browse/NCN-109366

This change was failing some test earlier, is it safe to ignore this test? Can we just check to see if this can fix the issue without publishing a new git operator image altogether?

N/A

**Checklist**

### DIFF
--- a/services/git-operator/0.1.4/additional-images.txt
+++ b/services/git-operator/0.1.4/additional-images.txt
@@ -1,0 +1,1 @@
+docker.io/bitnamilegacy/kubectl:1.32.3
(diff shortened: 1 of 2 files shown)


gracedo:
Test will continue to fail bc the manifests now differ from the git-operator "source of truth" manifests at v0.13.7... otherwise to pull in the image update from upstream we would need to update git-operator to v0.13.11, which pulls in other changes we probably don't want to for this bitnami patch release https://github.com/mesosphere/git-operator/compare/v0.13.7...v0.13.11

gracedo:
I think this should also resolve it https://github.com/mesosphere/kommander-applications/pull/3931

PR #3936, opened 2025-08-25
fix: 2.14.2 + bitnamilegacy ONLY

cwyl02:
**What problem does this PR solve?**:
bitnami legacy fix for release-2.14

**Which issue(s) does this PR fix?**:

**Special notes for your reviewer**:

kommander: https://github.com/mesosphere/kommander/pull/6367

**Does this PR introduce a user-facing change?**:

```release-note

```

**Checklist**

- [ ] If the PR adds a version bump, ensure there is no breaking change in Licensing model (or NA).
- [ ] If a chart is changed or app configuration is significantly changed, the chart version is correctly incremented (so that apps are not automatically upgraded from a previous version of DKP).

PR #3937, opened 2025-08-25
fix: 2.14 bitnamilegacy

cwyl02:

```release-note

```

**Checklist**


gracedo:
I think we can just use https://github.com/mesosphere/kommander-applications/pull/3936 -- the only diff since 2.14.1 release were post-release bumps

cwyl02:
using https://github.com/mesosphere/kommander-applications/pull/3936

PR #4036, opened 2025-09-12
refactor: remove defaults dir for applications

takirala:

This PR has commits organized by the scope but since it has close to 90 commits, following is the gist of the changes:
- Removed `defaults` directory from all applications. 
- Introduced a new `releaseName` substitution var. 
  - I will follow up with another `s/releaseName/appName` but that is a breaking change and would require a version bump for each and every application.
- Adds a bunch of new CRDs in the bloodhound config. This is needed cuz now each application is processed with isolation. We can soon replace this with having a .bloodhound.yml file inside the application version directory itself.
- Moves the top level `charts/*` files into their respective application directories.
- Drops the invocation of bloodhound directly. We use nkp cli now (which uses the catalog CLI which in turn calls bloodhound)
- Remove the manifest validation check as the new license manifest check validates and build the artifacts
- Switches to beefy nutanix runners (the new check is more resource hungry and I was getting some flaky behavior on current runners)
- Made some fixes in our apptests/appscenarios tooling (diag bundle collection has been broken for a while)

I realize this is an overwhelming change, I will quickly follow up with more fixes if we discover things break after merging this.

```release-note

```

**Checklist**

### DIFF
--- a/.github/workflows/validate-licenses.yaml
+++ b/.github/workflows/validate-licenses.yaml
@@ -31,6 +31,8 @@ jobs:
         uses: jetify-com/devbox-install-action@v0.13.0
         with:
           skip-nix-installation: true
+          enable-cache: 'false'
+          refresh-cli: 'true'
 
       - name: Login to Docker Hub
         uses: docker/login-action@v3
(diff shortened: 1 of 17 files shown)


.github/workflows/validate-licenses.yaml
```diff
@@ -14,7 +14,7 @@ on:
       - main
 jobs:
   check-license-yaml:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted-nutanix-docker-large
```
mhrabovcin, line 17: Changing a single workflow to use different kind of runner will result in incompatible cache generated by devbox GHA for the existing ubuntu runners. This will require also turning off the cache or running settings the `refresh-cli` to `true` for the devbox GHA.

Example https://github.com/mesosphere/kommander-applications/actions/runs/18095323069/job/51484923099?pr=4066

cwyl02 (approved):

PR #4045, opened 2025-09-16
feat: opencost init

cwyl02:
adding the new opencost k-app

https://jira.nutanix.com/browse/NCN-109220

```release-note

```

**Checklist**


applications/opencost/2.2.7/helmrelease/cm.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+      dataRetention:
+        dailyResolutionDays: 14
+
+      # frontend
+      ui:
+        image:
+          registry: ghcr.io
+          repository: opencost/opencost-ui
+          tag: "1.117.3"
```
gracedo, line 33: is there a reason why we're pinning the tag instead of just using the default?

applications/opencost/2.2.7/helmrelease/cm.yaml
```diff
@@ -0,0 +1,78 @@
...
+                - /dkp/opencost/frontend
+
+      # backend
+      exporter:
+        image:
+          registry: ghcr.io
+          repository: opencost/opencost
+          tag: "1.117.3"
+        extraEnv:
+          CURRENT_CLUSTER_ID_FILTER_ENABLED: "false"
```
gracedo, line 53: I think for single cluster opencost, this should be `true` or maybe even omitted? (we'd set it for centralized opencost)

cwyl02: omitted

gracedo (approved):

PR #4102, opened 2025-10-03
feat: add dependencies to opencost

cwyl02:

1. add dependencies
2. bump apps
3. add kustomization to opencost k-app

```release-note

```

**Checklist**

### DIFF
--- a/applications/opencost/2.3.1/kustomization.yaml
+++ b/applications/opencost/2.3.1/kustomization.yaml
@@ -1,4 +1,4 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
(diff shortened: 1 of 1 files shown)


applications/opencost/2.3.1/metadata.yaml
```diff
@@ -3,6 +3,8 @@ displayName: OpenCost
 description: OpenCost provides cost monitoring and resource optimization for Kubernetes workloads.
 category:
   - monitoring
+requiredDependencies:
```
gracedo, line 6: maybe not a hard dep bc users can update config to point to their own prom

cwyl02: done

PR #4671, opened 2026-03-24
chore: handle pre-release overrides for release operator

kshcore:
Release operator will point to staging harbor registry and pre-release tag incase of dev/rc builds (so only discoverable version would be this pre-release version). For now since QA isn’t sure either how/what will be tested with custom/fake releases during upgrade qual. For prod, operator will point to prod ghcr and a mutable “latest” tag (publishing/tagging this will be done as part of release-ops)

- Add releasemetadata package to generate staging OCI configuration for the release operator during pre-release builds
- Use separate Flux Kustomizations with dependsOn to ensure the ConfigMap is created before the release-operator Deployment
- Production builds remain unaffected (no config files generated)
- During post release, it undo the changes done in pre-release.

When the pre-release command runs, it should generate -
```
common/release/
├── kustomization.yaml                    
├── flux-kustomization.yaml       (include dependsOn flux-pre-release-kustomization)
├── flux-pre-release-kustomization.yaml   
└── config/                               
    ├── kustomization.yaml               
    └── release-operator-vars.yaml        # ConfigMap with staging OCI config details
```
for production, the config/ directory and flux-kustomization-config.yaml shouldn't exist, and kustomization.yaml only references flux-kustomization.yaml (no dependsOn).

```release-note

```

**Checklist**

### DIFF
--- a/hack/release/cmd/postrelease/postrelease.go
+++ b/hack/release/cmd/postrelease/postrelease.go
@@ -11,6 +11,7 @@ import (
 	"github.com/mesosphere/kommander-applications/hack/release/pkg/appversion"
 	"github.com/mesosphere/kommander-applications/hack/release/pkg/chartversion"
 	"github.com/mesosphere/kommander-applications/hack/release/pkg/extraimages"
+	"github.com/mesosphere/kommander-applications/hack/release/pkg/releasemetadata"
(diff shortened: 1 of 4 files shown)


guillermoz-nutanix (approved):
IDK if this is the best approach to do this.

PR #4751, opened 2026-04-09
fix: remove upgradeplan and release operators

mhrabovcin:
Remove upgrade and release operators from default installed list.

https://jira.nutanix.com/browse/NCN-113256

```release-note

```

**Checklist**


gracedo (approved):

gracedo (commented):
actually this is written from the script in kommander
https://github.com/mesosphere/kommander/blob/b22d68b0798b652d83857a0a3a3586f7a955130a/hack/prepare-sync-manifests.sh#L140-L149

PR #4796, opened 2026-04-18
feat: use helm v4 SSA for application upgrades

takirala:

Follow up of #4621 

Explicitly use SSA for applications during upgrades.
For fresh installs, flux already defaults to using SSA if the SSA field is omitted.

```release-note

```

**Checklist**


mikolajb (approved):

cwyl02 (approved):

takirala:
Admin merging in interest of dev cut. Will follow up with CI failures. Given that fresh install worked in last dev cut with SSA, I am bettting that upgrades should work with SSA as well. Tests aren't being run on main and the ones failing on this PR haven't been run in a while.

PR #4816, opened 2026-04-23
chore: bump helm-controller to 1.5.4

takirala:

Consuming latest helm-controller to get in couple bug fixes https://github.com/fluxcd/helm-controller/blob/v1.5.4/CHANGELOG.md#154

https://jira.nutanix.com/browse/NCN-113741

### DIFF
--- a/applications/kommander-flux/2.8.5/helmrelease/cm.yaml
+++ b/applications/kommander-flux/2.8.5/helmrelease/cm.yaml
@@ -13,6 +13,7 @@ data:
     policies:
       create: true
     helmController:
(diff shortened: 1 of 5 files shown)


cwyl02 (approved):

takirala:
I tested this manually on a kind cluster where:
1. Installed flux
2. installed kubefed using `helm` cli (no ssa)
3. waited for kubefedconfig to be managed by kubefed controller-manager
4. applied flux HR so that flux will honor the `disableTakeOwnership` and not overwrite the manager.

This is not the bug that was observed in the original cluster but the fix should work I think.

PR #4821, opened 2026-04-24
fix: add dependsOn to opencost apps

cwyl02:
currently opencost pods can restart couple time because waiting for prometheus pod. add an explicit dependson so that HR of opencost won't even try to spin up if KPS is not ready

https://jira.nutanix.com/browse/NCN-113306

```release-note

```

**Checklist**

### DIFF
--- a/applications/centralized-opencost/2.5.14/helmrelease.yaml
+++ b/applications/centralized-opencost/2.5.14/helmrelease.yaml
@@ -4,6 +4,9 @@ metadata:
   name: centralized-opencost-release
   namespace: ${releaseNamespace}
 spec:
+  dependsOn:
+    - name: thanos
+      namespace: ${releaseNamespace}
   force: false
   prune: true
   wait: true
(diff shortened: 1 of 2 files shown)


mikolajb (approved):

gracedo:
do we allow users to configure opencost to use their own prometheus? if we support this scenario and they disable KPS, then this app would never become ready 🤔

cwyl02:
> do we allow users to configure opencost to use their own prometheus? if we support this scenario and they disable KPS, then this app would never become ready 🤔

great call. we have https://jira.nutanix.com/browse/NCN-109996 which is still open right now. 

but on the other hand, if they really are obsessed with running a custom KPS, they should also be able to overcome this by a kustomization patch modifying this dependson block?

an easier hack for them would be they use flux to deploy the HR with the same name in the same releaseNamespace?

cwyl02:
> do we allow users to configure opencost to use their own prometheus? if we support this scenario and they disable KPS, then this app would never become ready 🤔

got cleared from PM to fix this now for most customers

PR #4822, opened 2026-04-24
fix: cert-manager circular dependency

cwyl02:
found in kommander repo's kuttl multicluster CI, as well as QA testing, where the cluster attachment takes 30+ minutes. turns out the certmanager on workload cluster is not ready, due to its new dependency of gateway API. 

This PR eliminated this hard dependency, but when gateway api crd App is finishes its install,  eventually the cert-manager HR will pick up the override value of gateway api crd feature flag of cert-manager chart, enabling NAI's feature.

https://jira.nutanix.com/browse/NCN-113822

```release-note

```

**Checklist**


takirala (approved):

applications/cert-manager/1.19.3/release/release.yaml
```diff
@@ -50,6 +50,10 @@ spec:
   valuesFrom:
     - kind: ConfigMap
       name: ${releaseName}-${appVersion}-config-defaults
+    # this is needed to support NAI
+    - kind: ConfigMap
+      name: cert-manager-gateway-api-override
```
takirala:
```suggestion
      name: cert-manager-gateway-api-crds-${releaseNamespace}-overrides
```
we can use the same naming in other configmap. just to make sure some uniqueness is gaurenteed.

applications/gateway-api-crds/1.11.1/helmrelease/cm.yaml
```diff
@@ -10,3 +11,17 @@ data:
     traefik: false
     hub: false
     deleteOnUninstall: false
+---
+# this is needed to support NAI
+# https://jira.nutanix.com/browse/NCN-112584
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cert-manager-gateway-api-override
```
takirala:
```suggestion
  name: cert-manager-gateway-api-crds-${releaseNamespace}-overrides
```

applications/gateway-api-crds/1.11.1/helmrelease/gateway-api-crds.yaml
```diff
@@ -26,6 +26,9 @@ spec:
   valuesFrom:
     - kind: ConfigMap
       name: ${releaseName}-${appVersion}-config-defaults
+    - kind: ConfigMap
+      name: gateway-api-crds-config-overrides
```
takirala: why is this change here in this PR? this should be done automatically by our controllers anyway right.

applications/cert-manager/1.19.3/release/release.yaml
takirala: in the future, if we want to be pedantic about version dependency between a specific cert manager and a specific gateway-api-crds we can add version of gateway-api-crds here but this is fine for 2.18.

applications/gateway-api-crds/1.11.1/helmrelease/gateway-api-crds.yaml
cwyl02: oops good catch

applications/gateway-api-crds/1.11.1/helmrelease/cm.yaml
```diff
@@ -10,3 +11,17 @@ data:
     traefik: false
     hub: false
     deleteOnUninstall: false
+---
+# this is needed to support NAI
+# https://jira.nutanix.com/browse/NCN-112584
```
takirala: instead of NAI, can we adjust the comment so that it tells how (or what CRD) cert manager consumes from this change. we can retain the jira number if you prefer. 

Techncially this change would make sense even outside NAI world.

takirala (approved):

applications/gateway-api-crds/1.11.1/helmrelease/cm.yaml
cwyl02: done

takirala (approved):

PR #4860, opened 2026-05-11
fix: fixing chart version for rook-ceph and rook-ceph cluster

pawan-bobbili-ntnx:
 - fixing chart version for rook-ceph and rook-ceph cluster, pointing to v1.19.5 tag

https://jira.nutanix.com/browse/NCN-114203

**Screenshots**
 - Pre-req job was failing before
 
Screenshot 2026-05-11 at 9 53 27 PM

### DIFF
--- a/applications/rook-ceph-cluster/1.19.5/helmrelease/dashboard-cm.yaml
+++ b/applications/rook-ceph-cluster/1.19.5/helmrelease/dashboard-cm.yaml
@@ -9,5 +9,5 @@ data:
   name: "Rook Ceph Cluster"
   dashboardLink: "/dkp/kommander/ceph-dashboard/"
   docsLink: "https://docs.ceph.com/en/latest/mgr/dashboard/"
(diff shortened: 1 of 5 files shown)


takirala (approved):

deepakm-ntnx:
bug was introduced thru https://github.com/mesosphere/kommander-applications/pull/4853
@pawan.bobbili just wondering why the folder name was changed earlier without the bump in the version? how do we not have such issues next time? is the automation which upgrade the version is incorrect? cc @shivendra-ntnx @gracedo @takirala

deepakm-ntnx:
so many tests are failing. whats the plan for it

gracedo (approved):

pawan-bobbili-ntnx:
@deepakm-ntnx ,
yes we realized this while doing RCA, looks like it was missed before while bumping apps, and charts of v1.19.5 were also not pushed to OCI for rook-ceph and rook-ceph-cluster.
Looks like upgrade flow was breaking from long time, thus these tests were ignored and old PR was merged.
@gracedo can you please throw some light here.

NRoshanRai (approved):

PR #4867, opened 2026-05-12
feat(cert-manager): Use upstream oci chart

gracedo:
We should directly use upstream OCI helm chart (this is also the source that konvoy is using)

https://jira.nutanix.com/browse/NCN-114270

```release-note

```

**Checklist**


faiq (approved):
https://github.com/nutanix-cloud-native/cluster-api-stack-operator/blob/main/make/helm.mk#L50

aligns with this

PR #4912, opened 2026-05-29
feat: add ai applications catalog for future

takirala:

Adding an empty catalog collection for now. In near future, this catalog will get updates as ai catalog work progresses.

https://jira.nutanix.com/browse/NCN-114727

### DIFF
--- a/applications/kommander/0.18.0/helmrelease/cm.yaml
+++ b/applications/kommander/0.18.0/helmrelease/cm.yaml
@@ -141,5 +141,5 @@ data:
     - resourceURI: oci://ghcr.io/nutanix-cloud-native/nkp-ai-applications-catalog/collection
       catalogName: nkp-ai-applications-catalog
       extraLabels: {}
-      tag: "" # Defaults to MAJOR.MINOR of KommanderCore when empty.
+      tag: 2.18.0
       airgappedDiscovery: false
--- a/applications/kommander/0.18.0/helmrelease/cm.yaml
+++ b/applications/kommander/0.18.0/helmrelease/cm.yaml
@@ -138,3 +138,8 @@ data:
       extraLabels: {}
       tag: "" # Defaults to MAJOR.MINOR of KommanderCore when empty.
       airgappedDiscovery: true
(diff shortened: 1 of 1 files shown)


kaiwalyajoshi (approved):

applications/kommander/0.18.0/helmrelease/cm.yaml
```diff
@@ -138,3 +138,8 @@ data:
       extraLabels: {}
       tag: "" # Defaults to MAJOR.MINOR of KommanderCore when empty.
       airgappedDiscovery: true
+    - resourceURI: oci://ghcr.io/nutanix-cloud-native/nkp-ai-applications-catalog/collection
+      catalogName: nkp-ai-applications-catalog
+      extraLabels: {}
+      tag: "" # Defaults to MAJOR.MINOR of KommanderCore when empty.
```
takirala:
```suggestion
      tag: 2.18.0
```
lets pin this to 2.18.0 cuz we are not sure if there would be any breaking changes b/w 2.18.0 -> 2.18.1 from ai catalog perspective. If we don't get any breaking changes we will make `2.18` and `2.18.0` tags identical which would be the most likely outcome. This is just being defensive.

applications/kommander/0.18.0/helmrelease/cm.yaml
```diff
@@ -138,3 +138,8 @@ data:
       extraLabels: {}
       tag: "" # Defaults to MAJOR.MINOR of KommanderCore when empty.
       airgappedDiscovery: true
+    - resourceURI: oci://ghcr.io/nutanix-cloud-native/nkp-ai-applications-catalog/collection
+      catalogName: nkp-ai-applications-catalog
+      extraLabels: {}
+      tag: 2.18.0
```
gracedo, line 144: just double checking but this'll be different from `# Defaults to MAJOR.MINOR of KommanderCore when empty.` and we will be releasing patch versions of the catalog, so each nkp release will bring in a new release of ai catalog?

gracedo: just saw https://github.com/mesosphere/kommander-applications/pull/4912#discussion_r3325765065

PR #4913, opened 2026-05-29
fix: Postrelease bumps

gracedo:
Fix postrelease script (see https://github.com/mesosphere/kommander-applications/pull/4906)
Run it to execute the fixed script.

```release-note

```

**Checklist**


takirala (approved):

gracedo:
image
Meh, I guess this automation also needs to be fixed.. it's deleting entries that it shouldn't be

PR #4975, opened 2026-06-10
chore: Add backport configuration file (.backportrc.json)

lukeogg:

Backport PRs generated by the `sorenlouv/backport-github-action` use the action's
default title template, `[{{targetBranch}}] {{commitMessages}}`. The leading
`[release-x.y]` prefix means the conventional-commit type (`fix:`, `feat:`, etc.)
is no longer at the start of the title, so the `Lint PR` workflow
(`amannn/action-semantic-pull-request`) fails on every backport with
"No release type found in pull request title".

Until now this has required manually editing each backport PR title to move the
release marker after the type. This PR adds a `.backportrc.json` with a custom
`prTitle` so backport titles keep the conventional type first
(e.g. `fix: Adjust Knative webhook memory (#4970) (backport to release-2.18)`),
passing both the conventional-commit check and the 72-char title length check
automatically.

- Uses `{{{commitMessages}}}` (triple braces) to avoid the Handlebars HTML-escaping
  issue documented for this action.
- Uses `commitMessages` rather than `sourcePullRequest.title` so the original
  `(#xxxx)` reference is preserved, matching the content of the action's previous
  default — just reordered.
- Only `prTitle` is overridden; all other action defaults still apply, and
  `repoOwner`/`repoName` come from the workflow context so they're not needed here.
- This applies to future backports only; existing open backport PRs still need
  their titles edited manually.

```release-note
NONE
```

**Checklist**

### DIFF
--- a/.backportrc.json
+++ b/.backportrc.json
@@ -1,0 +1,3 @@
+{
+  "prTitle": "{{{commitMessages}}} (backport to {{targetBranch}})"
+}


mikolajb:
Development of kommander-applications is moved to https://github.com/mesosphere/kommander/tree/main/kommander-applications

PR #5080, opened 2026-07-28
feat: update the k8's version to 1.33.13

saikumar98419:


NRoshanRai (approved):

.bloodhound.yml
```diff
@@ -3,7 +3,7 @@
 k8s_versions:
   - 1.31.9
   - 1.32.3
-  - 1.33.5
+  - 1.33.13
```
NRoshanRai, line 6: there are multiple references of kubetoolsImageTag:=v1.33.5-alpine , don't we need those to be change ?

saikumar98419: Generally for minor version we used to change in Bloodhound .yaml

PR #5082, opened 2026-07-29
feat: istio app 1.23.6 to 1.25.1

pawan-bobbili-ntnx:
 - Removing from documentation, addition of redundant env variables of ztunnel while enabling ambient mode, as chart in built block has the same and 2.8 SSA gives error for this scenario.


NRoshanRai (approved):

NRoshanRai:
waiting for the apptest result

pawan-bobbili-ntnx:
Screenshot 2026-08-21 at 9 52 22 PM

